PR #9, opened 2022-08-15
Fix undefined names in Python type hints

cclauss:
$ `pre-commit autoupdate`
```
Updating https://github.com/pre-commit/pre-commit-hooks ... updating v2.3.0 -> v4.3.0.
Updating https://github.com/myint/autoflake ... already up to date.
Updating https://github.com/asottile/pyupgrade ... updating v2.32.0 -> v2.37.3.
Updating https://github.com/psf/black ... updating 22.3.0 -> 22.6.0.
```
Fix the following undefined names in Python type hints:
$ `flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics`
```
./ofrak_core/ofrak/core/patch_maker/linkable_binary.py:150:22: F821 undefined name 'PatchMaker'
        patch_maker: "PatchMaker",  # type: ignore
                     ^
./ofrak_core/ofrak/service/error.py:55:49: F821 undefined name 'DataNode'
    def __init__(self, message, new_child_node: "DataNode", existing_child_node: "DataNode"):  # type: ignore
                                                ^
./ofrak_core/ofrak/service/error.py:55:82: F821 undefined name 'DataNode'
    def __init__(self, message, new_child_node: "DataNode", existing_child_node: "DataNode"):  # type: ignore
                                                                                 ^
3     F821 undefined name 'PatchMaker'
3
```
Add `from __future__ import annotations` so that `pyupgrade` can reduce typing imports by using the updated syntax.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v4.3.0
     hooks:
       - id: end-of-file-fixer
   - repo: https://github.com/myint/autoflake
@@ -9,11 +9,11 @@ repos:
       - id: autoflake
         args: [ '--in-place', '--remove-all-unused-imports', '--ignore-init-module-imports' ]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.32.0
+    rev: v2.37.3
     hooks:
       - id: pyupgrade
         args: ['--py37-plus']
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 22.6.0
     hooks:
       - id: black
--- a/ofrak_core/ofrak/core/patch_maker/linkable_binary.py
+++ b/ofrak_core/ofrak/core/patch_maker/linkable_binary.py
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 import dataclasses
 import logging
 import os
 from dataclasses import dataclass
-from typing import Dict, Iterable, Optional, Tuple
+from typing import Iterable
 
 from ofrak.component.modifier import Modifier
 from ofrak.core.addressable import Addressable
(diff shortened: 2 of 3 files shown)


EdwardLarson:
The problem with `from __future__ import annotations` is that it is implemented as "stringifying" all type annotations:

```
>>> def foo1(x: int) -> int:
...     return x+1
...
>>>
>>> foo1.__annotations__
{'x': <class 'int'>, 'return': <class 'int'>}  # Notice that the annotations are actual Python types
>>>
>>> from __future__ import annotations
>>>
>>> def foo2(x: int) -> int:
...     return x+1
...
>>>
>>> foo2.__annotations__
{'x': 'int', 'return': 'int'}  # Notice that the annotations are now simple strings
```

The reason this is a problem is that the [synthol](https://pypi.org/project/synthol/) tool we use for dependency injection relies on having the actual types available from the annotations. To be clear not every annotation needs to be a type literal, and in many files we could use `from __future__ import annotations` without stringifying any critical annotations that `synthol` needs access to. However, for consistency's sake and to avoid unexpected behavior, we are strictly avoiding `from __future__ import annotations` and only stringifying the specific annotations which need it (which are relatively few).

This also means we cannot do the `Dict[A, B]` -> `dict[A, B]` (and similar) changes you have also proposed, as [before Python 3.9](https://peps.python.org/pep-0585/) `dict` is not a `Generic` type and requires `from __future__ import annotations` to be treated as such. We still support Python 3.7, and disallow `from __future__ import annotations`, so this change cannot be accepted.

I also see changes to use the overloaded `|` operator instead of the `Optional` type (and could expand to cover `Union` types as well), which is a great change to Python typing I was not personally aware of! However the [relevant PEP](https://peps.python.org/pep-0604/) is only introduced in Python 3.10 so again for backwards compatibility reasons we cannot accept the changes.

There are other changes to the pre-commit revisions, which I believe are an entirely separate issue. I'd like to ask to please keep the title/description of Pull Requests descriptive of the changes within. In this case the changes are not all clearly related to the title and description, which makes the review process less efficient.

Thank you for your enthusiasm! Hopefully this is not too disappointing :)

PR #12, opened 2022-08-17
Fix undefined names in Python type hints - again

cclauss:
Replace #9 with a solution that does not use `from __future__ import annotations`
1. Run https://pypi.org/project/com2ann to translate type comments into Python type annotations.
2. Resolve the undefined names in type hints but **_without_** `from __future__ import annotations` which caused problems in #9

### DIFF
--- a/ofrak_core/ofrak/core/patch_maker/linkable_binary.py
+++ b/ofrak_core/ofrak/core/patch_maker/linkable_binary.py
@@ -19,6 +19,7 @@ from ofrak.service.resource_service_i import (
 )
 from ofrak.core.patch_maker.linkable_symbol import LinkableSymbol, LinkableSymbolType
 from ofrak_patch_maker.model import BOM, PatchRegionConfig
+from ofrak_patch_maker.patch_maker import PatchMaker
 from ofrak_patch_maker.toolchain.model import Segment
 from ofrak_type.error import NotFoundError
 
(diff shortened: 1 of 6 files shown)


ofrak_core/ofrak/core/patch_maker/linkable_binary.py
```diff
@@ -19,6 +19,7 @@
 )
 from ofrak.core.patch_maker.linkable_symbol import LinkableSymbol, LinkableSymbolType
 from ofrak_patch_maker.model import BOM, PatchRegionConfig
+from ofrak_patch_maker.patch_maker import PatchMaker
```
EdwardLarson, line 22: Introduces a circular import, as `patch_maker.py` imports from `program.py` which imports from `linkable_binary`. 

We are planning a PR which will remove the PatchMaker dependency on `Program` and other OFRAK classes, so this change will be possible in the future.

ofrak_core/ofrak/service/error.py
```diff
@@ -4,6 +4,7 @@
 import sys
 
 import ofrak_type.error
+from ofrak.service.data_service import DataNode
```
EdwardLarson, line 7: Introduces a circular import, as `data_service.py` will import from `error.py`

EdwardLarson (changes requested):
There are two locations where "de-stringifying" some type hints creates some circular imports which would prevent OFRAK from running. The other changes look good, so if those are fixed then this should be all set!

PR #27, opened 2022-09-07
Add GitHub CI Pipeline

rbs-jacob:
**Link to Related Issue(s)**

N/A

- Add a CI pipeline that creates a GitHub Actions workflow to run tests for pull requests
- Implement caching in the CI pipeline such that minimal time is spent rebuilding images
  - This is done by adding a `--cache-from` option to the `build_image.py` script so that it can specifically try and use cached layers from a different image. In this particular case, the cached image is `ofrak-core-base`, which is built once and passed to each of the concurrent test jobs. Each of those jobs builds its own, more specific image to test with, but reuses some of the cached layers.
- Remove a few duplicated dependencies (namely keystone, capstone, and cmake) from `ofrak_components` when they're already built and installed in `ofrak_core`

@whyitfor


ofrak_components/Dockerstub
```diff
@@ -16,39 +16,6 @@ RUN apt-get -y update && \
       u-boot-tools \
       unar
 
-# Install CMake
```
whyitfor, line 19: Can you document what this change is in the PR title/description?

whyitfor: Also, is the Keystone install in `ofrak_core/Dockerstub` still needed?

build_image.py
```diff
@@ -73,13 +74,18 @@ def main():
 
     if config.build_base:
         full_base_image_name = "/".join((config.registry, config.base_image_name))
+        cache_args = []
```
whyitfor, line 77: Can you explain this change in the PR description?

.github/workflows/test-all.yml
```diff
@@ -0,0 +1,206 @@
...
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          lfs: true
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
```
whyitfor, line 29: Aren't we using 3.7 inside of the Docker image?

whyitfor (changes requested):
Looks good! 

Please see a few comments

ofrak_components/Dockerstub
rbs-jacob: Documented the change in the description.

For now, the RBS Keystone build is still used in `ofrak_core`. I don't know if there will be test failures from switching to a `pip install keystone` setup, but it feels out of scope for this PR.

whyitfor (approved):

PR #29, opened 2022-09-13
Add optional "pack" parameter to `Resource.write_to()` and `Resource.flush_to_disk()`

sjossi:
This PR adds an optional "pack" parameter to `Resource.write_to()` and `Resource.flush_to_disk()` to suppress the `self.pack_recursively()` that usually happens in the beginning.

This solves an issue that happens when a resource should be written to disk as is, but was previously `identify`ed and no packer is available.

MRE:
```
from ofrak import OFRAK, OFRAKContext

from pathlib import Path
import tempfile

async def oopsie(ofrak_context: OFRAKContext):
    root_resource = await ofrak_context.create_root_resource("lzmainlzma", b"\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x2e\x80\x2d\x00\x00\x7f\xeb\x0b\xed\x6e\xd4\x23\x99\xb4\xf3\x0e\x8d\x71\xaf\xff\xff\xf9\xbe\x40\x00",
    )

    await root_resource.unpack()

    with tempfile.TemporaryDirectory() as tempdir:
        for child_resource in await root_resource.get_children():
            # works
            try:
                await child_resource.flush_to_disk(f"{tempdir}/{child_resource.get_id().hex()}-unidentified")
            except Exception as e:
                print(f"Could not flush: {e}")

            # breaks
            try:
                await child_resource.identify()
                await child_resource.flush_to_disk(f"{tempdir}/{child_resource.get_id().hex()}-identified")
            except Exception as e:
                print(f"Could not flush: {e}")

        [print(f) for f in Path(tempdir).iterdir()]

ofrak = OFRAK()
ofrak.run(oopsie)
```

The resource built here is a LZMA in a LZMA file. We unpack it to write all its children to disk (the contained LZMA file). If we don't `identify()` the resource, it will be written as bytes as expected. If we `identify()` it first, `pack_recursively()` tries to use the LZMA packer, but fails. With the new parameter we can suppress that behavior and just write the bytes of the resource as is.


ofrak_core/test_ofrak/unit/test_resource.py
```diff
@@ -170,3 +172,61 @@ async def test_get_most_specific_tags(resource: Resource):
...
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        want = set()
+
+        flush_dir = Path(tempdir, "flush_to_disk")
+        write_dir = Path(tempdir, "write_to")
+
+        Path.mkdir(flush_dir, exist_ok=True)
+        Path.mkdir(write_dir, exist_ok=True)
+
```
andresito00, line 199: This the perfect situation to decomp with a simple helper function. No need for anything more complicated :).

Also, please make sure to use the `os`/`Path` functionality everywhere possible to make sure paths are delimited appropriately.

```python
def _test_flush(child_resource: Resource, flush_dir: str, write_dir: str, test_identify: bool):
    suffix = "identified" if test_identify else "unidentified"
    do_pack = not test_identify
    try:
        out_file_name = f"{child_resource.get_id().hex()}-{suffix}"
        want.add(out_file_name)
        
        await child_resource.flush_to_disk(os.path.join(flush_dir, out_file_name)")
        with open(Path(write_dir, out_file_name), "wb") as f:
            await child_resource.write_to(f, pack=do_pack)
    
    except Exception as e:
        print(f"Could not flush: {e}")


# ...
         for child_resource in await root_resource.get_children():
             _test_flush(child_resource, flush_dir, write_dir, test_identify=False)
             _test_flush(child_resource, flush_dir, write_dir, test_identify=True)
```

See below for my comment on the try/except.

ofrak_core/ofrak/resource.py
```diff
@@ -1464,14 +1466,16 @@ async def delete(self):
         self._resource.is_modified = True
         self._resource.is_deleted = True
 
-    async def flush_to_disk(self, path: str):
+    async def flush_to_disk(self, path: str, pack: Optional[bool] = True):
```
andresito00, line 1469: We use `Optional` if we need to support `None`. I defer to @EdwardLarson here, but I don't think `None` makes sense here, really. Same goes for the other function signatures.

Since you are giving the parameter a default value of `True`, I don't think we don't need `Optional`.

andresito00 (changes requested):

ofrak_core/test_ofrak/unit/test_resource.py
```diff
@@ -170,3 +172,61 @@ async def test_get_most_specific_tags(resource: Resource):
...
+                with open(Path(write_dir, out_file_name), "wb") as f:
+                    await child_resource.write_to(f, pack=False)
+
+            except Exception as e:
+                print(f"Could not flush: {e}")
+
+        got_flush = set()
+        got_write = set()
+        [got_flush.add(str(f.name)) for f in Path(flush_dir).iterdir()]
+        [got_write.add(str(f.name)) for f in Path(write_dir).iterdir()]
```
rbs-jacob, line 229: These can be written as set comprehensions.

```suggestion
        got_flush = {str(f.name)) for f in Path(flush_dir).iterdir()}
        got_write = {str(f.name)) for f in Path(write_dir).iterdir()}
```

ofrak_core/test_ofrak/unit/test_resource.py
```diff
@@ -170,3 +172,61 @@ async def test_get_most_specific_tags(resource: Resource):
...
+@pytest.mark.asyncio
+async def test_flush_to_disk_pack(ofrak_context: OFRAKContext):
+    # This test works as long as LZMA fails to pack() and the default is to
+    # pack recursively. The root resource is a LZMA archive in a LZMA archive,
+    # which as of 2022-09-13 will fail to pack and therefore pack_recursively
+    # won't work.
+
+    root_resource = await ofrak_context.create_root_resource(
+        "lzma",
+        b"\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff\x00\x2e\x80\x2d\x00\x00\x7f\xeb\x0b\xed\x6e\xd4\x23\x99\xb4\xf3\x0e\x8d\x71\xaf\xff\xff\xf9\xbe\x40\x00",
```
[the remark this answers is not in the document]
rbs-jacob: Black doesn't know how to break up long strings – this needs to be done manually. There are probably a bunch of long strings elsewhere in the codebase.

In this case, is there a reason to inline the file like this as opposed to LFS tracking a file that can be reused for other tests?

rbs-jacob (changes requested):

ofrak_core/test_ofrak/unit/test_resource.py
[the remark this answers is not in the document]
sjossi: I was inspired by another test that did a similar thing and liked the fact that it's obvious what bytes are operated on, especially when something fails and it can be compared to the output. Not sure how much use we have for an empty LZMA in LZMA, but I don't mind checking in a separate file. If we keep it, I can break it up to fit into 120 chars.

ofrak_core/ofrak/resource.py
sjossi: Good catch, I forgot to remove it after I set the default. Fixed it also in `write_to()`

ofrak_core/test_ofrak/unit/test_resource.py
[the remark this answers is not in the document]
andresito00: I think let's go with a small, lfs-tracked file in an assets/ directory for this one. Thanks @sjossi.

ofrak_core/test_ofrak/unit/test_resource.py
```diff
@@ -170,3 +172,61 @@ async def test_get_most_specific_tags(resource: Resource):
...
+                want.add(out_file_name)
+
+                await child_resource.identify()
+                await child_resource.flush_to_disk(f"{flush_dir}/{out_file_name}", pack=False)
+
+                with open(Path(write_dir, out_file_name), "wb") as f:
+                    await child_resource.write_to(f, pack=False)
+
+            except Exception as e:
+                print(f"Could not flush: {e}")
```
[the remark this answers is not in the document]
sjossi: I'm mocking the packer failure and capture the it much earlier, so we don't rely on creating and counting files and therefore make it more reliable and probably faster

ofrak_core/test_ofrak/unit/test_resource.py
sjossi: obsolete because of the mock

ofrak_core/test_ofrak/unit/test_resource.py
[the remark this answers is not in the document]
sjossi: I'm creating a mock resource, since we don't rely on (or even want to use) LZMA anymore.

sjossi:
Instead of relying on LZMA packer to fail for random reasons, which could be patched in the future, I'm using a `MockFailPacker` that will always throw an exception and test for a "successful failure".

This obviated the use of a tracked file, and we're reducing the use of temporary files and remove the silly counting of files.

EdwardLarson:
What's with all the changes to Dockerstubs and `build_image.py` @sjossi ?

sjossi:
I undid the bad merges since they're not conflicting with master.

ofrak_core/test_ofrak/unit/test_resource.py
```diff
@@ -170,3 +179,37 @@ async def test_get_most_specific_tags(resource: Resource):
...
+
+    root_resource = await ofrak_context.create_root_resource("mock", b"\x00" * 0x100)
+    root_resource.add_tag(MockFile)
+
+    child = await root_resource.create_child((MockFailFile,), data_range=Range(0x10, 0x20))
+    child.add_tag(MockFailFile)
+
+    with BytesIO() as buffer:
+        # should fail because write_to runs pack_recursively and MockFailFile will fail on packing
+        with pytest.raises(ComponentAutoRunFailure):
```
sjossi, line 204: Let me add a proper [Exception catch](https://docs.python.org/3.7/library/exceptions.html#built-in-exceptions
The expression following from must be an exception or None. It will be set as __cause__ on the raised exception.)

EdwardLarson: This the only thing blocking this MR for me.

EdwardLarson (changes requested):

sjossi:
I RTFM'd and am now checking for the expected `MockFailException`

EdwardLarson (approved):

Issue #17, opened 2022-08-19
UF2 unpacker/packer


whyitfor:
**What is the use case for the feature?**
To unpack/pack UF2 file format (https://github.com/microsoft/uf2/).

**Does the feature contain any proprietary information about another company's intellectual property?**
No: https://github.com/microsoft/uf2.

**How would you implement this feature?**
Probably by wrapping https://github.com/microsoft/uf2/blob/master/utils/uf2conv.py .

**Are there any (reasonable) alternative approaches?**
Other uf2 tools probably exist.

**Are you interested in implementing it yourself?**
Maybe :).

whyitfor:
Closed with #31

PR #31, opened 2022-09-16
Uf2 support

sjossi:
**Link to Related Issue(s)**
UF2 unpacker support was requested in #17. 

Implemented the Unpacker and Packer for UF2, using [Microsoft Opencode's UF2 utilities](https://github.com/microsoft/uf2). Their python module is not packaged in pypi, but I created the required setup.{py,cfg} files in a fork, and am asking for it to be merged as discussed in microsoft/uf2#42


ofrak_components/ofrak_components/uf2.py
```diff
@@ -0,0 +1,74 @@
...
+
+    async def pack(self, resource: Resource, config=None):
+        """
+        Pack a resource into a UF2 file
+
+        :param resource:
+        :param config:
+        """
+        uf2_child = await resource.get_only_child()
+        uf2_packed = uf2conv.convert_to_uf2(await uf2_child.get_data())
```
whyitfor, line 62: This uses a default `familyid` of 0x0.

We'll want `Uf2Data` to know the family ID for the binary it corresponds to, and be able to correctly pack it up here.

ofrak_components/ofrak_components/uf2.py
```diff
@@ -0,0 +1,74 @@
...
+        Unpack a UF2 file.
+
+        UF2 files contain blocks of binary data.
+        """
+
+        # TODO: technically there's multiple types of data. At least
+        # investigate what they mean - sj
+
+        resource_data = await resource.get_data()
+        uf2_data = uf2conv.convert_from_uf2(resource_data)
```
whyitfor, line 43: I'm not sure this approach works.

We'll want to extract the `familyid` and `appstartaddr` either as an analyzer or part of unpack step (https://github.com/microsoft/uf2/blob/162acf4e4cb7c8e9ec589c842d277dcb5c43db89/utils/uf2conv.py#L19).

The `familyid`, at a minimum, needs to be saved so that the packer uses it, no?

ofrak_components/ofrak_components_test/test_uf2_component.py
```diff
@@ -0,0 +1,48 @@
...
+
+INITIAL_DATA = b"Hello World"
+EXPECTED_DATA = b"Hello OFRAK" + b"\x00" * 245
+TARGET_UF2_FILE = "hello_uf2.uf2"
+UF2_ENTRY_NAME = "hello_uf2_file"
+
+
+class TestUf2UnpackModifyPack(UnpackModifyPackPattern):
+    async def create_root_resource(self, ofrak_context: OFRAKContext) -> Resource:
+        packed_data = uf2conv.convert_to_uf2(INITIAL_DATA)
```
whyitfor, line 22: I'd recommend using a real test file here -- if I understand correctly this unpacker/packer will not work for any file other than one when `familyid` is 0x0.

whyitfor (changes requested):
Please see comments.

It might end up being easier to use the CLI interface, given the tools' use of global variables.

ofrak_components/ofrak_components/uf2.py
sjossi: I got ahead of myself with the PR, forgot to tag as TODO, so my checks didn't complain. I'll read up on it and either guard or test for it.

sjossi:
Regarding using the CLI. Are you sure that their global variables become a problem? From what I understand they should be exposed in the module just like anything else, so even importing uf2conv should not break. Otherwise even the existing code should have failed, right?

ofrak_components/ofrak_components_test/test_uf2_component.py
sjossi: I'm updating this to use the [Micropython on RP2 Pico UF2 file](https://micropython.org/download/rp2-pico/), specifically the [1.19 release](https://github.com/micropython/micropython/releases/tag/v1.19.1). According to the [micropython license](https://github.com/micropython/micropython/blob/master/LICENSE) there should be no issue to use that file if I understand the MIT license correctly. Can somebody please double check me? Thanks

ofrak_components/ofrak_components/uf2.py
sjossi: Covered by new analyzer

sjossi:
Converting to Draft again, so I can make the Unpacker unpack into a CodeRegion or MemoryRegion.

sjossi:
I'm packing and unpacking fully manually now and return contingent CodeRegions. As discussed, we fail at multiple Family_IDs, but that would be a small change (I almost had it half done, but then simplified to get it done).

Also, currently we don't support any embedded files and further information because it was not in the original scope. But extending the Unpacker should be quite easy, I purposely left the if/else in to make it obvious how and where to implement it.

I had more tests that I ran jankily. I will parametrize the tests and check in a few more sample files (from [https://arcade.makecode.com/](Makecode) and other sources that don't obviously use uf2conv.py.)

sjossi:
Fixed the notebook tests, so all checks pass now. Assinging Edward for code review

sjossi:
@EdwardLarson Thanks again for the review, I implemented all fixes. Please have a quick look.

EdwardLarson (approved):

PR #34, opened 2022-09-20
Maintenance/free space improvements

thejoelpatrol:
Fixing incorrect end of freed range


andresito00:
Closing. The stub bytes are assumed to reside _within_ the provided subrange. 

The proposed change would extend the freed range beyond what should be intended by the user of the Modifier.

PR #35, opened 2022-09-21
Feature/flash

jaredgonzales:
**Link to Related Issue(s)**
N/A

Adds a component for unpacking/packing a flash dump with OOB data included. To aid this, adds another component for implementing Error Correcting Codes (ECC) for use in other components. This also includes an additional python dependency requirement.

@andresito00


ofrak_components/ofrak_components/flash.py
```diff
@@ -0,0 +1,695 @@
...
+    async def unpack(self, resource: Resource, config: FlashConfig = FlashConfig):
+        ecc_config: FlashEccConfig = config.ecc_config
+        if ecc_config is None:
+            UnpackerError("Tried unpacking FlashEccProtectedResource without FlashEccConfig")
+
+        data = await resource.get_data()
+        data_len = len(data)
+        magic = ecc_config.ecc_magic
+        if magic is not None:
+            ecc_magic_offset = data.find(magic)
```
rbs-jacob, line 353: `.find` returns `-1` if the argument is not found. Should this case be handled?

ofrak_components/ofrak_components/flash.py
```diff
@@ -0,0 +1,695 @@
...
+    children = (
+        FlashEccResource,
+        FlashHeaderBlock,
+        FlashBlock,
+        FlashTailBlock,
+        FlashLogicalDataResource,
+        FlashLogicalEccResource,
+    )
+
+    async def unpack(self, resource: Resource, config: FlashConfig = FlashConfig):
```
rbs-jacob, line 344: This function is pretty huge, and it seems like there are some discrete bits of functionality that could be factored out into their own functions. Doing so might make this code a bit more readable and maintainable.

rbs-jacob (commented):
I haven't looked closely at this, but just glancing through I have a couple of minor comments.

ofrak_components/setup.py
```diff
@@ -42,6 +42,7 @@ def run(self):
         "pycdlib==1.12.0",
         "python-magic",
         "synthol~=0.1.1",
+        "reedsolo",
```
rbs-jacob, line 45: You should probably pin the version of this dependency so an upgraded version doesn't break the tests.

ofrak_components/ofrak_components/ecc/reedsolomon.py
```diff
@@ -0,0 +1,40 @@
+import reedsolo as rs
+from ofrak_components.ecc.abstract import EccAlgorithm, EccError
+
+
+class ReedSolomon(EccAlgorithm):
+    RSC = rs.RSCodec
```
rbs-jacob, line 6: Is it intentional to define this as a class attribute? It doesn't look like it's ever used as a class attribute.

```suggestion
```

rbs-jacob (changes requested):
Below are some (mostly minor) initial comments. I haven't yet reviewed the core functionality of these components, but will wait to do so until the main unpacker has been broken up a little bit.

Overall the structure looks very flexible in a great way! Can't wait to put this to good use.

ofrak_components/ofrak_components/flash.py
```diff
@@ -0,0 +1,777 @@
...
+    Packs the FlashResource into binary and cleans up logical data representations
+    """
+
+    id = b"FlashResourcePacker"
+    targets = (FlashResource,)
+
+    async def pack(self, resource: Resource, config=None):
+        """
+        :param resource:
+        """
```
rbs-jacob, line 523: If there is no docstring, the docs will use one for the superclass. But if there is a docstring, it will override the superclass one. 

This should either be filled out with a docstring or removed.

```suggestion
```

rbs-jacob (changes requested):
Some pretty minor comments. Overall, looks good! Once these are wrapped up, it should be good to merge.

rbs-jacob (approved):
Thanks for the quick turnaround on applying the fixes, looks good!

PR #38, opened 2022-09-28
Decouple free space creation from ISA

### LINKED ISSUE #37: FreeSpaceModifier and PartialFreeSpace Modifier Improvements
1) The modifiers should be consolidated.

2) The modifiers should _not_ require knowledge of instruction set architecture. We should opt for an ISA-agnostic `bytes`-based contract if any custom code should be used in replacement.

- `ofrak_core/ofrak/core/free_space.py`
- `ofrak_core/ofrak/service/job_service.py`
- `ofrak_core/test_ofrak/unit/resource_view/test_view.py`

Yes

Yes

andresito00:
**Link to Related Issue(s)**
https://github.com/redballoonsecurity/ofrak/issues/37

- Consolidates the free space modifiers.
- Removes dependency on ISA, especially dependence on NOP instruction size
- Opts for stub bytes parameter

@EdwardLarson @whyitfor


ofrak_core/ofrak/core/free_space.py
```diff
@@ -387,176 +382,97 @@ class FreeSpaceModifier(Modifier[FreeSpaceModifierConfig]):
...
+                await parent_mr_view.resource.create_child_from_view(
+                    ComplexBlock(stub_range.start, stub_range.end, name),
+                    data_range=stub_offset_range,
+                    data=config.stub,
+                )
+            elif is_bb:
+                await parent_mr_view.resource.create_child_from_view(
+                    BasicBlock(
+                        stub_range.start, stub_range.length(), InstructionSetMode.NONE, True, None
+                    ),
```
andresito00, line 469: I question whether or not `BasicBlock` should need to know the `InstructionSetMode`, but imo it's out of scope of this set of changes.

whyitfor:
A couple thoughts:
1. I don't understand the motivation for combining the `FreeSpaceModifier` and the `PartialFreeSpaceModifier` into one modifier. These seem to be two distinct use cases : The `FreeSpaceModifier` frees an entire resource, while the `PartialFreeSpaceModifier` frees a range within the resource. I'd prefer keeping these use cases as separate modifiers. A modifier is essentially a function: it should do one thing, do it well, and do it only.
2. This PR actually adds a third behavior with the `stubs` argument. It's unclear from the code what the use case for this is (it appears that it is supposed to targets ComplexBlocks or BasicBlocks, but the docs do not say this and this is not enforced in code). Either way, this feels like a separate modifier.
3. As a user, I'd like to be able to specify what the free space data is. I'd appreciate a default value of b'\x00' that I could override.
 
Another observation (which is beyond the scope of this PR): why do we delete the resource and create a new child? I wonder if this modifier could modify the bytes of the resource and change any resource attributes that would need be changed as a result. `Resource.create_child_from_view` is an expensive function call...

andresito00:
Per @whyitfor's feedback, I'll remove creation of new children other than `FreeSpace` (which is currently required for fancier `Allocatable` functionality).

Not happy with decoupling the `FreeSpaceModifier` and `PartialFreeSpaceModifier`, but also not married to it.

I'll rename the PR accordingly.

ofrak_core/ofrak/core/free_space.py
```diff
@@ -370,15 +358,50 @@ async def _get_partially_overlapping_resources(
...
 
     :var permissions: memory permissions to give the created free space.
+    :var fill: bytes to fill the free space with
     """
 
     permissions: MemoryPermissions
+    fill: Optional[bytes] = None
 
 
 class FreeSpaceModifier(Modifier[FreeSpaceModifierConfig]):
```
whyitfor, line 407: The Docstring (line 41) also should be updated, something like:

_...by replacing its data with fill or null bytes and tagging..._

ofrak_core/ofrak/core/free_space.py
```diff
@@ -460,23 +435,21 @@ async def modify(self, resource: Resource, config: FreeSpaceModifierConfig):
...
+    :var fill: bytes to fill the free space with
     """
 
+    permissions: MemoryPermissions
     range_to_remove: Range
-    mode: InstructionSetMode = InstructionSetMode.NONE
+    fill: Optional[bytes] = None
 
 
 class PartialFreeSpaceModifier(Modifier[PartialFreeSpaceModifierConfig]):
```
whyitfor, line 455: This docstring also references NOP instructions

whyitfor (approved):
Looks good! 

I appreciate the initiative to decouple free space creation from inserting NOP instructions -- inserting NOPs as default free space can cause hard-to-trace bugs in many cases.

ofrak_core/ofrak/core/free_space.py
```diff
@@ -370,15 +358,50 @@ async def _get_partially_overlapping_resources(
...
+            ),
+        )
+    )
+    for possible_overlapping_child in overlap_resources:
+        if (
+            possible_overlapping_child.virtual_address > freed_range.start
+            or possible_overlapping_child.end_vaddr() > freed_range.start
+        ):
+            await possible_overlapping_child.resource.delete()
+            await possible_overlapping_child.resource.save()
```
EdwardLarson, line 381: All modified/deleted resources are saved after the modifier exits, and this is done before patches are applied, so why is this deleting every resource individually? If performance is a concern as the comment above implies, this will not help.

ofrak_core/ofrak/core/free_space.py
```diff
@@ -370,15 +358,50 @@ async def _get_partially_overlapping_resources(
...
+                ),
+            ),
+        )
+    )
+    for possible_overlapping_child in overlap_resources:
+        if (
+            possible_overlapping_child.virtual_address > freed_range.start
+            or possible_overlapping_child.end_vaddr() > freed_range.start
+        ):
+            await possible_overlapping_child.resource.delete()
```
EdwardLarson, line 380:
```suggestion
                attribute_filters=(
                    ResourceAttributeRangeFilter(MemoryRegion.VirtualAddress, max=freed_range.end),
                    ResourceAttributeRangeFilter(MemoryRegion.EndVaddr, min=freed_range.start),
                ),
            ),
        )
    )
    for overlapping_child in overlap_resources:
        await possible_overlapping_child.resource.delete()
```

Indexable attributes have improved a bit since the original implementation. The overlapping children can now be found with one query.

EdwardLarson (changes requested):
Removing the NOP bytes makes a lot of sense and removes significant complexity + prerequisites. Good change!

ofrak_core/ofrak/core/free_space.py
andresito00: While using this Partial modifier in more complex operations, not having this additional `save` was causing problems. For now, I propose allowing this workaround and we can further explore the underlying root cause of that failure, open another issue, and pull request as needed. We can link to this PR if desired.

ofrak_core/ofrak/core/free_space.py
andresito00: Testing the suggested change in my application of this modifier, will take it if results match.

EdwardLarson (approved):

PR #40, opened 2022-09-30
Strip DataServiceInterface and rewrite DataService

EdwardLarson:
**Link to Related Issue(s)**

Strips a number of unused features and methods from the DataService.

Includes a new DataService implementation (call `NewDataService` for now) which is simpler and should be easier to maintain.

@whyitfor, @rbs-jacob, and @andresito00 , do you agree that this new version is easier to understand? Especially look at the `_apply_patches_to_root` method and the `_DataRoot` + `_Waypoint` model of data mapping. Are these easy to understand?


ofrak_core/ofrak/service/new_data_service.py
```diff
@@ -0,0 +1,409 @@
...
+    offset: int
+    models_starting: Set[DataId]
+    models_ending: Set[DataId]
+    # models_overlapping: Set[DataId]
+
+    def is_empty(self) -> bool:
+        return not self.models_starting and not self.models_ending
+
+    def validate(self):
+        # assert self.models_overlapping.issubset(self.models_overlapping)
```
whyitfor, line 276: Should these be removed?

ofrak_core/ofrak/service/new_data_service.py
```diff
@@ -0,0 +1,409 @@
...
+        ]
+
+
+# Helper classes
+@dataclass
+class _Waypoint:
+    offset: int
+    models_starting: Set[DataId]
+    models_ending: Set[DataId]
+    # models_overlapping: Set[DataId]
```
whyitfor, line 270: Can this be removed?

ofrak_core/ofrak/service/new_data_service.py
```diff
@@ -0,0 +1,409 @@
...
+        self.model: DataModel = model
+        self.data = data
+        self._waypoints: Dict[int, _Waypoint] = dict()
+        self._waypoint_offsets: SortedList[int] = SortedList()
+        self._children: Dict[DataId, DataModel] = dict()
+
+    def get_children(self) -> Iterable[DataModel]:
+        return self._children.values()
+
+    def add_new_model(self, model: DataModel):
```
whyitfor, line 296: Could this method be called `add_mapped_child`?

ofrak_core/ofrak/service/new_data_service.py
```diff
@@ -0,0 +1,409 @@
...
+        ending_waypoint = self._waypoints.get(ending_offset)
+        if ending_waypoint is None:
+            ending_waypoint = _Waypoint(ending_offset, set(), set())
+            self._waypoints[ending_offset] = ending_waypoint
+            self._waypoint_offsets.add(ending_offset)
+        ending_waypoint.models_ending.add(model.id)
+
+        self._children[model.id] = model
+
+    def delete_model(self, model: DataModel):
```
whyitfor, line 321: Similarly, `delete_mapped_child`?

ofrak_core/ofrak/service/data_service_i.py
```diff
@@ -35,42 +32,16 @@ async def get_by_ids(self, data_ids: Iterable[bytes]) -> Iterable[DataModel]:
...
-        raise NotImplementedError()
-
-    @abstractmethod
-    async def set_alignment(self, data_id: bytes, alignment: int):
+    async def get_range_within_other(self, data_id: bytes, within_data_id: bytes) -> Range:
         raise NotImplementedError()
 
     @abstractmethod
-    async def set_overlaps_enabled(self, data_id: bytes, enable_overlaps: bool):
+    async def get_data(self, data_id: bytes, data_range: Range = None) -> bytes:
```
whyitfor, line 44: `data_range: Optional[Range] = None`?

ofrak_core/ofrak/service/new_data_service.py
```diff
@@ -0,0 +1,409 @@
...
+            return r
+        else:
+            root = self._get_root_by_id(model.root_id)
+            absolute_range = r.translate(model.range.start)
+            if absolute_range.end > root.model.range.end:
+                raise OutOfBoundError(
+                    f"The requested range {r} of model {model.id.hex()} is outside the "
+                    f"root's range {root.model.range}"
+                )
+            return absolute_range
```
whyitfor, line 67: I would opt to moving these helper functions to the bottom.

As a reader, I'd like to read through the API first. `_get_by_id` should be pretty explanatory: if I need to dig into its implementation details, I can then look at the bottom of the class.

ofrak_core/ofrak/service/new_data_service.py
```diff
@@ -0,0 +1,409 @@
...
+                    f"area of {patch.data_id.hex()} and cannot be applied!"
+                )
+            target_model = self._get_by_id(patch.data_id)
+
+            raw_absolute_patch_range = self._get_absolute_range(target_model, patch.range)
+            raw_patch_ranges_in_root.append(raw_absolute_patch_range)
+            absolute_patch_range = resize_tracker.translate_range(raw_absolute_patch_range)
+            models_intersecting_patch_range = root.get_children_with_boundaries_intersecting_range(
+                absolute_patch_range
+            )
```
whyitfor, line 222: What is the difference between a `raw_absolute_patch_range` and an `absolute_patch_range`?

whyitfor (commented):
This seems to be coming along well. It is much more concise!

See some comments. I think the docstrings you are planning on adding will definitely help with readability.

ofrak_core/ofrak/service/new_data_service.py
EdwardLarson: `absolute_patch_range` is translated according to any shifts in the data that may have happened as a result of earlier patches. For example, Path 1 resizes range (0,4) to have length 2 (size diff = -2), then a patch to range (6, 10) should be shifted to be applied to range (4, 8)

ofrak_core/ofrak/model/data_model.py
```diff
@@ -43,26 +43,19 @@ def __repr__(self):
...
 class DataPatchesResult:
     data_id: bytes
     patches: List[DataPatchResult]
 
-    def get_size_change(self):
-        return sum(r.size_change for r in self.patches)
 
-
-@dataclass
+@dataclass(unsafe_hash=True)
```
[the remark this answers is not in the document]
EdwardLarson: It makes the models hashable so they can be added to sets. This only is used in a scope where the models cannot change while the set is active, but since the models are an external-facing data structure, it might not be safe to include. I'll work it out of there.

ofrak_core/ofrak/service/new_data_service.py
whyitfor: So if I understand correctly:

- `raw_absolute_patch_range` is the range of the patch in the root data
- `absolute_patch_range` is the translated range if any shifts have occurred

What about renaming:
- `_get_absolute_range` -> `_get_range_in_root_data`
- `raw_absolute_patch_range -> `patch_range_in_root`
- `absolute_patch_range` -> `translated_patch_range_in_root` (or `shifted_...` ?)

ofrak_core/ofrak/service/data_service.py
```diff
@@ -1,1464 +1,402 @@
...
+        data_id: DataId,
+        parent_id: DataId,
+        mapped_range: Range,
+    ) -> DataModel:
+        if data_id in self._model_store:
+            raise AlreadyExistError(f"A model with {data_id.hex()} already exists!")
+
+        parent_model = self._get_by_id(parent_id)
+        mapped_range = mapped_range.translate(parent_model.range.start)
+        if mapped_range.end > parent_model.range.end:
```
whyitfor, line 49: I find all of these mapped ranges confusing. It took me a while to figure out why the translation here is necessary.

The "mapped_range" passed to this method is the range of the data in the parent's data.

This line:
```
mapped_range = mapped_range.translate(parent_model.range.start)
```

Seems to then get the range of the data mapped into the root data (which may or may not be the parent.

What do you think of
- Renaming the parameter for `create_mapped` to `range_in_parent`
- Changing the following:
```
mapped_range = mapped_range.translate(parent_model.range.start)
```
to
```
range_in_root = range_in_parent.translate(parent_model.range.start)
```
?

whyitfor (commented):
Looks good. I left a few comments.

ofrak_core/ofrak/service/data_service.py
```diff
@@ -1,1464 +1,402 @@
...
-        if unmapped_range_index > 0:
-            unmapped_range_left: Optional[Range]
-            _, unmapped_range_left = self._unmapped_ranges[unmapped_range_index - 1]
-            if position < unmapped_range_left.end:
-                return unmapped_range_left
-            elif unmapped_range_left.end != position:
-                unmapped_range_left = None
-        else:
-            unmapped_range_left = None
+    async def create_root(self, data_id: DataId, data: bytes) -> DataModel:
```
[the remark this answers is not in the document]
EdwardLarson: In my opinion having a type like `DataId` is more cognitive load that isn't worth it in the interface, but does make it easier to follow in the implementation. That's why I split it so that is only exists in the implementation, and that's where it's actually useful.

I think my comment at the definition of `DataId` is sufficient to explain that it means the same thing as `bytes`, I don't think it should be confusing.

ofrak_core/ofrak/model/data_model.py
```diff
@@ -43,26 +35,29 @@ def __repr__(self):
...
 class DataPatchesResult:
     data_id: bytes
     patches: List[DataPatchResult]
 
-    def get_size_change(self):
-        return sum(r.size_change for r in self.patches)
-
 
 @dataclass
 class DataModel:
```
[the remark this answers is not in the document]
EdwardLarson: Done, and but I somewhat cheated by deleting several of the classes in this file since they were unused and basically unnecessary.

EdwardLarson:
By the way, I did the profiling that we discussed offline, and for a recursive_unpack on busybox that takes 15 min, the new data service was faster overall by a few seconds, which isn't that significant. But what's notable is that abut 16s less was spent in the new data service compared to the old one, out of abut 30s spent in the data service total (originally). So I don't know if I'd call it a 50% performance improvement, but it's certainly not a drop, and it isn't going to make the data service the bottleneck.

ofrak_core/ofrak/service/data_service.py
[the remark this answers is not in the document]
whyitfor: Ok

whyitfor (approved):
Looks good! Only outstanding thing seems to be updating ofrak test coverage figure from 85 to 84

PR #50, opened 2022-10-04
Increase Function Test Coverage for `ofrak.service.*`

whyitfor:
**Link to Related Issue(s)**
N/A

This PR increases the test coverage for all services but the `DataService` to 100% function coverage.

- ofrak_core's test target runs all ofrak tests to report a combined functional coverage percentage
- missing `ofrak_component` imports are ignored by `mypy`
- unused, untested methods from `AssemblerServiceInterface` are removed
- unused, untested `_ComponentAutoRunResult` is deleted
- tests added for `SerializedError`
- full function coverage reached for `ResourceService`

With these changes, function coverage of ofrak tests is 78%.


ofrak_core/ofrak/service/resource_service.py
```diff
@@ -33,8 +33,8 @@
 
 
 class LowValue:
-    def __lt__(self, other):
-        return True
+    def __gt__(self, other):
+        return False
```
EdwardLarson, line 37:
```suggestion
    def __gt__(self, other):
        return False
```

```suggestion
    def __lt__(self, other):
        return True
```

I think we should keep this to `__lt__` as that is the default comparison method, and `<` is the only operator used in `bisect_left`/`bisect_right` where `LowValue` is used. Is there a compelling reason to change this?

ofrak_core/ofrak/service/resource_service.py
```diff
@@ -218,21 +218,27 @@ class ResourceSortLogic(Generic[T], ABC):
     def has_effect(self) -> bool:
         raise NotImplementedError()
 
+    @abstractmethod
```
EdwardLarson, line 221: If 1/2 implementations is allowed to not implement this, is it really an "abstract method"? It seems more like a default implementation. Perhaps just add the `#pragma: no cover` to these. What do you think?

EdwardLarson (changes requested):

ofrak_core/ofrak/service/resource_service.py
whyitfor: Function coverage revealed that the `__lt__` function never gets run, which is why I switched it.

`LOW_VALUE` is used as such:
```
min_index = bisect.bisect_left(self.index, (self.min, LOW_VALUE))
```

Inside of `bisect_left` this ends up being:
```
a: ResourceNode
b: LowValue

a < b
```

Since `ResourceNode` was:
```
    def __lt__(self, other):
        if not isinstance(other, ResourceNode):
            return False
        return self.model.id < other.model.id
```

The `LowValue.__lt__` was never evaluated. This PR changes `ResourceNode.__lt__` to:

```
    def __lt__(self, other):
        if not isinstance(other, ResourceNode):
            return NotImplemented
        return self.model.id < other.model.id
```

Returning `NotImplemented` changes the expression Python evaluates:
```
a < b
```

to:

```
b > a
```

since Python falls back to the the reflected operation (https://docs.python.org/3/library/constants.html#NotImplemented).

This is why I changed it from `__lt__` to __gt__`

ofrak_core/ofrak/service/resource_service.py
whyitfor: I considered doing it this way. I decided against it because they are not default implementations. If they were default implementations, then they should probably be tested (they are not) or invoked. 

Moreover, I don't see any other future implementation of `ResourceSortLogic` beyond `NullResourceSortLogic` using these as "default implementations" -- most of them would probably implement each of those interfaces, so they really are abstract methods that `NullResourceSortLogic` just doesn't need to implement.

Because of this, I think the above solution makes the most sense. As an alternative, I think deleting these methods would also work.

Let me know how you want me to proceed.

ofrak_core/ofrak/service/resource_service.py
EdwardLarson: > Returning NotImplemented changes the expression Python evaluates:
> a < b
> to:
> b > a

Woah! That's kind of cool. Makes sense.

EdwardLarson (approved):

PR #68, opened 2022-10-07
Angr in GUI

EdwardLarson:
**Link to Related Issue(s)**
N/A

* Add an option to set Angr as the disassembler backend when starting the GUI
* Instead of Ghidra being the default backend if none is provided, no backend is used (no disassembly is possible, but this allows the GUI server to be run in an image/install that only has core OFRAK + frontend with no extra dependencies).
* Add a serializer class to "catch" the Angr analysis object so that it is not erroneously serialized when communicating with the frontend. This serializer doesn't actually correctly serialize the object, but instead always serializes it to `None` and raises a hard error on deserializing, as saving disassembler backend metadata is not currently supported.
* A few updates/improvements for the GUI development rules

@rbs-jacob

### DIFF
--- a/frontend/backend/ofrak_server.Dockerfile
+++ b/frontend/backend/ofrak_server.Dockerfile
@@ -6,6 +6,5 @@ COPY $BACKEND_DIR/ofrak_server.py /ofrak_server.py
 RUN python3 -m pip install aiohttp~=3.8.1
 
 ENTRYPOINT python3 -m ofrak_ghidra.server start \
-  & mkdocs serve --dev-addr 0.0.0.0:8000 \
   & python3 /ofrak_server.py 0.0.0.0 8877 \
   & sleep infinity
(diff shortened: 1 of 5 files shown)


frontend/backend/ofrak_server.Dockerfile
```diff
@@ -1,12 +1,11 @@
 FROM redballoonsecurity/ofrak/dev:latest
 
-COPY ofrak_server.py /ofrak_server.py
-COPY docs /docs
-COPY mkdocs.yml /mkdocs.yml
```
rbs-jacob, line 5: Why remove the docs from the development Dockerfile?

rbs-jacob: Moreover, if you remove the docs from the image, the `mkdocs serve` run in the entrypoint probably won't work?

EdwardLarson: Clicking on a tag in the frontend takes me to the public ofrak.com docs anyway, so the dev server hosted docs are basically unused. If we were testing the documentation build here, we should have a dedicated server build for that (in the `docs` directory).

EdwardLarson: And you're right about the mkdocs serve commend, I mean to remove that and could have sworn I had. Definitely removed now.

rbs-jacob (approved):

PR #70, opened 2022-10-07
GUI vaddr search

EdwardLarson:
**Link to Related Issue(s)**
N/A

Adds a search functionality in the GUI, accessed from the toolbar. With a resource selected, one can search its descendants for resources with a `VirtualAddress` matching a specific value or within a range. These are displayed in the search view; each one can be selected as usual, and after going back to the main view, it remains selected.

@rbs-jacob


frontend/src/SearchView.svelte
```diff
@@ -0,0 +1,243 @@
...
+    height: 0.45em;
+    transform: scale(0);
+    transition: 120ms transform ease-in-out;
+    box-shadow: inset 1em 1em var(--main-fg-color);
+  }
+
+  input[type="checkbox"]:checked::before {
+    transform: scale(1);
+  }
+  */
```
rbs-jacob, line 97: This commented-out CSS can be removed. 

Aside: on the OFRAK site, I made consistently-styled, cross-platform, cross-browser, accessible checkboxes. I've yet to port them to the GUI, but this would be a good thing to make a new issue for. Specifically, we'll want a "checkbox" Svelte component. 

For reference, see [the `ofrak.com` stylesheet](https://ofrak.com/style.css):

``` css
/* Taken from https://ofrak.com/style.css */

input[type="checkbox"] {
  margin: 0 1ch;
}

label.checkbox {
  flex-direction: row;
  align-items: center;
  max-width: min-content;
}

label.checkbox input[type="checkbox"] {
  display: none;
}

label.checkbox input[type="checkbox"] + span {
  width: 1em;
  height: 1em;
  background: var(--main-bg-color);
  color: var(--main-fg-color);
  border: 1px solid var(--main-fg-color);
  margin-left: 2ch;
  box-sizing: content-box;
}

label.checkbox input[type="checkbox"]:checked + span::after {
  content: "";
  display: block;
  position: relative;
  top: 0;
  bottom: 0;
  margin: 0.25em;
  width: 0.5em;
  height: 0.5em;
  color: var(--main-fg-color);
  background: var(--main-fg-color);
  border: none;
}
```

This CSS is associated with HTML like the following:

``` html
<label class="checkbox">
  Sign Up for Mailing List
  <input type="checkbox" />
  <span></span>
</label>

```

The checkbox svelte component will pretty much only include this HTML and CSS and a slot for the label, as well as a reactive variable with the state of whether it's checked or not.

frontend/backend/ofrak_server.py
```diff
@@ -330,6 +338,37 @@ async def delete_comment(self, request: Request) -> Response:
...
+                )
+            else:
+                matching_resources = await resource.get_descendants(
+                    r_filter=ResourceFilter(
+                        attribute_filters=(
+                            ResourceAttributeValueFilter(Addressable.VirtualAddress, vaddr_start),
+                        )
+                    ),
+                    r_sort=ResourceSort(Addressable.VirtualAddress),
+                )
```
rbs-jacob, line 367: In my opinion, this is a little bit easier to read:

```suggestion
            if vaddr_end is not None:
                filter = ResourceAttributeRangeFilter(Addressable.VirtualAddress, vaddr_start, vaddr_end)
            else:
                filter = ResourceAttributeValueFilter(Addressable.VirtualAddress, vaddr_start),
            matching_resources = await resource.get_descendants(
                r_filter=ResourceFilter(attribute_filters=(filter,)),
                r_sort=ResourceSort(Addressable.VirtualAddress),
            )
```

frontend/src/SearchView.svelte
```diff
@@ -0,0 +1,243 @@
...
+      } catch (err) {
+        try {
+          errorMessage = JSON.parse(err.message).message;
+        } catch (_) {
+          errorMessage = err.message;
+        }
+      }
+    }
+  }
+  async function leave_view() {
```
rbs-jacob, line 182:
```suggestion
  }
  
  async function leave_view() {
```

frontend/src/SearchView.svelte
```diff
@@ -0,0 +1,243 @@
...
+      }
+    }
+    modifierView = undefined;
+  }
+</script>
+
+<div class="container">
+  <div class="inputs">
+    <p>
+      Searching for descendants of {searchTarget.get_id()} whose virtual address
```
rbs-jacob, line 196: It may be more helpful to have the resource's label instead of its ID here? Perhaps both?

rbs-jacob (changes requested):
Looks good! Very minor changes recommended.

frontend/src/SearchView.svelte
EdwardLarson: I specifically chose not to use the label because it isn't unique to the resource. I wanted it to be clear that the same resource is being searched. I don't think having the label here is helpful as the user should already know what resource they are looking at, having selected it in the main tree view.

frontend/src/ResourceTreeToolbar.svelte
```diff
@@ -187,6 +188,14 @@
           modifierView = CommentView;
         },
       },
+
+      {
+        text: "Search",
+        iconUrl: "/icons/identify.svg",
```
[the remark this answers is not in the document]
EdwardLarson: Yeah I consider this out of scope.

frontend/backend/ofrak_server.py
EdwardLarson: I applied a similar change, with different name for `filter` so it doesn't overlap the builtin

frontend/src/SearchView.svelte
rbs-jacob: I am going to push back on this a little. If the user misclicks a resource or has one highlighted that they didn't intend, then this doesn't actually provide them with any useful information. While it's true that they could look at the attributes for the highlighted resource, I think having the full hex of the UUID front-and-center is not actually helpful.

If the resource ID is necessary, I think this should have both the resource ID and the label.

EdwardLarson: I can add the label, but arguably the root of the problem is that the GUI does not display any unique identifier for resources at the moment. A lot of the labels overlap ("ElfSection", "ElfSection, CodeRegion", "File" etc.) and the UUID is not displayed anywhere.

rbs-jacob: I think displaying the UUID in the attributes view area would solve both of those problems. Then it's showing whenever you click a resource, and when it's selected to modify. 

I would argue that for regular browsing around, having the UUID display for every resource is overstimulating and mostly uninformative. It's only useful when you want to confirm the selected resource has the correct ID. Otherwise, the consistent order of the child resources allows for a basic level of differentiation between resources with the same labels.

rbs-jacob (approved):

PR #71, opened 2022-10-11
Trim Elf model interface, add tests

whyitfor:
This PR refactors OFRAK's elf resource views and components, removing unnecessary methods and classes and adding function test coverage for the OFRAK Elf API.

Relevant changes:
- Fix a type in `MemoryPermissions` enum such that executable is 1 and read is 4.
- Refactor Elf Resource views:
  - Remove unneeded `get_parent`, `get_elf` methods in Elf resource views.
  - Delete `ElfProgramHeaderPermission`, replacing its usage with `MemoryPermissions`.
  - Simplify `ElfProgramHeader` interface to only include one method, `get_memory_permissions`.
  - Remove unneeded methods of `UnanalyzedElfSegment`.
  - Remove unused methods on `Elf`: `get_sections_after_index`, `get_sections_before_index`, `get_section_header_by_name`, `get_string_section_header`
 
This PR also deletes three analyzers: `ElfSegmentStructureIndexAnalyzer`, `ElfSectionStructureIndexAnalyzer`, and `ElfSymbolStructureIndexAnalyzer`. This was done for several reasons:

- They do not appear to be necessary for current Elf capabilities
- As implemented, they contain bugs:
  - `ElfSegmentStructureIndexAnalyzer` does not actually handle `ElfSegment` correctly
  - `ElfSectionStructureIndexAnalyzer` fails when run on an `ElfSection` with no data

These analyzers can be revisited when OFRAK users have use cases that require these indexes to be 
updated.

**Future Improvements**
#87 was discovered when writing this PR.
Also, see https://github.com/redballoonsecurity/ofrak/pull/71#pullrequestreview-1144735735 for ideas for further improving OFRAK Elf APIs.

@andresito00, @EdwardLarson


whyitfor:
@andresito00, @EdwardLarson, can you review this Draft MR.

Especially review:
- the methods that I propose to delete. It seems that we don't need them; if they become necessary we can always update the model's API
- the methods tagged with `#TODO`. These methods need tests written for them. If any of these should be deleted, please chime in.

Once we agree on path forward, I will add tests.

ofrak_core/ofrak/core/elf/model.py
```diff
@@ -436,36 +405,17 @@ class ElfSectionHeader(ElfSectionStructure):
...
         return self.sh_flags & flag.value > 0
 
-    def get_flags(self) -> Iterable[ElfSectionFlag]:
+    def get_flags(self) -> Iterable[ElfSectionFlag]:  # TODO
         for flag in ElfSectionFlag:
             if self.has_flag(flag):
                 yield flag
 
-    def get_file_range(self) -> Range:
+    def get_file_range(self) -> Range:  # TODO
```
EdwardLarson, line 413: Can delete by unwrapping any usages into this `Range.from_size` call

EdwardLarson (commented):
Mostly lgtm. A general comment on the current ELF implementation is that a lot of the fields of these dataclasses are pointlessly saving the raw unpacked integers when practically all users want the enums they map to. We could get rid of a lot of the `get_(field)` methods by just unpacking the binary into the desirable types in the first place.

But in the interest of time, I imagine that is out of scope.

ofrak_core/ofrak/core/elf/model.py
```diff
@@ -272,42 +272,14 @@ class ElfProgramHeader(ElfSegmentStructure):
...
-        return Range.from_size(self.p_vaddr, self.p_memsz).contains_value(vaddr)
-
-    def get_type(self) -> ElfProgramHeaderType:
-        try:
-            return ElfProgramHeaderType(self.p_type)
-        except ValueError:
-            return ElfProgramHeaderType.UNKNOWN
-
-    def is_writable(self) -> bool:
+    def is_writable(self) -> bool:  # TODO: Write test
```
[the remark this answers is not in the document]
whyitfor: Good point -- I've updated this to return `MemoryPermissions` (and fixed a typo in `MemoryPermissions` along the away.

ofrak_core/ofrak/core/elf/model.py
whyitfor: Added a test so opted to keep this for now.

whyitfor:
@EdwardLarson I think this is ready for review. I've updated the PR description with more details.

EdwardLarson (approved):

Issue #51, opened 2022-10-04
Add ZStandard components


whyitfor:
**What is the use case for the feature?**
ZStandard is a modern compression format known for its speed and high compression ratios.

- <https://github.com/facebook/zstd>
- <http://facebook.github.io/zstd/>

It has a command line tool that can be installed with:

``` bash
# Ubuntu / Debian / OFRAK Docker
sudo apt install zstd
```

Or:

``` bash
# macOS
brew install zstd
```

The packer and unpacker components would probably end up being very similar to the [LZO components](https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_components/ofrak_components/lzo.py) in structure.

There are some example ZStandard-compressed files [here](https://github.com/indygreg/python-build-standalone/releases), but they may not make for great test files because they're fairly large, so it is probably worth finding better (or smaller) ones...

**Does the feature contain any proprietary information about another company's intellectual property?**
No, the format is open-source https://github.com/facebook/zstd/blob/dev/LICENSE.

**How would you implement this feature?**
See above.

**Are there any (reasonable) alternative approaches?**
N/A.

**Are you interested in implementing it yourself?**
No -- this is a great first contributor issue!

thejoelpatrol:
I will take a stab at this enhancement for the super cool and good ofrak project next week.

whyitfor:
Closed by #72.

PR #72, opened 2022-10-13
feature/zstd

thejoelpatrol:
**Link to Related Issue(s)**

https://github.com/redballoonsecurity/ofrak/issues/51

Add zstd packer/unpacker and test

@whyitfor


whyitfor:
@rbs-jacob can you review this?

rbs-jacob:
I have some fixes for this PR, but do not have push access to the branch. 


@thejoelpatrol would it be possible to give this account (@rbs-jacob) push access to your fork (or just to the branch), please?

Thanks!

ofrak_components/ofrak_components/zstd.py
```diff
@@ -0,0 +1,96 @@
...
+    A zstd binary blob.
+    """
+
+    async def get_child(self) -> GenericBinary:
+        return await self.resource.get_only_child_as_view(GenericBinary)
+
+
+@dataclass
+class ZstdPackerConfig(ComponentConfig):
+    compression_level: int
```
rbs-jacob, line 29: Out of scope for this PR, but I believe that packers and unpackers should not use configs. Instead, they should use attributes of the resource when possible. For example the flash components use resource attributes instead of configs.

https://github.com/redballoonsecurity/ofrak/blob/4c9723e23f9b22a7f745d469354174a4f3be0f3e/ofrak_components/ofrak_components/flash.py#L77-L90

https://github.com/redballoonsecurity/ofrak/blob/4c9723e23f9b22a7f745d469354174a4f3be0f3e/ofrak_components/ofrak_components/flash.py#L313-L314

rbs-jacob (changes requested):

rbs-jacob (approved):

ofrak_components/ofrak_components/zstd.py
```diff
@@ -0,0 +1,96 @@
+import subprocess
+import tempfile
+from dataclasses import dataclass
```
rbs-jacob, line 3:
```suggestion
from dataclasses import dataclass
from typing import Optional
```

PR #77, opened 2022-10-20
Component binwalk to 100% test coverage

sjossi:
We're excluding the function from the test coverage because it's not part of the public API. The analyzer test tests this function anyways, but the coverage doesn't see that.


ofrak_components/ofrak_components/binwalk.py
```diff
@@ -39,14 +39,14 @@ async def analyze(self, resource: Resource, config=None) -> BinwalkAttributes:
...
-        for result in module.results:
-            offsets[result.offset] = result.description
-    return offsets
+    @staticmethod
+    def _run_binwalk_on_file(filename):
+        offsets = dict()
+        for module in binwalk.scan(filename, signature=True):
+            for result in module.results:
+                offsets[result.offset] = result.description
+        return offsets
```
whyitfor, line 52: ``
```suggestion
def _run_binwalk_on_file(filename):  # pragma: no cover
    offsets = dict()
    for module in binwalk.scan(filename, signature=True):
        for result in module.results:
            offsets[result.offset] = result.description
    return offsets
```

ofrak_components/ofrak_components/binwalk.py
```diff
@@ -39,14 +39,14 @@ async def analyze(self, resource: Resource, config=None) -> BinwalkAttributes:
             # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
             # overkill here.
             offsets = await asyncio.get_running_loop().run_in_executor(
-                self.pool, _run_binwalk_on_file, temp_file.name
+                self.pool, self._run_binwalk_on_file, temp_file.name
```
whyitfor, line 42:
```suggestion
                self.pool, _run_binwalk_on_file, temp_file.name
```

Issue #92, opened 2022-10-31
Address PEP 585 & BeartypeDecorHintPep585DeprecationWarning


whyitfor:
**What is the issue?**
When `beartype` is installed, OFRAK emits `BeartypeDecorHintPep585DeprecationWarning` warnings on python >= 3.9. See https://github.com/beartype/beartype#pep-585-deprecations for more information.

As the above link explains, the typing module will deprecate some type hints in 2025 or 2026.

**Which files would be affected?**
All files with the deprecated typehints.

**What is the proposed fix?**
https://github.com/beartype/beartype#pep-585-deprecations lists a couple options for addressing this.

For now, users can filter these warnings with:
```
# Do it globally for everyone, whether they want you to or not!
# This is the "Make Users Suffer" option.
from beartype.roar import BeartypeDecorHintPep585DeprecationWarning
from warnings import filterwarnings
filterwarnings("ignore", category=BeartypeDecorHintPep585DeprecationWarning)
```

whyitfor:
#94 addresses the issue for the moment from the user perspective ("what are these annoying warnings I'm getting).

As stated above, a more permanent solution is required.

PR #94, opened 2022-10-31
Move beartype from test requirement to ofrak package requirement

### LINKED ISSUE #91: beartype not included with "pip install ofrak"
OFRAK does not list beartype in its install_requires list: https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/setup.py.

As a result, `pip install ofrak` does not work if beartype is not installed.

```
% python3 -m venv venv
% source venv/bin/activate
(venv) % python3 -m pip install ofrak
(venv) % python3 ex1_simple_string_modification.py
...
    from beartype import beartype
ModuleNotFoundError: No module named 'beartype'
(venv) %
```

`beartype` is not installed with `ofrak`.

Users can work around this by installing `beartype`.

Every time.

Include `beartype` in ofrak's install_requires: https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/setup.py.

whyitfor:
- Make beartype an ofrak package requirement, since it is used by serializers.
- Suppress BeartypeDecorHintPep585DeprecationWarning (pending resolution of #92) so that users can focus on ofrak without getting these error messages.

Closes #91.

**Link to Related Issue(s)**
#91 

No.

### DIFF
--- a/ofrak_core/ofrak/service/serialization/serializers/__init__.py
+++ b/ofrak_core/ofrak/service/serialization/serializers/__init__.py
@@ -1,0 +1,7 @@
+# Choosing the "Make Users Suffer" option pending resolution of
+# https://github.com/redballoonsecurity/ofrak/issues/92.
+# This is actually nicer for OFRAK users, who don't want to see these warnings.
+from beartype.roar import BeartypeDecorHintPep585DeprecationWarning
+from warnings import filterwarnings
+
+filterwarnings("ignore", category=BeartypeDecorHintPep585DeprecationWarning)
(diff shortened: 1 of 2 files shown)


ofrak_core/ofrak/service/serialization/serializers/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Choosing the "Make Users Suffer" option pending resolution of
```
EdwardLarson, line 1: Clean up this comment - no one is suffering. As you said, this is nicer for users because it resolves these warning.

EdwardLarson (changes requested):

ofrak_core/ofrak/service/serialization/serializers/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+# https://github.com/redballoonsecurity/ofrak/issues/92.
+# This is actually nicer for OFRAK users, who don't want to see these warnings.
```
whyitfor, line 3:
```suggestion
# Suppress all BeartypeDecorHintPep585DeprecationWarning warnings pending resolution of
# https://github.com/redballoonsecurity/ofrak/issues/92.
# This is nice for OFRAK users, who don't want to see these warnings.
```

PR #95, opened 2022-11-01
Improve function coverage of ofrak_components from 84 to 88

whyitfor:
- Remove unused `ReedSolomon.check` method
- Skip entropy helper functions that are run by the DataSummaryAnalyzer in an executor (pycoverage does not seem to see them running; they are not part of OFRAK API)
- Remove unused `OpenWrtTrxHeader.get_length` method

Improves function coverage for `ofrak_components` package from 84 to 88

### DIFF
--- a/ofrak_components/ofrak_components/ecc/reedsolomon.py
+++ b/ofrak_components/ofrak_components/ecc/reedsolomon.py
@@ -33,6 +33,3 @@ class ReedSolomon(EccAlgorithm):
             return self.RSC.decode(data=payload)[0]
         except rs.ReedSolomonError:
             raise EccError()
-
-    def check(self, payload: bytes) -> bool:
-        return self.RSC.check(data=payload)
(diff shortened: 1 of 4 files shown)


ofrak_components/ofrak_components/ecc/reedsolomon.py
```diff
@@ -33,6 +33,3 @@ def decode(self, payload: bytes) -> bytes:
             return self.RSC.decode(data=payload)[0]
         except rs.ReedSolomonError:
             raise EccError()
-
-    def check(self, payload: bytes) -> bool:
-        return self.RSC.check(data=payload)
```
rbs-jacob, line 38: @jaredgonzales how important is this code? 

Even though there don't appear to be any uses of it in tests, should it be expected to remain as part of the API? If so, we should add tests for it instead of removing it.

jaredgonzales: It can be removed since it's not used anywhere. That part was added before the flash component fully implemented ecc support.

rbs-jacob (approved):

PR #96, opened 2022-11-01
Upping test coverage for DTB component to 100%

sjossi:
Getting DTB property path with property.get_path() and asserting correct path


whyitfor (approved):
LGTM

PR #99, opened 2022-11-03
Bugfix: Fixes for subtle errors as a result of new Data Service

EdwardLarson:
**Link to Related Issue(s)**

There were two subtle bugs in the new DataService not caught by the data service tests, both having to do with mapped data models: Shifts to consecutive mappings were not updated properly when applying a resizing patch to the entirety of a mapped model, and the get_data_range_within_other did not correctly translate the returned range when comparing two mapped ranges.

* Bugfixes for the two issues described above
* Test cases for those two issues as well
* Add an optional function to the CodeRegionUnpacker tests which is now allowed to be unpacked since it has overlaps

Wyatt or Jacob, I expect Wyatt has more time in this case

### DIFF
--- a/disassemblers/ofrak_capstone/ofrak_capstone/components.py
+++ b/disassemblers/ofrak_capstone/ofrak_capstone/components.py
@@ -1,5 +1,7 @@
 import asyncio
 import logging
+
+from ofrak.component.unpacker import UnpackerError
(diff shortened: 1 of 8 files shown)


whyitfor (approved):
Looks good

PR #102, opened 2022-11-07
Add test coverage for all functions in flash component

sjossi:
Added a test file with MAGIC in the tail block and it's _verify binary.

Found a bug in the untested function that would not find the magic in the tail block. The offset was incremented to early in the while loop which would scan one byte too far to unpack the block and therefore fail.

Finally, bumped function coverage for ofrak_components to 100%

### DIFF
--- a/ofrak_components/ofrak_components/flash.py
+++ b/ofrak_components/ofrak_components/flash.py
@@ -579,9 +579,6 @@ def _get_end_from_magic(attributes: FlashAttributes, start_index: int, data: byt
             )
 
         while 0 <= search_offset <= data_len:
-            import pdb
-
-            pdb.set_trace()
             search_offset = data.find(search_key, search_offset, data_len)
             # search_offset += 1
 
--- a/ofrak_components/ofrak_components/flash.py
+++ b/ofrak_components/ofrak_components/flash.py
@@ -314,8 +314,18 @@ class FlashResourceUnpacker(Unpacker[None]):
             magic = flash_attr.ecc_attributes.ecc_magic
             if magic is not None:
                 ecc_magic_offset = data.find(magic)
-                if ecc_magic_offset != -1:
-                    start_index = ecc_magic_offset
+
+                if ecc_magic_offset == -1:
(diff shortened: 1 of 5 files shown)


whyitfor:
@jaredgonzales, how does this look to you?

jaredgonzales:
Looks good to me. Much cleaner with the fail early and covers some edge cases that were overlooked. I like it

ofrak_components/ofrak_components/flash.py
```diff
@@ -314,8 +314,18 @@ async def unpack(self, resource: Resource, config=None):
             magic = flash_attr.ecc_attributes.ecc_magic
             if magic is not None:
                 ecc_magic_offset = data.find(magic)
-                if ecc_magic_offset != -1:
-                    start_index = ecc_magic_offset
+
+                if ecc_magic_offset == -1:
+                    raise UnpackerError("No header magic found")
+
+                search_offset_in_block = flash_attr.get_field_range_in_block(
```
whyitfor, line 321: Shouldn't this variable be named `field_range_in_block`?

Then line 382 would be `field_range_in_block.start`?

ofrak_components/ofrak_components/flash.py
```diff
@@ -568,32 +584,37 @@ def _get_end_from_magic(attributes: FlashAttributes, start_index: int, data: byt
...
-            search_offset = data.find(search_key, search_offset, data_len)
-            search_offset += 1
+    while 0 <= search_offset <= data_len:
+        search_offset = data.find(search_key, search_offset, data_len)
+        if search_offset == -1:
+            break
 
-            search_offset_in_block = attributes.get_field_range_in_block(
-                attributes.tail_block_format, search_field
+        search_offset_in_block = attributes.get_field_range_in_block(
```
whyitfor, line 604: Should this variable be `field_range_in_block`?

whyitfor (changes requested):
Looks good.

See some minor comments.

whyitfor (approved):
LGTM

PR #103, opened 2022-11-07
Divided PeOptionalHeader into 64 bit and 32 bit as reflected by https…

dannyp303:
…://learn.microsoft.com/en-us/windows/win32/debug/pe-format\#optional-header-windows-specific-fields-image-only

The 64 Bit version of the PE format does not include the "BaseOfData" field, causing the PE component to break on 64 Bit images. Additionally added some documentation on a small discrepancy in the PeOptionalHeader format in which non-Windows images do not have the optional header fields after BaseOfData(32 bit) or BaseOfCode(64 bit) but are populated anyway by the pefile python package. Judging by the pefile package, the header is always the same size, but the fields in the non-Windows images will be zero for most cases and inaccurate, should maybe be removed from the OFRAK info.


ofrak_core/test_ofrak/components/test_pe_component.py
```diff
@@ -7,8 +7,13 @@
...
 
-@pytest.fixture(params=["jumpnbump.exe", "kernel32.dll"])
-async def pe_root_resource(ofrak_context: OFRAKContext, request):
+@pytest.fixture(params=["jumpnbump.exe", "kernel32.dll", "x64kernel32.dll"])
+async def pe_root_resource(ofrak_context: OFRAKContext, request) -> Resource:
+    return await _get_test_resource_from_file_name(ofrak_context, request.param)
+
+
+@pytest.fixture(params=["kernel32.dll", "x64kernel32.dll"])
+async def kernel32_dll(ofrak_context: OFRAKContext, request) -> Resource:
```
whyitfor, line 16:
```suggestion
async def kernel_dll(ofrak_context: OFRAKContext, request) -> Resource:
```

whyitfor: If applied would need to fix up usages of fixture

ofrak_core/ofrak/core/pe/model.py
```diff
@@ -78,6 +82,42 @@ class PeOptionalHeader(ResourceView):
     size_of_uninitialized_data: int
     address_of_entry_point: int
     base_of_code: int
+
+
+@dataclass
+class Pe64POptionalHeader(PeOptionalHeader):
+    """
+    64 bit PE optional header with base_of_data field.
```
whyitfor, line 90: Isn't `base_of_data` not present in 64 optional header?

ofrak_core/ofrak/core/pe/model.py
```diff
@@ -65,7 +65,11 @@ class PeOptionalHeaderMagic(Enum):
 @dataclass
 class PeOptionalHeader(ResourceView):
```
whyitfor, line 66: Since it does not appear that the order matters for these resource views, I wonder if it would be clearer to have all of the default options be defined in `PeOptionalHeader`.

This would make `Pe32OptionalHeader` simply:

```
@dataclass
class Pe32OptionalHeader(PeOptionalHeader):
    base_of_data: int
```

You could even define `Pe64OptionalHeader` as such:
```
@dataclass
class Pe64OptionalHeader(PeOptionalHeadeR):
```

I found it hard to immediately see the difference between 32/64 bit headers with all of the repetition as it is now.

whyitfor (changes requested):
Good improvement!

See some minor suggestions.

ofrak_core/test_ofrak/components/test_pe_component.py
dannyp303: 64bit kernels in windows are still kalled `kernel32.dll` i think this is fine. Hence why i named the new test file `x64_kernel32.dll`.

ofrak_core/ofrak/core/pe/model.py
dannyp303: Yeah, I was split because the original has it in the proper order, which made finding certain fields and double checking the structure nicer. I'll change it to this, I think I'll split out the windows specific fields too, even though its not necessary ATM.

whyitfor (approved):

PR #107, opened 2022-11-09
significant speedups to applying data patches

EdwardLarson:
**Link to Related Issue(s)**
If there are too many patches being applied at once, the data service becomes very slow because the check for overlaps between patches and data models is quadratic-time. Besides this, actually applying many patches to the binary data is very slow because the data is reconstituted many times like: `data[:patch_range_start] + patch_data + data[patch_range_end:]`.

* Much more efficient algorithm for finding data models overlapping with patches
* Use bytearrays when constructing the new data

@whyitfor

### DIFF
--- a/ofrak_core/ofrak/service/data_service.py
+++ b/ofrak_core/ofrak/service/data_service.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Dict, Iterable, List, Optional, Set, Tuple
+from typing import Dict, Iterable, List, Optional, Set, Tuple, Union, cast
 
 from dataclasses import dataclass
 from sortedcontainers import SortedList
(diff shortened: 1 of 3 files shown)


ofrak_core/ofrak/service/data_service.py
```diff
@@ -381,6 +362,75 @@ def get_children_with_boundaries_intersecting_range(self, r: Range) -> List[Data
...
+            )
+            points_of_interest.append(
+                (
+                    waypoint.offset,
+                    POINT_CHILDREN_STARTS,
+                    waypoint.models_starting.difference(waypoint.models_ending),
+                )
+            )
+
+        points_of_interest.sort()
```
whyitfor, line 406: Can you add a note here explaining how this sorting works, for readability?

From reading it, I'm assuming that it sorts by offset first, then point type? having a trail here will help future readers.

whyitfor (approved):
Overall looks good -- see one comment about adding a comment.

Are there any high-level descriptions of the speed improvements that could be added to this PR for posterity?

ofrak_core/ofrak/service/data_service.py
EdwardLarson: Yes - I think I didn't add as many comments as I originally meant to. I'll add a few more.

Issue #53, opened 2022-10-04
Use asyncio interface for subprocess calls


whyitfor:
Many components use `subprocess` to make calls to external programs via the command line. These are currently blocking, non-asynchronous calls.

The [`asyncio.create_subprocess_exec` and `asyncio.create_subprocess_shell` functions](https://docs.python.org/3/library/asyncio-subprocess.html) represent an `asyncio`-compatible API for making subprocess calls that don't block the event loop. We should replace existing `subprocess` calls in OFRAK with async calls via this API.

**Which files would be affected?**
Any file that uses a `subprocess` call. (`ofrak_patch_maker` should not be updated since it is a synchronous package).

**Does the proposed maintenance include non-doc string _functional_ changes to the Python code?**
Yes (see above)

**Are you interested in implementing it yourself?**
No -- this is a great first contributor issue!

marczalik:
@rbs-jacob Should this change be made for `subprocess` calls inside of fixtures and test cases too?

rbs-jacob:
> @rbs-jacob Should this change be made for `subprocess` calls inside of fixtures and test cases too?

That seems less important to me. The main issue is that the `subprocess` calls currently block the event loop to unpack a filesystem, for example. This lag directly affects a user by fully blocking OFRAK concurrency for an inherently asynchronous operation, whereas the async concurrency is less meaningful in the tests. This is because the tests are either run fully serially, or explicitly concurrently using multiprocessing via pytest. 

Does that clarify?

PR #109, opened 2022-11-10
Components mark their external tool dependencies and a CLI to show/filter these

EdwardLarson:
**Link to Related Issue(s)**
Will help with easier installation, as all the Python dependencies can be installed and then list all the external dependencies yet to be installed.

~~This PR is also would accomplish #53 , because external usages are done through `subprocess` and it was trivial to use the `asyncio` subprocess as part of this.~~ Removed as out of scope.

* Adds a class representing a component's external tool dependency, including a method to check if the tool is installed.
* Components in core `ofrak` are updated to mark these external tools. 
* Components also automatically check for a couple specific error types:
  * `FileNotFoundError`s are transformed to `ComponentMissingDependencyError` if the missing file is one of the components's dependencies.
  * `CalledProcessError`s are transformed to `ComponentSubprocessError` which just reports the stdout and stderr better than standard `CalledProcessError` and replaces widely-used `format_called_process_error` utility function.
* A CLI interface for ORFAK has been added to 1) list installed OFRAK packages/components, and 2) report info on installed components' dependencies (such as whether they are installed and what `apt` packages can provide them). 

@andresito00 I think you have a good critical eye for this sort of API, I'd appreciate your feedback on this.
@whyitfor and @rbs-jacob as well.

I expect this will be a common pattern, so we should make sure it feels simple to write and understand.


ofrak_core/ofrak/component/interface.py
```diff
@@ -31,6 +31,11 @@ def get_version(self) -> int:
     def targets(self) -> Tuple[ResourceTag, ...]:
         raise NotImplementedError()
 
+    @classmethod
+    @abstractmethod
+    def get_external_dependencies(cls) -> Tuple[ComponentExternalTool, ...]:
```
whyitfor, line 36: Shouldn't this method be part of the `ComponentInterface`?

ofrak_core/ofrak/model/component_model.py
```diff
@@ -17,6 +19,62 @@ class ComponentConfig:
...
+class ComponentExternalTool:
+    """
+    An external tool or utility (like `zip` or `squashfs`) a component depends on
+    """
+
+    tool: str
+    install_hints: str = None  # e.g. version
+    install_check_arg: str = "--help"
+
+    async def run_tool(self, *args: str) -> str:
```
whyitfor, line 32: I don't think `ComponentExternalTool` needs a `run_tool` method. 

First, it adds a level of indirection that makes the individual components less readable.
Second, it is valuable to let component implementations define how subprocesses are used. For example, see https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_components/ofrak_components/cpio.py#L93, which differs from https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_components/ofrak_components/tar.py#L88.

What I think does make sense is a `get_command` method, that returns `zip`, or `tar` command.

ofrak_core/ofrak/model/component_model.py
```diff
@@ -17,6 +19,62 @@ class ComponentConfig:
...
+            )
+        except FileNotFoundError:
+            return False
+
+        retcode = await proc.wait()
+        if retcode == 0:
+            return True
+        else:
+            return False
+
```
whyitfor, line 76: I'd consider also adding a `install_tool` command.

A couple other things to consider, is that we have different types of tools

1. Command line tools (`pigz`, `tar`, for example)
2. Python tools with custom installation steps (`binwalk`; binary ninja)
3. Tools that are not invoked directly at the command line: ghidra, for example.

Ideally this class would handle all of these cases

ofrak_components/ofrak_components/zip.py
```diff
@@ -61,17 +56,19 @@ class ZipPacker(Packer[None]):
 
     targets = (ZipArchive,)
 
+    ZIP_TOOL = ComponentExternalTool("zip")
```
whyitfor, line 59: What if, instead, components used tools directly. For example:

```
PIGZ_TOOL = ComponentExternalTool("pigz")
```
And then the unpacker/packer both use the tool directly? This would be cleaner for situations where multiple components use the same tool (This is the case with binary ninja components, and even the gzip unpacker/packers).

whyitfor (commented):
See my comments in the code below.

ofrak_components/ofrak_components/zip.py
EdwardLarson: This makes it impractical to tie dependencies to the components that need them, but that may not be important.

whyitfor: Hm, my thinking here was that `get_external_dependencies` could handle that. So, for example of [APK](https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_components/ofrak_components/apk.py), something like

```
UBER_APK_SIGNER = CommandExternalTool(["java", "-jar", "/usr/local/bin/uber-apk-signer.jar"])
APK_TOOL = CommandExternalTool("apktool")

class ApkPacker(Packer):
...
    @classmethod
    def get_expernal_dependencies(cls):
        return UBER_APK_SIGNER, APK_TOOL
```

rbs-jacob: One general question that I have is: why is the "component external tool" management implemented outside of the `Unpacker` and `Packer` classes as opposed to extending them with methods to manage the dependencies? I would expect it's easier to reason about dependencies if they're installed near where they're used. Likewise it might lower the barrier to writing new components.

What are the design justifications for having this additional, separate class?

whyitfor: @rbs-jacob 
It makes sense to make a separate class because of the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). 

Also, as I point out above, some external dependencies are used across multiple components, and components, such as the `ApkPacker`, can also have multiple external dependencies.

ofrak_core/ofrak/model/component_model.py
EdwardLarson: For this reason I want to stay away from a general "install tool" command. For tools that can be easily installed via a package manager like `brew` or `apt`, I've added a standard way to define that.

ofrak_core/ofrak/model/component_model.py
EdwardLarson: I implemented some additions to the API that allow supporting such edge cases pretty trivially.

I get that using shell scripting is intuitive, however it's also probably less portable and more vulnerable. My latest examples show that Python & OFRAK allow us to do the same thing without too much more code.

ofrak_components/ofrak_components/zip.py
EdwardLarson: Implemented pretty much as suggested here, but using a "pseudo-property" like `targets` for components. `external_dependencies` should be a tuple of external tools. The `AbstractComponent` defaults to providing an empty Tuple for this, I think that makes sense: the pro is that all the components without external dependencies don't need to be marked as such, the downside is that it might be easier for a contributor to miss marking external dependencies on a component because they forgot.

ofrak_core/ofrak/model/component_model.py
```diff
@@ -17,6 +20,128 @@ class ComponentConfig:
...
+    required if it is not possible to install via the supported package managers.
+    :ivar install_check_arg: Argument to pass to the tool to check if it can be found and run on
+    the host; defaults to `--help` so that by default, `tool`'s install is checked with:
+        `<tool> --help`
+    """
+
+    tool: str
+    install_packages: Dict[str, Optional[str]] = dataclasses.field(default_factory=dict)
+    install_hints: Optional[str] = None  # e.g. version
+    install_check_arg: str = "--help"
```
andresito00, line 52: A few thoughts:
I don't think we should be prescribing any package manager by default. If we're not providing Python infrastructure to install the tool for the user/dev in the first place, then I really don't like it. I also wouldn't set a default `install_check_arg`.

At a minimum, maybe you can consider removing much of this __post_init__ code. I don't think it's necessary.

Continuing...:

Part of me believes folks should still be allowed to craft a shell command inside of a Component implementation complete with flags, piping, other commands despite certain command injection concerns. I concede, I don't really know how truly "dark" a path that is, but this might make the installation of the tool more important to me than wrapping its execution. 

The `Toolchain` implementations handle this differently than what I suggest above, but I propose that our expectations of those programs (compilers, linkers, assemblers) are much more consistent/limited across the board than what we are hoping to do here. (Though that approach may still need to change with enough diversity.)

Installing dependencies for folks also opens a can of worms, but consider the following:
```
AGNUTOOL_TOOL = ExternalTool(
    linux_install="apt install -y agnutool", 
    windows_install="python3 -m pip install \"agnutool @ git+https://github.com/agnutool@master\"", 
    darwin_install="git clone https://github.com/agnutool@master && make -C agnutool && make install"
) 
```

If the external tool is not found on either first attempted usage or through some other preemptive search, maybe OFRAK could prompt the user:
```
$ Could not find <TOOL>, install with <TOOL_INSTALL_COMMAND>? Y/n: n
$ Provide your own install command? Y/n Y
$ sudo install agnutool --dest-path /some/other/place
```

This is closer to what I'm envisioning when we discuss moving away from a soft docker requirement to use OFRAK.

EdwardLarson: I'm curious what problems you see with the default `--help` `install_check` arg?

ofrak_core/ofrak/model/component_model.py
EdwardLarson: Removed the `run_tool`, we shouldn't needlessly herd contributors to a method arbitrarily.

I did leave my changes to those edge cases qualitatively intact, to remove the `shell=True` param and show that `subprocess` API does allow these very script-like usages, with hardly any extra code.

ofrak_core/ofrak/model/component_model.py
andresito00: I think I suggested the above because I usually veer away from half measures: Install the thing or don't mention installation at all. Having to maintain Python code that only provides install hints felt like a half measure to me.

I reflected a little bit more on this, the goal of pip install-able OFRAK.

Folks looking for an easy-to-use reverse analysis tool will usually find a git repo with a requirements list, but the various components we have complicate this. After all, why should I have to install the "kitchen sink" when I just want to run this one unpacker?

A requirements list in the form of a Component tree may not be the best alternative. So, here we are, suggestion of what to do if a package isn't found is, after further consideration, a real option here and probably the right one for now. Now, as far as how to suggest...
>I really don't want to try to generalize installation on different systems. 

I propose that `REQUIRED_PKG_MANAGERS: ClassVar[Tuple[str, ...]] = ("brew", "apt")` and everything that uses it is trying to do this. 

I'll revise my feedback to the following:
- I'd rather suggest whole command strings or not suggest anything other than "go install cpio for your platform using your favorite package manager or by building from source." It sounds like, of the two, you prefer the latter. This way we don't need to maintain a required package managers list or validation logic and string formatting for that list...
- To me, there is enough potential variance in a preferred `install_check_arg` that I'd rather it be required _and_ I see future potential for unifying that code to be a version check, deviation from which is a potential source of nasty bugs (unless you want to handle that now). That is where I see a need for some extra validation logic.
- In the future, let's consider providing a prompt for the user to provide the command so they don't have to completely leave the OFRAK runtime (out of scope here I think, but I guess this would be done in the MissingError?)

whyitfor:
Regarding the following CLI options:
```
% python3 -m ofrak deps --help
usage: __main__.py deps [-h] [--missing-only] [--dependency-packages {apt,brew}] [--check]

options:
  -h, --help            show this help message and exit
  --missing-only        Only output information for missing dependencies (will check all dependencies)
  --dependency-packages {apt,brew}
                        List names of packages (known to <package_manager>) which provide dependencies required by installed OFRAK packages.
  --check, -c           Check that each dependency is present
```

As an OFRAK user, I only care about the dependency packages for the system I am on. As such I'd prefer for the options to be:
```
% python3 -m ofrak deps --help
usage: __main__.py deps [-h] [--missing-only] [--check]

options:
  -h, --help            show this help message and exit
  --missing-only        Only output information for missing dependencies (will check all dependencies)
  --check, -c           Check that each dependency is present
```

whyitfor:
For the following command:
```
% python3 -m ofrak list --help
usage: __main__.py list [-h] [--packages] [--components] [--dependencies]

options:
  -h, --help          show this help message and exit
  --packages, -p      List installed OFRAK packages
  --components, -c    List installed OFRAK components
  --dependencies, -d  List dependencies of OFRAK components
```

The first thing I did was run the default command which returns nothing:
```
% python3 -m ofrak list
%
```
Can this default to one of the options (perhaps packages)? If for some reason this doesn't make sense, then it should at least show the help option so I don't have to rerun it. 

It is also not intuitive that the command-line flags are options that can be mixed and matched. This took me a while to figure out. If this is the design we end up going with, then the help should make this explicit.

whyitfor:
This command line utility chokes whenever 3-rd party python dependencies that cannot be pip-installed are not installed.

Examples of this are:
- binary ninja Python bindings
- binwalk
- entropy library

As a user, I would like for this CLI tool to be able to tell me:
- Are these dependencies installed?
- If not, how can I install them?

rbs-jacob:
I don't see any docs added in this PR, but it would be helpful to amend the "contributor guide" to explain how to add package dependencies when writing components.

EdwardLarson:
@rbs-jacob 

> I don't see any docs added in this PR, but it would be helpful to amend the "contributor guide" to explain how to add package dependencies when writing components.

I added a doc in the contributor guide and linked to it from each of the individual component guides. Could you review that?

docs/contributor-guide/component/external_tools.md
```diff
@@ -0,0 +1,142 @@
...
+
+It often makes sense to write a component which makes use of third-party tools in one way or 
+another. If these tools are distributed in Python modules, the module can simply be added to the 
+requirements in the OFRAK package's `setup.py` file. But for dependencies on tools which cannot be
+simply installed with `pip`, OFRAK has another mechanism for the making the dependency clear to users
+and easy to install.
+
+### ComponentExternalTool
+
+`ofrak.model.component_model.ComponentExternalTool` is a class that encapsulates the information
```
rbs-jacob, line 11: Should this use the MkDocs cross reference syntax? I don't know if that syntax will work outside of docstrings...

ofrak_core/ofrak/model/component_model.py
EdwardLarson: Per offline conversation together and with users, we did decide not to include installing the tools as a direct responsibility of OFRAK in this PR.


There is not longer a default `install_check_arg`, but it is required. For cases where this is not sufficient, the expectation is to subclass `ComponentExternalTool` and override the `is_tool_installed` method with the appropriate check. There are 2 examples of this and documentation to accompany.

There are install command suggestions in the errors raised when a dependency is missing at runtime.

ofrak_core/ofrak/model/component_model.py
EdwardLarson: We decided not to install tools within OFRAK as part of this PR.

rbs-jacob (commented):
The docs page added looks really clear and very helpful! Love this change!

I'm assuming you ran the docs build locally to confirm that the generated pages look as expected? Pending that, plus one minor comment, this looks very good to me.

docs/contributor-guide/component/external_tools.md
EdwardLarson: Yes it should have, I was using docstring format indeed

rbs-jacob (approved):

EdwardLarson:
@whyitfor 

> As an OFRAK user, I only care about the dependency packages for the system I have OFRAK installed on. 
I shied away from choosing a package manager automatically based on the system. I think it's better for OFRAK to be more explicit (and expect the user to be similarly explicit) about what it is showing. So I kept the same setup where an option tells OFRAK to print the dependency packages for an explicitly given package manager.

> I also would like to run "check" automatically on each dependency, and know what packages require which dependencies.

Checking each component is now the default, though can be disabled with `--no-check`

> Can this default to one of the options (perhaps packages)? 

It now defaults to printing packages and components. The description (printed by `python3 -m ofrak list --help`) makes this clear. There are also fewer arguments to `list` (just `--components` and `--packages`).

> As a user, I would like for this CLI tool to be able to tell me:

> Are these dependencies installed?
> If not, how can I install them?

This is accomplished in the current version, and the instructions are provided by the tool website link:

```
python3 -m ofrak deps
[✓] binwalk
	https://github.com/ReFirmLabs/binwalk
	[BinwalkAnalyzer]
[✓] mksquashfs
	https://github.com/plougher/squashfs-tools.git
	[SquashfsPacker]
[✓] zstd
	http://facebook.github.io/zstd/
	[ZstdPacker, ZstdUnpacker]
[✓] unar
	https://theunarchiver.com/command-line
	[RarUnpacker]
[✓] pigz
	https://zlib.net/pigz/
	[GzipUnpacker, GzipPacker]
...
```

>This might be out of scope of this PR, but what I would really like as a user was a way to step through missing dependencies and try to install with the OFRAK hint.

This can partially done with: `python3 -m ofrak deps --missing-only --packages-for apt | xargs apt install -y`

For dependencies not distributed via package manager (nor via PyPI, remember), we make no attempt to install the tool for the user and just point them to that website. Based on the offline discussions and more feedback from users, this is preferable.

> This would be useful and save time having to look up how to install each individual tool, especially each time I set up my OFRAK environment.

I think if a user is doing this more than once, they should be using Docker, that's essentially what it's for.

docs/contributor-guide/component/external_tools.md
whyitfor, line 11:
```suggestion
[ComponentExternalTool][ofrak.model.component_model.ComponentExternalTool] is a class that encapsulates the information 
```

docs/contributor-guide/component/external_tools.md
```diff
@@ -0,0 +1,142 @@
...
+another. If these tools are distributed in Python modules, the module can simply be added to the 
+requirements in the OFRAK package's `setup.py` file. But for dependencies on tools which cannot be
+simply installed with `pip`, OFRAK has another mechanism for the making the dependency clear to users
+and easy to install.
+
+### ComponentExternalTool
+
+`ofrak.model.component_model.ComponentExternalTool` is a class that encapsulates the information 
+that OFRAK tracks about an external tool which a component depends on. Each component that uses an 
+external tool should include a `ComponentExternalTool` object for that tool in its
```
whyitfor, line 13:
```suggestion
external tool should include a [ComponentExternalTool][ofrak.model.component_model.ComponentExternalTool] object for that tool in its 
```

docs/contributor-guide/component/external_tools.md
```diff
@@ -0,0 +1,142 @@
...
+simply installed with `pip`, OFRAK has another mechanism for the making the dependency clear to users
+and easy to install.
+
+### ComponentExternalTool
+
+`ofrak.model.component_model.ComponentExternalTool` is a class that encapsulates the information 
+that OFRAK tracks about an external tool which a component depends on. Each component that uses an 
+external tool should include a `ComponentExternalTool` object for that tool in its 
+`external_dependencies` field (empty by default, this field does not need to be provided for 
+components which do not use an external tool). The `ofrak_components.zip.ZipUnpacker` is an example:
```
whyitfor, line 15:
```suggestion
components which do not use an external tool). The [ZipUnpacker][ofrak_components.zip.ZipUnpacker] is an example:
```

docs/contributor-guide/component/external_tools.md
```diff
@@ -0,0 +1,142 @@
...
+	brew installation: brew install unzip
+	See https://linux.die.net/man/1/unzip for more info and installation help.
+```
+
+See the [ComponentExternalTool][ofrak.model.component_model.ComponentExternalTool] docs for a 
+breakdown of the fields of that class.
+
+### Edge Cases
+
+One of the functions of `ofrak.model.component_model.ComponentExternalTool` is to provide a way for
```
whyitfor, line 75:
```suggestion
One of the functions of [ComponentExternalTool][ofrak.model.component_model.ComponentExternalTool] is to provide a way for 
```

docs/contributor-guide/component/external_tools.md
```diff
@@ -0,0 +1,142 @@
...
+One of the functions of `ofrak.model.component_model.ComponentExternalTool` is to provide a way for 
+OFRAK to check if each dependency is installed. By default, this is done by running a command 
+formed from fields of the `ComponentExternalTool`:
+
+```python
+retcode = subprocess.call([self.tool, self.install_check_arg], ...)
+```
+
+This works for most cases (the `install_check_arg` provides a lot of flexibility), but does not 
+cover certain edge cases. For example, `ofrak_components.squashfs.SquashfsUnpacker` requires
```
whyitfor, line 84:
```suggestion
cover certain edge cases. For example, [SquashfsUnpacker][ofrak_components.squashfs.SquashfsUnpacker] requires 
```

docs/contributor-guide/component/external_tools.md
```diff
@@ -0,0 +1,142 @@
...
+retcode = subprocess.call([self.tool, self.install_check_arg], ...)
+```
+
+This works for most cases (the `install_check_arg` provides a lot of flexibility), but does not 
+cover certain edge cases. For example, `ofrak_components.squashfs.SquashfsUnpacker` requires 
+specifically versions of  `unsquashfs` with the `-no-exit-code` flag. A user may already have 
+`unsquashfs` installed, but an unsuitable version, so simply checking for `unsquashfs` can give a 
+false negative result when a user is checking for missing dependencies.
+
+In such cases, `ofrak.model.component_model.ComponentExternalTool` should be subclassed for that
```
whyitfor, line 89:
```suggestion
In such cases, [ComponentExternalTool][ofrak.model.component_model.ComponentExternalTool] should be subclassed for that 
```

ofrak_components/ofrak_components_test/test_gzip_component.py
```diff
@@ -86,6 +82,6 @@ async def verify(self, root_resource: Resource) -> None:
                 if e.returncode == 2 or e.returncode == -2:
                     data = e.stdout
                 else:
-                    raise RuntimeError(format_called_process_error(e))
+                    raise
```
whyitfor, line 85: What error does this raise?

ofrak_core/pytest_ofrak/patterns/unpack_modify_pack.py
```diff
@@ -20,22 +22,27 @@ class UnpackModifyPackPattern(ABC):
...
+                        modified_tree.splitlines(keepends=True),
+                    )
+                ),
+                end="",
+            )
+            await self.repack(root_resource)
+            await self.verify(root_resource)
+        except CalledProcessError as e:
+            # Better printing of errors if something goes wrong in test setup/execution
+            ComponentSubprocessError(e)
```
whyitfor, line 45: Is this needed? I thought the component itself already raises a `ComponentSubprocessError`.

whyitfor (approved):
Can't wait to use this!

I left a few comments for your consideration.

ofrak_components/ofrak_components_test/test_gzip_component.py
EdwardLarson: A `raise` in an `except` block will re-raise the same error that was caught. It is equivalent to `raise e`

ofrak_core/pytest_ofrak/patterns/unpack_modify_pack.py
EdwardLarson: A lot of the tests ran their own subprocess calls to setup the test files, with the common pattern that they catch and pretty-print the error. I debated just removing that catch entirely - those should never fail, these are the commands to set up the ground truth test assets - but it would help with debugging, so I added the catch here that takes care of the pretty-printing.

Issue #87, opened 2022-10-28
ElfPointerArraySectionAddModifier modifications do not propogate to children view (ElfVirtualAddress)


whyitfor:
**What is the problem? (Here is where you provide a complete Traceback.)**
The following test, when added to `TestElfPointerArraySectionModifier` (from #71 ), fails.
```
    async def test_elf_pointer_array_section_modifier_virtual_address(self, elf_resource: Resource):
        """
        Test that `ElfPointerArraySectionModifier` results in updates to the children
        `ElfVirtualAddress`.
        """
        pointer_array_section = await self._unpack_and_get_first_pointer_array_section(elf_resource)
        original_values = list(await pointer_array_section.get_entries())

        await pointer_array_section.resource.run(
            ElfPointerArraySectionAddModifier,
            ElfPointerArraySectionAddModifierConfig(skip_list=(), add_value=self.add_value),
        )
        updated_pointer_array_section = await pointer_array_section.resource.view_as(
            ElfPointerArraySection
        )

        for i, entry in enumerate(await updated_pointer_array_section.get_entries()):
            assert entry.value - self.add_value == original_values[i].value
```

This test appears to fail because the `ElfVirtualAddress.value` is not updated when the section's entries are retrieved after `ElfPointerArraySectionAddModifier` is run.

**Please provide some information about your environment.**
At minimum we would like the following information on your platform and Python environment:
This was first observed when running the `make image` container on #71.

**If you've discovered it, what is the root cause of the problem?**
This could possibly be related to the way in which `ElfPointerArraySectionAddModifier` is implemented: it does not iterate over its children, but rather modifies its own data directly.

**How often does the issue happen?**
Every time the above-referenced test is run

PR #114, opened 2022-11-18
Update views of a resource when .save() is called on it

EdwardLarson:
**Link to Related Issue(s)**

A subtle bug that could cause some confusion. Usually, when a resource is modified by a component, when those modifications are fetched to the local context the views of that resource are updated (i.e. their fields may change if those values were changed in the resource). However, if one makes a modification locally to a resource then calls .save(), those changes are not applied to the local views.

When `save()` is called on a resource, update the views. A small change is made to the internal method `Resource._update_views` to make this cleaner.

@whyitfor

### DIFF
--- a/ofrak_core/test_ofrak/unit/resource_view/test_view.py
+++ b/ofrak_core/test_ofrak/unit/resource_view/test_view.py
@@ -257,13 +257,16 @@ async def test_resource_view_updates(ofrak_context: OFRAKContext):
     await instr_view.resource.run(
         InstructionModifier, InstructionModifierConfig("add", "r4, r5", InstructionSetMode.NONE)
     )
-    instr_r.add_view(
-        Instruction(0x100, 0x4, "add r4, r5", "add", "r4, r5", InstructionSetMode.NONE)
-    )
-    await instr_r.save()
 
     assert instr_view.mnemonic == "add"
     assert instr_view.operands == "r4, r5"
+
+    instr_r.add_view(
+        Instruction(0x100, 0x4, "sub r4, r5", "sub", "r4, r5", InstructionSetMode.NONE)
+    )
+    await instr_r.save()
+
+    assert instr_view.mnemonic == "sub"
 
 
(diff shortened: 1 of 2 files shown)


whyitfor:
@EdwardLarson will this PR fix #87?

ofrak_core/test_ofrak/unit/resource_view/test_view.py
```diff
@@ -257,13 +257,16 @@ async def test_resource_view_updates(ofrak_context: OFRAKContext):
...
+
     instr_r.add_view(
-        Instruction(0x100, 0x4, "add r4, r5", "add", "r4, r5", InstructionSetMode.NONE)
+        Instruction(0x100, 0x4, "sub r4, r5", "sub", "r4, r5", InstructionSetMode.NONE)
     )
     await instr_r.save()
 
-    assert instr_view.mnemonic == "add"
-    assert instr_view.operands == "r4, r5"
+    assert instr_view.mnemonic == "sub"
```
whyitfor, line 269: This test confused me. Part of the reason is that it is testing two different behaviors (and the original test didn't make it clear what it was testing).

I'd appreciate it if it was split into tow different tests:
```
# test that InstructionModifier changes propogate to resource view.
await instr_view.resource.run(
        InstructionModifier, InstructionModifierConfig("add", "r4, r5", InstructionSetMode.NONE)
    )

    assert instr_view.mnemonic == "add"
    assert instr_view.operands == "r4, r5"
```
and
```
    instr_r.add_view(
        Instruction(0x100, 0x4, "add r4, r5", "add", "r4, r5", InstructionSetMode.NONE)
    )
    await instr_r.save()

assert instr_view.mnemonic == "add"
    assert instr_view.operands == "r4, r5"
```

whyitfor (commented):
Great change!

See my one comment w.r.t. the test case.

whyitfor (approved):
LGTM

PR #120, opened 2022-11-23
More efficient view_as with less async overhead

EdwardLarson:
**Link to Related Issue(s)**
For some cases of large resource trees, the overhead from managing async becomes a large portion of the runtime. One of the "hot" areas of code is the view creation method, which involves lots of async calls and gathers which are not always necessary.

* non-async check for up-to-date attrs before calling analyze in create_view
* remove `Resource.analyze_attributes` method (redundant with `Resource.analyze`)
* Create a non-async private method that returns a view if possible, and an awaitable to get the view if creating the view synchronously is not possible.

@whyitfor

### DIFF
--- a/disassemblers/ofrak_capstone/ofrak_capstone/components.py
+++ b/disassemblers/ofrak_capstone/ofrak_capstone/components.py
@@ -89,7 +89,7 @@ class CapstoneInstructionAnalyzer(InstructionAnalyzer):
         parent_block = await resource.get_parent_as_view(Addressable)
         mode: InstructionSetMode
         if parent_block.resource.has_tag(BasicBlock):
-            bb_attrs = await parent_block.resource.analyze_attributes(BasicBlock.attributes_type)
+            bb_attrs = await parent_block.resource.analyze(BasicBlock.attributes_type)
             mode = bb_attrs.mode
         else:
             mode = InstructionSetMode.NONE
--- a/disassemblers/ofrak_capstone/ofrak_capstone/components.py
+++ b/disassemblers/ofrak_capstone/ofrak_capstone/components.py
@@ -90,7 +90,7 @@ class CapstoneInstructionAnalyzer(InstructionAnalyzer):
         mode: InstructionSetMode
         if parent_block.resource.has_tag(BasicBlock):
             bb_attrs = await parent_block.resource.analyze(BasicBlock.attributes_type)
-            mode = bb_attrs.mode
+            mode = bb_attrs.mode  # type: ignore
         else:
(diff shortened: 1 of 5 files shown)


ofrak_core/ofrak/resource.py
```diff
@@ -783,21 +832,25 @@ def get_most_specific_tags(self) -> Iterable[ResourceTag]:
...
-        #  version of the component in a client-server environment. We could do that efficiently by
-        #  adding a service method that list all available components (and their version)
-        # Check that the attributes are there to begin with
+    def _check_attributes(self, attributes_type: Type[RA]) -> Optional[RA]:
+        """
+        Try to get the current attributes.
+
+        TODO: Should we be using the version as well? The client wouldn't know the
+        version of the component in a client-server environment. We could do that efficiently by
+        adding a service method that list all available components (and their version)
```
rbs-jacob, line 841: I don't understand this comment... What does it mean by "version"?

disassemblers/ofrak_capstone/ofrak_capstone/components.py
```diff
@@ -89,8 +89,8 @@ async def analyze(self, resource: Resource, config=None) -> Instruction:
         parent_block = await resource.get_parent_as_view(Addressable)
         mode: InstructionSetMode
         if parent_block.resource.has_tag(BasicBlock):
-            bb_attrs = await parent_block.resource.analyze_attributes(BasicBlock.attributes_type)
-            mode = bb_attrs.mode
+            bb_attrs = await parent_block.resource.analyze(BasicBlock.attributes_type)
+            mode = bb_attrs.mode  # type: ignore
```
rbs-jacob, line 93: A `typing.cast` on line 92 would probably be clearer than a type ignore below.

rbs-jacob (changes requested):
Pending two minor changes, it looks good to me!

disassemblers/ofrak_capstone/ofrak_capstone/components.py
EdwardLarson: The type it would cast to is `BasicBlock.attributes_type` which MyPy doesn't understand (its created dynamically). All it can know is that it is `ResourceAttributes` which doesn't have a `mode` attribute.

ofrak_core/ofrak/resource.py
EdwardLarson: It's a pretty old comment. IIRC it means that if the attributes were created by a component with a version less than the latest version, maybe we should also invalidate the attributes in that case.

rbs-jacob (approved):

PR #121, opened 2022-11-29
Adding tests for Ghidra disassembler module

sjossi:
Unfortunately had to add a new package to test for the Ghidra server process. I was able to write the tests without major refactoring or even building a separate test image.

Currently marking as draft while I'm making sure all commits are included as I was out for a week.


sjossi:
Wrote tests for the start/stop server and other required functions.

Excluded a few very short functions that immediately call other tested functions because refactoring would have added a lot of overhead without any benefit.

whyitfor:
Refactored config and server tests for clarity.

whyitfor (approved):

PR #122, opened 2022-11-30
Optimizations when auto-running components

EdwardLarson:
**Link to Related Issue(s)**

The `JobService` coordinating component auto-runs introduces significant overhead when lots of components are being auto-run on resources.

There are many small changes in the MR which each contribute some incremental improvement to performance. The main categories of changes are:

* Aggressive caching of things like filters and filter results, which run often and for each resource
* Changes to enable this caching (e.g. frozen dataclasses)
* Micro-optimizations in very hot code (e.g. use pre-populated 2D array rather than defaultdict of lists)
* Reducing asyncio overhead with fewer tasks created, smarter/less usage of `gather` and `wait`
* Batching the code to save resources after a component runs

@whyitfor or @rbs-jacob


ofrak_core/ofrak/model/component_filters.py
```diff
@@ -19,64 +20,66 @@ def _isinstance(*args, **kwargs):
...
     Only allow components which belong to a specific set to be run.
     """
 
-    whitelisted_component_ids: Set[bytes]
+    whitelisted_component_ids: FrozenSet[bytes]
 
-    def __init__(self, *whitelisted_component_ids: bytes):
-        self.whitelisted_component_ids = set(whitelisted_component_ids)
+    def __init__(self, *whitelisted_component_ids):
+        object.__setattr__(self, "whitelisted_component_ids", frozenset(whitelisted_component_ids))
```
whyitfor, line 32: Now that thesse filters are dataclasses, are these `__init__`s still needed?

EdwardLarson: To avoid changing all the usages of the constructors, yes. Otherwise they all needed to be updated to call `frozenset` or other changes... creates a much bigger and muddier changeset, as well as less readable usage code. I prefer to keep the init details contained here, and also leaves it more open to future implementation changes.

ofrak_core/ofrak/model/tag_model.py
```diff
@@ -1,21 +1,25 @@
 import functools
-from collections import defaultdict
-from typing import Set, Iterable, Tuple
+from typing import Iterable, Set, Tuple, Optional, cast, List
 
 
 class ResourceTag(type):
-    @functools.lru_cache(None)
```
whyitfor, line 7: Did you intentionally delete `@functools.lru_cache` from `tag_specificity`?

ofrak_core/ofrak/model/tag_model.py
```diff
@@ -1,21 +1,25 @@
...
-        specificity = 0
-        for base in cls.base_tags():
-            specificity = max(specificity, base.tag_specificity())
+        if cls._specificity is None:
+            specificity = 0
+            for base in cls.base_tags():
+                specificity = max(specificity, base.tag_specificity())
 
-        return specificity + 1
+            cls._specificity = specificity + 1
```
whyitfor, line 21: Could this be a helper function that runs in __init__?

whyitfor (approved):
Appreciate the speedup.

See minor comments.

ofrak_core/ofrak/model/component_filters.py
whyitfor: Downside of this approach is that the constructor doesn't match typehints in dataclass definition. This might be something worth revisiting at some point.

ofrak_core/ofrak/model/tag_model.py
EdwardLarson: Yes, because the caching is within that function now (`cls._specificity` check)

ofrak_core/ofrak/model/tag_model.py
EdwardLarson: Yep, I should have just done that. It's an easy fix.

PR #123, opened 2022-12-02
Use Resource.identify interface in OFRAK server

whyitfor:
Use the `Resource.identify` interface in the OFRAK server.

### DIFF
--- a/frontend/backend/ofrak_server.py
+++ b/frontend/backend/ofrak_server.py
@@ -242,7 +242,7 @@ class AiohttpOFRAKServer:
     @exceptions_to_http(SerializedError)
     async def identify(self, request: Request) -> Response:
         resource = await self._get_resource_for_request(request)
-        result = await resource.auto_run(all_identifiers=True)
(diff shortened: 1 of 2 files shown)


rbs-jacob (changes requested):
This will not work without an additional change. The `resource.identify` function does not return a component run result, which is required for this server method to work. In order to make this change successfully, you must also return the value of `auto_run` in the `Resource.identify` method. That is the reason `auto_run` was used here instead of `identify` in the first place.

rbs-jacob:
If you're changing the return value of `Resource.identify`, it is probably also a good idea to make sure the return values are consistent across the `Resource` component run methods, since I believe `auto_run` is also required for the OFRAK server `analyze` method.

whyitfor:
Updated `Resource.identify` signature to return `ComponentRunResult`.

I'm leaving the `Resource.analyze` as future work -- it is more involved than this change.

rbs-jacob (approved):

PR #126, opened 2022-12-02
Maintenance/patch maker test

whyitfor:
Split ofrak_patch_maker tests out into separate files by architecture.

This is useful as a user can only install a specific architecture and then run tests to validate that the architecture is properly installed.

@andresito00


andresito00:
Looks good. Should be a good precursor to separation of toolchains.

PR #131, opened 2022-12-08
Bump fun-coverage to 0.1.4

whyitfor:
**Link to Related Issue(s)**
fun-coverage 0.1.3 was incorrectly excluding all functions with decorators from coverage results. 

Bump fun-coverage to 0.1.4.

### DIFF
--- a/ofrak_components/ofrak_components/uf2.py
+++ b/ofrak_components/ofrak_components/uf2.py
@@ -12,7 +12,6 @@ from ofrak.component.packer import Packer
 from ofrak.component.identifier import Identifier
 from ofrak.core.binary import GenericBinary
 from ofrak_type.range import Range
-from ofrak.model.resource_model import index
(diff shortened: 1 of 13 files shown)


ofrak_components/ofrak_components/uf2.py
```diff
@@ -66,10 +65,6 @@ class Uf2BlockHeader(ResourceAttributes):
     num_blocks: int
     filesize_family_id: int
 
-    @index
-    def BlockNo(self) -> int:
-        return self.block_no
```
rbs-jacob, line 71: Why remove this instead of adding a test for it? Are you confident people won't want to sort by block number in their results?

whyitfor: This index isn't needed for current functionality of Uf2 unpacker and packer.

If a future use case required it, it can always be added back.

rbs-jacob (approved):

Issue #134, opened 2022-12-10
ofrak CLI tool does not work when binwalk is not installed


whyitfor:
**What is the problem? (Here is where you provide a complete Traceback.)**
Running `python3 -m ofrak deps` or `python3 -m ofrak list` fails when the binwalk package is not installed:

```
% python3 -m ofrak list
Traceback (most recent call last):
 <REDACTED>/ofrak/ofrak_components/ofrak_components/binwalk.py", line 7, in <module>
    import binwalk
ModuleNotFoundError: No module named 'binwalk'
```

**Please provide some information about your environment.**
Running ofrak on 385ad3eff4d24a45747a752b1c8a77bec54b8da7.

**If you've discovered it, what is the root cause of the problem?**
There are two related problems:
1. Binwalk is imported at the top of the file. If the package is not installed, this results in a `ModuleNotFoundError`.
2. Binwalk requires a custom `ComponentExternalTool.is_tool_installed` implementation, since ofrak relies on the Python package being installed and not the CLI tool (which may be installed on a system but using a different version of python, different python environment).

**How often does the issue happen?**
Every time the ofrak cli tool is run.

**What are the steps to reproduce the issue?**
Ideally, give us a short script that reproduces the issue.
`python3 -m ofrak deps`.

**How would you implement this fix?**
1. Add a custom `ComponentExternalTool.is_tool_installed` implementation for binwalk to check if the python package is installed.
3. Handle the `ModuleNotFoundError` so that OFRAK can run even when this tool is not installed. 

**Are there any (reasonable) alternative approaches?**
Additionally, I would like to add a check that excludes `BinwalkAnalyzer` from discovery at runtime if binwalk is not installed. This is probably beyond scope of this issue.

**Are you interested in implementing it yourself?**
Yes.

whyitfor:
Fixed in #143.

PR #135, opened 2022-12-10
OFRAKContext doesn't register components with missing dependencies

whyitfor:
This PR updates the behavior of `OFRAK.create_ofrak_context` to check if components have their dependencies installed before registering them.

Before registering components with the component locator, each component's dependencies are checked to ensure that they are installed. Only components with installed dependencies are added to the component locator.

Additionally, a log message (WARNING level) reports on dependencies that were not installed. Example output:

```
In [1]: from ofrak import *
   ...: ofrak = OFRAK()
   ...: context = await ofrak.create_ofrak_context()
[ofrak_context.py:  119] Skipped registering the following components due to missing dependencies: ApkPacker, 
ApkUnpacker, BinwalkAnalyzer, LzoPacker, LzoUnpacker, RarUnpacker. Run `python3 -m ofrak deps --missing-only` 
for more details.
```

This functionality tries to strike a balance between alerting users to missing dependencies and allowing users to run OFRAK with a minimal set of installed dependencies.

Additionally, this PR addresses #134 by creating a `_BinwalkExternalTool`.

**Link to Related Issue(s)**
#134.

### DIFF
--- a/ofrak_core/ofrak/ofrak_context.py
+++ b/ofrak_core/ofrak/ofrak_context.py
@@ -110,9 +110,19 @@ class OFRAK:
         component_locator = await self.injector.get_instance(ComponentLocatorInterface)
 
         resource_factory = await self.injector.get_instance(ResourceFactory)
-        component_locator.add_components(
-            await self.injector.get_instance(List[ComponentInterface]), self._discovered_modules
+        component_auditor = _ComponentDepAuditor()
+        components = [
+            component
+            for component in await self.injector.get_instance(List[ComponentInterface])
+            if component_auditor.are_dependencies_installed(component)
+        ]
+        LOGGER.warning(
+            f"Skipped registering the following components due to missing dependencies: "
+            f"{', '.join(component_auditor.components_missing_deps)}. Run `python3 -m ofrak deps "
+            f"--missing-only` for "
+            f"more details."
         )
+        component_locator.add_components(components, self._discovered_modules)
 
         id_service = await self.injector.get_instance(IDServiceInterface)
         data_service = await self.injector.get_instance(DataServiceInterface)
@@ -158,3 +168,22 @@ class OFRAK:
             self.discover(ofrak_components)
         except ModuleNotFoundError:
             pass
+
+
+class _ComponentDepAuditor:
+    """
+    Helper class to track components that have missing dependencies.
+    """
+
+    def __init__(self):
+        self.components_missing_deps: List[str] = list()
+
+    def are_dependencies_installed(
+        self,
+        component: ComponentInterface,
+    ) -> bool:
+        for dependency in component.external_dependencies:
+            if not dependency.is_tool_installed():
+                self.components_missing_deps.append(component.get_id().decode())
+                return False
+        return True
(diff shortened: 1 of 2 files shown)


ofrak_core/ofrak/ofrak_context.py
```diff
@@ -158,3 +168,22 @@ def _setup(self):
             self.discover(ofrak_components)
         except ModuleNotFoundError:
             pass
+
+
+class _ComponentDepAuditor:
```
EdwardLarson, line 173: Adding a class for this seems unnecessary, I found it slower to see what was going on than a simple `if not all(dependency.is_tool_installed() for dependency in component.external_dependencies)` check in the loop over all components would be. If we really want to keep such mechanics out of the `create_ofrak_context` method, then I think adding an `dependencies_are_installed` method to `ComponentInterface`/`AbstractComponent` makes more sense.

ofrak_core/ofrak/ofrak_context.py
```diff
@@ -110,9 +110,19 @@ async def create_ofrak_context(self) -> OFRAKContext:
         component_locator = await self.injector.get_instance(ComponentLocatorInterface)
 
         resource_factory = await self.injector.get_instance(ResourceFactory)
-        component_locator.add_components(
-            await self.injector.get_instance(List[ComponentInterface]), self._discovered_modules
+        component_auditor = _ComponentDepAuditor()
+        components = [
```
EdwardLarson, line 114: I think components should only be audited when a specific flag is passed. Perhaps add a note in the missing component error message, suggesting this option. By default I believe all components should be imported and used.

EdwardLarson (changes requested):

ofrak_core/ofrak/ofrak_context.py
rbs-jacob: To clarify: if all components are imported and used as you suggest, then there will be a dependency error when the user tries to actually run the component, right? Whereas if they're audited at the beginning, some dependency injector stuff will fail earlier in the script.

As an OFRAK user, I'd probably rather know about any errors in my script as soon as possible since complicated scripts can take a while. If I'm understanding correctly, that could be a case for auditing as early as possible.

dannyp303: I agree with jacob here, especially during development, on occasion waiting 5+ minutes to see that a typo made it into my docker and failed the import stage, not hitting the error until component lookup.

ofrak_core/ofrak/ofrak_context.py
dannyp303: I'm not sure about this, new class does seem unnecessary, but from a user perspective, I could see something similar being used to clean up some DI stuff that is confusing to users, I.E conflicting/duplicated/stale components.

ofrak_core/ofrak/ofrak_context.py
EdwardLarson: So during development, use that flag to audit components. I'm arguing that the default behavior should simply import and use the components as the user directed it to (with `.discover`), and auditing should be an option that needs to be toggled on. I think our default assumption should be that OFRAK is installed correctly, and code to deal with cases where it is not (especially code that increases runtime) should not be run by default.

I'm not sure how dependency auditing will help with typos in components though, so maybe I'm misunderstanding the conversation here.

EdwardLarson:
Closing this is duplicated by https://github.com/redballoonsecurity/ofrak/pull/143 (remade the PR with an accurate branch name)

PR #139, opened 2022-12-13
Ghidra PIE Support

SamL98:
Works by shifting addresses from ELF unpacker in and out of the backend. Would like to just set the CodeRegion's virtual_address from the backend in the future.

### DIFF
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
@@ -35,31 +35,40 @@ class GhidraCodeRegionUnpacker(CodeRegionUnpacker, OfrakGhidraMixin):
         os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetComplexBlocks.java"),
     )
 
+    get_code_regions_script = OfrakGhidraScript(
+        os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetCodeRegions.java"),
+    )
+
     async def unpack(self, resource: Resource, config=None) -> None:
         program = await resource.get_only_ancestor_as_view(
             GhidraProject, ResourceFilter(tags=[GhidraProject], include_self=True)
         )
         program_attributes = await program.resource.analyze(ProgramAttributes)
         code_region = await resource.view_as(CodeRegion)
-        code_region_start = code_region.virtual_address
+        code_region_start = await self.fixup_address(code_region.virtual_address, program.resource)
         code_region_end = code_region_start + code_region.size
+        va_addend = code_region.virtual_address - code_region_start
 
         complex_blocks = await self.get_complex_blocks_script.call_script(
             resource,
             hex(code_region_start),
             hex(code_region_end),
         )
-        complex_blocks_created = list()
+
+        complex_blocks_created = []
+
         for complex_block in complex_blocks:
             complex_block = ComplexBlock(
-                complex_block["loadAddress"], complex_block["size"], complex_block["name"]
+                complex_block["loadAddress"] + va_addend, complex_block["size"], complex_block["name"]
             )
+
             complex_blocks_created.append(
                 code_region.create_child_region(
                     complex_block,
                     additional_attributes=(program_attributes,),
                 )
             )
+
         await asyncio.gather(*complex_blocks_created)
 
 
@@ -70,8 +79,13 @@ class GhidraComplexBlockUnpacker(
     get_basic_blocks_script = OfrakGhidraScript(
         os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetBasicBlocks.java")
     )
+
     get_data_words_script = OfrakGhidraScript(
         os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetDataWords.java")
+    )
+
+    get_code_regions_script = OfrakGhidraScript(
+        os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetCodeRegions.java"),
     )
 
     def __init__(
@@ -86,12 +100,17 @@ class GhidraComplexBlockUnpacker(
         super().__init__(resource_factory, data_service, resource_service, component_locator)
 
     async def unpack(self, resource: Resource, config=None):
-        cb_view: ComplexBlock = await resource.view_as(ComplexBlock)
+        program = await resource.get_only_ancestor_as_view(
+            GhidraProject, ResourceFilter(tags=[GhidraProject], include_self=True)
+        )
+
+        cb_view = await resource.view_as(ComplexBlock)
 
         program_attrs = await resource.analyze(ProgramAttributes)
 
         cb_data_range = await resource.get_data_range_within_root()
-        cb_start_vaddr = cb_view.virtual_address
+        cb_start_vaddr = await self.fixup_address(cb_view.virtual_address, program.resource)
+        va_addend = cb_view.virtual_address - cb_start_vaddr
 
         children_created = []
 
@@ -100,12 +119,12 @@ class GhidraComplexBlockUnpacker(
         )
 
         for bb_info in basic_blocks:
-            bb_start_vaddr = bb_info["bb_start_vaddr"]
+            bb_start_vaddr = bb_info["bb_start_vaddr"] + va_addend
             bb_size = bb_info["bb_size"]
             is_exit_point = bb_info["is_exit_point"]
             mode_string = bb_info["instr_mode"]
-            exit_vaddr = bb_info["exit_vaddr"]
-            if exit_vaddr == -1:
+            exit_vaddr = bb_info["exit_vaddr"] + va_addend
+            if (exit_vaddr - va_addend) == -1:
                 exit_vaddr = None
 
             if bb_size == 0:
@@ -140,9 +159,9 @@ class GhidraComplexBlockUnpacker(
         )
 
         for data_word_info in data_words:
-            word_vaddr = data_word_info["word_vaddr"]
+            word_vaddr = data_word_info["word_vaddr"] + va_addend
             word_size = data_word_info["word_size"]
-            xrefs = data_word_info["xrefs"]
+            xrefs = [xref + va_addend for xref in data_word_info["xrefs"]]
 
             if (
                 word_vaddr < cb_view.virtual_address
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
@@ -17,6 +17,7 @@ from ofrak.service.data_service_i import DataServiceInterface
 from ofrak.service.resource_service_i import ResourceFilter, ResourceServiceInterface
 from ofrak_ghidra.constants import CORE_OFRAK_GHIDRA_SCRIPTS
 from ofrak_ghidra.ghidra_model import GhidraProject, OfrakGhidraMixin, OfrakGhidraScript
+from ofrak_ghidra.components.ghidra_analyzer import GhidraCodeRegionAnalyzer
 from ofrak_io.batch_manager import make_batch_manager
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
```diff
@@ -35,31 +35,40 @@ class GhidraCodeRegionUnpacker(CodeRegionUnpacker, OfrakGhidraMixin):
...
             hex(code_region_end),
         )
-        complex_blocks_created = list()
+
+        complex_blocks_created = []
+
         for complex_block in complex_blocks:
             complex_block = ComplexBlock(
-                complex_block["loadAddress"], complex_block["size"], complex_block["name"]
+                complex_block["loadAddress"] + va_addend, complex_block["size"], complex_block["name"]
```
EdwardLarson, line 62: This PR gets a lot simpler if you don't add the `va_addend` here and just let it use the vaddr the backend provides. Then the created ComplexBlocks will all have the "correct" vaddr and then we don't need to do any later checks when unpacking CBs or BBs.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
```diff
@@ -82,6 +87,38 @@ class OfrakGhidraMixin:
     `OfrakGhidraScript` member of the class.
     """
 
+    async def fixup_address(self, va: int, resource: Resource) -> int:
```
EdwardLarson, line 90: I think calling this on every ComplexBlock and BasicBlock will have a very heavy performance penalty. See my other comment on not using the `va_addend`, we should only need to do this fixup up in the `CodeRegionUnpacker`, nowhere else.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
```diff
@@ -56,8 +59,10 @@ async def call_script(self, resource: Resource, *script_args: str) -> Any:
         params = {f"__arg_{i}": arg for i, arg in enumerate(script_args)}
 
         async with aiohttp.ClientSession() as requests:
+            endpoint = self.script_name.split('.')[0].lower()
+
             response = await requests.get(
-                f"{root_ghidra_project.ghidra_url}/{self.script_name}", params=params
+                f"{root_ghidra_project.ghidra_url}/{endpoint}", params=params
```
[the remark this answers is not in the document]
SamL98: When spawning the analysis server, the arguments conclude with <(script name, endpoint), (script name, endpoint), ...>. But the analysis server was just iterating over all the remaining arguments and interpreting each single argument as both an endpoint and a script name. I didn't think this made much sense so I changed it to likely what was intended with the analysis server arguments.

disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
SamL98: Wouldn't then ComplexBlocks (and then consequently BasicBlocks and Instructions) have virtual addresses that are either not contained or misaligned with the CodeRegion's virtual address? This seems not good to me. That's why I think we should switch to setting the CodeRegion's virtual address using the one from the backend.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
SamL98: I agree it's bad for performance. That's why we fix OFRAK's CodeRegion's in the CodeRegionUnpacker or earlier

disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
EdwardLarson: Yes it will be mismatched, that's why correcting for the misaligned code regions so we can get complex blocks was discussed as the first step towards feature completion. So now we can sync up on what is the best path forward.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
```diff
@@ -82,6 +88,37 @@ class OfrakGhidraMixin:
     `OfrakGhidraScript` member of the class.
     """
 
+    # Hi Ed, I know this isn't the right way to do this but I don't know how to add an attribute to a Resource.
+    ofrak_code_regions = {}
```
EdwardLarson, line 92: I was thinking that this caching would be good to put in the `GhidraProject` view - it's the bridge between the OFRAK project and the corresponding Ghidra project, so it makes sense to expand its metadata to include the vaddr offsets.

This would mean calling the script to get Ghidra's regions in `GhidraProjectAnalyzer` where the `GhidraProject` view is populated.

Also, this cache can definitely be simplified. The current data stored and the algo run on it each time represent something like a dictionary of ranges+offset pairs, where if an address is between one of those ranges, we apply to it the offset that goes with that range. Really we just need those 3 values for each code region (start, end, vaddr addend), no views or resources necessary.

SamL98: I think that it's necessary to key on the resource because if we are trying to translate addresses in two different resources, the memory mappings will be different but the address might fall in to a segment of the previously translated resource.

This would be solved by adding the memory map info to the `GhidraProject` (assuming that only one file is analyzed in on project) but I don't know how to do that.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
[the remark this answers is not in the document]
EdwardLarson: Good catch! It's a holdover from when the script name and endpoint were supposed to be different, in the `OfrakGhidraScript` you can see the script name is always stripped of the `.java` extension anyway:

```
self.script_name = os.path.basename(script_path).split(".")[0]
```

So script names and the "endpoints" calculated from them are always equal. I think the right place to fix the inconsistency is in `ghidra_analyzer.py`, removing line 223 so the redundant handler for each script is eliminated.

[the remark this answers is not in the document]
SamL98: Ok I fixed it in `AnalysisServer.java` by creating each endpoint with the endpoint name pointing to the script name. Should we fix it here or can we try to merge this PR in?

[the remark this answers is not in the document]
EdwardLarson: If you want to address this behavior in this PR, the correct code change to align the code with the intended behavior is to delete `ghidra_analyzer.py` line 223, and no changes are needed in `AnalysisServer.java` or anywhere else. It doesn't necessarily need to be addressed in this PR, it's easy enough to make a quick bugfix PR to address it. But this PR needs to either address it by removing `ghidra_analyzer.py:223` with none of the other current changes related to this (e.g. in `ghidra_model.py`), or not address it and revert those changes related to it.

disassemblers/ofrak_ghidra/ofrak_ghidra/components/ghidra_analyzer.py
```diff
@@ -223,3 +227,46 @@ def _build_ghidra_server_args(self) -> List[str]:
...
 
         return args
+
+
+@dataclass
+class GhidraCodeRegionConfig(ComponentConfig):
+    pass
+
+
+class GhidraCodeRegionAnalyzer(Analyzer[Optional[GhidraCodeRegionConfig], CodeRegion], OfrakGhidraMixin):
```
EdwardLarson, line 237: This actually acts as a `Modifier`, changing the definition to be that of a `Modifier` will I think make this ready for merge

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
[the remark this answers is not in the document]
SamL98: What do you mean revert the changes? The arguments passed to `AnalysisServer.java` currently are <(script name, endpoint), (script name, endpoint), ...> and the endpoints are set up as one would expect as <endpoint -> script name, endpoint -> script name, ...>

disassemblers/ofrak_ghidra/ofrak_ghidra/components/blocks/unpackers.py
```diff
@@ -86,7 +97,11 @@ def __init__(
         super().__init__(resource_factory, data_service, resource_service, component_locator)
 
     async def unpack(self, resource: Resource, config=None):
-        cb_view: ComplexBlock = await resource.view_as(ComplexBlock)
+        program = await resource.get_only_ancestor_as_view(
```
EdwardLarson, line 100: `program` unused, and the following changes in the file don't seem necessary

disassemblers/ofrak_ghidra/ofrak_ghidra/components/ghidra_analyzer.py
```diff
@@ -220,6 +225,51 @@ def _build_ghidra_server_args(self) -> List[str]:
...
+
+class GhidraCodeRegionModifier(Modifier[Optional[GhidraCodeRegionConfig]], OfrakGhidraMixin):
+    id = b"GhidraCodeRegionModifier"
+    targets = (CodeRegion,)
+
+    get_code_regions_script = OfrakGhidraScript(
+        os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetCodeRegions.java"),
+    )
+
+    async def modify(self, resource: Resource, config: Optional[GhidraCodeRegionConfig] = None):
```
EdwardLarson, line 245: Components don't need to define a config if they don't take any "parameter"

```suggestion

class GhidraCodeRegionModifier(Modifier[None], OfrakGhidraMixin):
    id = b"GhidraCodeRegionModifier"
    targets = (CodeRegion,)

    get_code_regions_script = OfrakGhidraScript(
        os.path.join(CORE_OFRAK_GHIDRA_SCRIPTS, "GetCodeRegions.java"),
    )

    async def modify(self, resource: Resource, config=None):
```

disassemblers/ofrak_ghidra/ofrak_ghidra/components/ghidra_analyzer.py
```diff
@@ -220,6 +225,51 @@ def _build_ghidra_server_args(self) -> List[str]:
...
+        backend_code_regions = []
+
+        for cr_j in backend_code_regions_json:
+            cr = CodeRegion(cr_j["start"], cr_j["size"])
+            backend_code_regions.append(cr)
+
+        ofrak_code_regions = sorted(ofrak_code_regions, key=lambda cr: cr.virtual_address)
+        backend_code_regions = sorted(backend_code_regions, key=lambda cr: cr.virtual_address)
+
+        for cr in ofrak_code_regions:
```
EdwardLarson, line 263: Why do you need to get and iterate all the `ofrak_code_regions` here? It looks like the only reason is to check `if cr.virtual_address == code_region.virtual_address`, but in that case `cr` is `code_region` so there is no need to check for which `cr` matches the vaddr. We already have that code region, it's the variable `code_region`.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_model.py
```diff
@@ -4,6 +4,7 @@
 
 import aiohttp
 
+from ofrak import ResourceFilter
```
EdwardLarson, line 7: Delete this line or 10, one of these is an unneeded/unused import - did this pass `autoflake`?

EdwardLarson (changes requested):

disassemblers/ofrak_ghidra/ofrak_ghidra/components/ghidra_analyzer.py
```diff
@@ -220,6 +225,51 @@ def _build_ghidra_server_args(self) -> List[str]:
...
+
+                for backend_cr in backend_code_regions:
+                    backend_relative_va = (
+                        backend_cr.virtual_address - backend_code_regions[0].virtual_address
+                    )
+
+                    if backend_relative_va == relative_va and backend_cr.size == cr.size:
+                        cr.resource.add_view(backend_cr)
+
+        LOGGER.debug(f"Could not find CodeRegion for {code_region}")
```
EdwardLarson, line 275: Looks like this will be printed every time the modifier runs, could also be clearer in what it's saying:
```suggestion
                        cr.resource.add_view(backend_cr)
                        return

        LOGGER.debug(f"Could not find matching Ghidra CodeRegion for {code_region}")
```

SamL98: This will only print if there was no code region found in ghidra that matches the relative offset and size of the one in ofrak

disassemblers/ofrak_ghidra/ofrak_ghidra/components/ghidra_analyzer.py
SamL98: Good catch. I was using the index in `ofrak_code_regions` of `code_region` hence iterating over them to find it but I changed the inner loop so no need to anymore

disassemblers/ofrak_ghidra/ofrak_ghidra/components/ghidra_analyzer.py
EdwardLarson: In the current version of this PR, there is no early exit in that case , so the log message is always printed. That's why in my suggestion I've added a `return` statement.

SamL98: oh whoops

PR #143, opened 2022-12-14
OFRAKContext doesn't register components with missing dependencies

EdwardLarson:
**Link to Related Issue(s)**
Replacement for #135 
Addresses #134 

@dannyp303 
@whyitfor 
@rbs-jacob 

As y'all were on the original PR


ofrak_components/ofrak_components/entropy/entropy.py
```diff
@@ -61,6 +82,9 @@ async def analyze(self, resource: Resource, config=None, depth=0) -> DataSummary
                 f"Analysis process killed more than {self.max_analysis_retries} times. Aborting."
             )
 
+        if not _ENTROPY_SO_DEPENDENCY.is_tool_installed():
+            raise ComponentMissingDependencyError(self, _ENTROPY_SO_DEPENDENCY)
+
```
whyitfor, line 87: I don't understand why this is necessary.

Are you proposing that each component checks in init if its dependencies are installed and raises an error if they are not?

whyitfor: If I understand https://github.com/redballoonsecurity/ofrak/pull/135#discussion_r1046228301 and https://github.com/redballoonsecurity/ofrak/pull/135#discussion_r1046291523, the recommendations were as follows:

1. By default, OFRAK should fail fast if any registered component is missing a dependency (better to find this out now as opposed to minutes into a script)
2. It could be nice to have a flag that allows OFRAK to continue running but ignore such incomplete components.

If this is the case, I'd recommend that `AbstractComponent.__init__` includes logic for 1:
```
for dep in self.external_dependencies:
    if not del.is_tool_installed():
        raise ComponentMissingDependencyError(self, dep)
``` 

@rbs-jacob, @dannyp303, would appreciate your input here as well

ofrak_core/ofrak/ofrak_context.py
```diff
@@ -83,12 +83,20 @@ async def shutdown_context(self):
 
 
 class OFRAK:
-    def __init__(self, logging_level: int = logging.WARNING):
+    def __init__(self, logging_level: int = logging.WARNING, audit_dependencies: bool = False):
```
whyitfor, line 86: What about `skip_incomplete_components` instead of `audit_dependencies`? I find it hard to understand what this flag means without reading the docstring.

ofrak_components/ofrak_components/entropy/entropy.py
EdwardLarson: > Are you proposing that each component checks in init if its dependencies are installed and raises an error if they are not?

Most external dependencies are CLI tools called with subprocess or equivalent, and if they are not installed, that subprocess call raises a FileNotFoundError which OFRAK can and does trivially catch and transform into a `ComponentMissingDependencyError` to make the problem very clear to the user. The entropy.so is one of the very few exceptions to this rule (binwalk is another), so is one of the few external dependencies which wouldn't raise `ComponentMissingDependencyError` without entropy.so-specific handling.

> If I understand https://github.com/redballoonsecurity/ofrak/pull/135#discussion_r1046228301 and https://github.com/redballoonsecurity/ofrak/pull/135#discussion_r1046291523, the recommendations were as follows:

No, my recommendation was that by default OFRAK assumes that it is fully installed and it doesn't need to check everything. There has been no further discussion as yet on whether this should be default or not. This implementation (with not checking as default) satisfies all the use cases raised in those comments.

ofrak_core/ofrak/ofrak_context.py
EdwardLarson: It's not obvious was "incomplete" means - the most clear flag I think is `skip_components_missing_dependencies`, but I thought that was too long.

ofrak_components/ofrak_components/entropy/entropy.py
whyitfor: Oh ok, I see that it is in `analyze` and not `__init__`

ofrak_core/ofrak/ofrak_context.py
whyitfor: Gotcha. If we can get the word `skip` into it I think it would make it clearer. Not critical though

whyitfor (approved):
Looks good! I left a few comments, feel free to resolve as you see fit and merge in.

ofrak_core/ofrak/ofrak_context.py
whyitfor: Sure!

PR #149, opened 2022-12-19
Replace .attributes_type with AttributesType

EdwardLarson:
**Link to Related Issue(s)**
Originally for the purpose of supporting #139 , this MR makes it more straightforward to write analyzers outputing specific pieces of a view. Views are composed of one or more auto-generated ResourceAttributes, and while these auto-generated types could be access, previously they could not be used as actual types that MyPy understands.

* In order to facilitate type-checking, a replacement for the `attributes_type` property of `ResourceView`s is implemented, `AttributesType`, which is a generic class: `NamedProgramSection.attributes_type` becomes `AttributesType[NamedProgramSection]`. Essentially this is necessary because mypy is not willing to consider `NamedProgramSection.attributes_type` as a type, but it does understand `AttributesType[NamedProgramSection]` with the proper hints, and the Python interpreter of course is fine with either one.
* All usages of `.attibutes_type` in OFRAK are appropriately replaced
* The names of the auto-generated attributes classes were changed to be `"AttributesType[X]"` rather than `"XAutoAttributes"`, so that when these show up in logs or errors, users are now seeing a name that they could actually import and use. I'm not 100% sure this is a good idea, since the class returned by `AttributesType[X]` is a separate class that isn't actually a subclass of `AttributesType` yet is literally called `"AttributesType[X]"` - which, by the way, would not be legal syntax to try to define statically with the `class` keyword. But, I do think it is better than having error messages with magic classes that don't seem to exist.
* The `ElfSectionNameAnalyzer` is split into the `ElfSectionNameAnalyzer` and `ElfSectionMemoryRegionAnalyzer` which are each more specific than the old `ElfSectionNameAnalyzer`, as a useful example of what this change allows.
* `.attribute_type` is deprecated (but remains usable).

@rbs-jacob


ofrak_core/ofrak/service/job_service.py
```diff
@@ -233,6 +233,7 @@ async def run_analyzer_by_attribute(
         else:
             raise NotFoundError(
                 f"Unable to find any analyzer for attributes {request.attributes.__name__}"
+                f"\nFilter:{component_filter}"
```
rbs-jacob, line 236:
```suggestion
                f"\nFilter: {component_filter}"
```

ofrak_core/ofrak/model/viewable_tag_model.py
```diff
@@ -23,6 +23,14 @@
 RVI = TypeVar("RVI", bound="ResourceViewInterface")
 
 
+class AttributesType(ResourceAttributes, Generic[RVI]):
+    def __init__(self, *args: Any, **kwargs: Any):
+        raise NotImplementedError("AttributesType cannot be instantiated")
+
+    def __class_getitem__(cls, item: Type[RVI]) -> Type[ResourceAttributes]:
+        return item.attributes_type
```
rbs-jacob, line 31: This is slick!

ofrak_core/ofrak/core/elf/analyzer.py
```diff
@@ -333,9 +336,26 @@ async def analyze(self, resource: Resource, config=None) -> ElfSection:
...
+        return AttributesType[NamedProgramSection](
             name=section_name,
+        )
+
+
+class ElfSectionMemoryRegionAnalyzer(Analyzer[None, MemoryRegion]):
+    """
+    Get the name of an ELF section. ELF section names are stored as null-terminated strings in
+    dedicated string section, and each ELF section header's `sh_name` field is an offset in this
+    section.
```
rbs-jacob, line 348: I think this docstring is incorrect – it's supposed to be for the other analyzer, right?

rbs-jacob (changes requested):
A couple minor comments, but otherwise, looks good to me! 

Did you `grep` for `attributes_type` in other places and replace usages with the new version? If not, should you open an issue for that task?

ofrak_core/ofrak/core/elf/unpacker.py
```diff
@@ -115,7 +116,7 @@ async def unpack(self, resource: Resource, config=None):
             e_section_header_r = await resource.create_child(
                 tags=(ElfSectionHeader,),
                 data_range=e_section_header_range,
-                attributes=(ElfSectionStructure.attributes_type(index),),  # type: ignore
+                attributes=(AttributesType[ElfSectionStructure](index),),  # type: ignore
```
rbs-jacob, line 119: Are  the `# type: ignore` comments still necessary?

ofrak_core/ofrak/core/instruction.py
```diff
@@ -176,7 +177,7 @@ async def modify(self, resource: Resource, config: InstructionModifierConfig):
             len(asm) == resource_memory_region.size
         ), "The modified instruction length does not match the original instruction length"
 
-        new_attributes = Instruction.attributes_type(  # type: ignore
+        new_attributes = AttributesType[Instruction](  # type: ignore
```
rbs-jacob, line 180: Same question as above (and throughout): is it still necessary to type ignore here?

ofrak_core/ofrak/model/viewable_tag_model.py
```diff
@@ -16,19 +16,37 @@
...
 RVI = TypeVar("RVI", bound="ResourceViewInterface")
 
 
 class AttributesType(ResourceAttributes, Generic[RVI]):
+    """
+    A Generic type for `ViewableResourceTag` to get the auto-generated `ResourceAttributes`
+    class associated with that view type. The returned class is a `dataclass` which encapsulates
+    the fields defined in one specific `ViewableResourceTag`.
+    For example if `B` inherits from `A` and `A` defines one or more new fields,
+    `AttributesType[B]` has only fields defined in `B`, and none of the fields defined in `A`.
```
rbs-jacob, line 31: If these are separate paragraphs, they need a blank line between to be separated in rendered Markdown. If they're not supposed to be separate paragraphs, the lines can be combined and re-wrapped.

```suggestion
    A Generic type for `ViewableResourceTag` to get the auto-generated `ResourceAttributes`
    class associated with that view type. The returned class is a `dataclass` which encapsulates
    the fields defined in one specific `ViewableResourceTag`.
    
    For example if `B` inherits from `A` and `A` defines one or more new fields,
    `AttributesType[B]` has only fields defined in `B`, and none of the fields defined in `A`.
```

ofrak_core/ofrak/model/viewable_tag_model.py
```diff
@@ -16,19 +16,37 @@
...
     def __class_getitem__(cls, item: Type[RVI]) -> Type[ResourceAttributes]:
-        return item.attributes_type
+        return getattr(item, _VIEW_ATTRIBUTES_TYPE)
+
+    def __getattr__(self, item):  # pragma: no cover
+        """
+        Type stub that 'tricks' MyPy into not trying to typecheck attribute accesses of an
+        `AttributesType` instance. This stub solves the following case:
+        x: AttributesType[X] = func()
+        x.any_field  <--- Always is a type error because MyPy thinks AttributesType has no fields!
```
rbs-jacob, line 45:
````suggestion

        ```
        x: AttributesType[X] = func()
        x.any_field  <--- Always is a type error because MyPy thinks AttributesType has no fields!
        ```
````

rbs-jacob (changes requested):
Some small questions, comments, and suggestions. Other than that, this looks good! Glad some more instances were caught.

ofrak_core/ofrak/core/elf/unpacker.py
EdwardLarson: I grepped for all the "type: ignore" comments in the lines I made this pattern of change to and removed them

rbs-jacob (approved):

PR #151, opened 2022-12-23
Decouple patch maker from ofrak

EdwardLarson:
**Link to Related Issue(s)**
`ofrak_patch_maker`'s dependency on OFRAK created a circular dependency. This wasn't necessarily problematic as pip handled it gracefully, but it did preclude a "lightweight" patch maker install without OFRAK.

Things were moved/refactored so that `ofrak_patch_maker` doesn't import from `ofrak`.

* `ProgramAttributes` gets a superclass `ArchInfo` defined in `ofrak_type` so that we don't rely on a `ResourceAttributes` for just grouping arch information
* The `BOM` and `AssembledObject` get some additional fields to include enough information that we don't need to call out to a `Toolchain` after creation - required BSS size and alignment, specifically
* The `get_required_alignment` method in `Toolchain` is also changed a bit, to a property, to support this
* `allocate_bom` was moved to `Allocatable`

@andresito00

### DIFF
--- a/ofrak_patch_maker/ofrak_patch_maker/patch_maker.py
+++ b/ofrak_patch_maker/ofrak_patch_maker/patch_maker.py
@@ -36,6 +36,7 @@ import os
 import tempfile
 from types import ModuleType
 from typing import Optional, List, Dict, Union, Tuple, Iterable, Mapping
+from warnings import warn
 
 from immutabledict import immutabledict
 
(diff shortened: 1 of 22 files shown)


EdwardLarson:
@andresito00 this is ready for review when you are, I don't think there should be any surprises as compared to what we discussed last week. 

What do you think of leaving the old `allocate_bom` method in for backwards compatibility? It just calls the new implementation, so the only OFRAK "dependency" there is just for the typehint, which I stringified so we don't even need to import from OFRAK. Just lets us delay having to fix up all the scripts that use `allocate_bom` for now so we can progress with OFRAK core, though we'll want to update those eventually.

ofrak_patch_maker/ofrak_patch_maker/patch_maker.py
```diff
@@ -482,74 +479,13 @@ def make_fem(
...
-
-        all_segments: Dict[str, Tuple[Segment, ...]] = {
-            object_path: tuple(segments) for object_path, segments in segments_by_object.items()
-        }
-
-        return PatchRegionConfig(bom.name + "_patch", immutabledict(all_segments))
+        # raise DeprecationWarning(
+        #     "PatchMaker.allocate_bom(allocatable, bom) is deprecated! Use "
+        #     "allocatable.allocate_bom(bom) instead."
+        # )
```
andresito00, line 490: I know we tend to be pretty aggressive about deleting code, but I think this is a situation where we can keep a deprecation warning around for a bit.

PR #158, opened 2023-01-04
Use Ghidra exclusively for examples, test that examples work in CI

whyitfor:
Refactor example scripts to use Ghidra. Test these examples in CI.

Note that this PR is possible because of #139. Examples 5 and 7 have to account for the discrepancy between the PIE-Elf addresses and the virtual addresses supplied by Ghidra. A more elegant solution, which would not need this workaround, is beyond the scope of this PR: for now, it is advantageous to be able to run the examples in CI and have users run them with publicly available software (examples previously used Binary Ninja backend).

@SamL98, FYI. @EdwardLarson

### DIFF
--- a/examples/ex5_binary_extension.py
+++ b/examples/ex5_binary_extension.py
@@ -38,8 +38,7 @@ Obey the kitteh 😼
 import argparse
 import os
 
-import ofrak_binary_ninja
-import ofrak_capstone
+import ofrak_ghidra
 from ofrak import OFRAK, OFRAKContext, ResourceFilter, ResourceAttributeValueFilter
 from ofrak.core import (
     BinaryPatchModifier,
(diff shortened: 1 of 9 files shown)


EdwardLarson (approved):

examples/ex5_binary_extension.py
```diff
@@ -103,13 +102,15 @@ async def main(ofrak_context: OFRAKContext, file_path: str, output_file_name: st
...
     )
+    main_cb_assembly = await main_cb.get_assembly()
     lea_instruction = await main_cb.resource.get_only_descendant_as_view(
         v_type=Instruction,
         r_filter=ResourceFilter(
             attribute_filters=(ResourceAttributeValueFilter(Instruction.Mnemonic, "lea"),)
         ),
     )
-    kitty_offset = empty_vaddr - lea_instruction.virtual_address - 7
+    ghidra_empty_vaddr = empty_vaddr + 0x100000  # Ghidra bases PIE executables at 0x100000
```
SamL98, line 112: after running `unpack_recursively`, you should be able to get the code region of the new segment (or view it as one? idk) to not hardcode the virtual address. but that's probably too verbose for here.

Issue #153, opened 2022-12-28
OpenWRT TRX unpacker fails to unpack when rootfs_offset is null


RLS-Rylan:
**What is the problem? (Here is where you provide a complete Traceback.)**
Many TRX images seem to zero out the `rootfs_offset` field in the TRX header, and store data in a UBIFS filesystem at `kernel_offset` within the image. When this happens, the `OpenWrtTrxKernel` is empty, and the `OpenWrtTrxRootfs` object contains all data from offset 0 to the end of the file.

**Please provide some information about your environment.**
`redballoonsecurity/ofrak/binaryninja`

**If you've discovered it, what is the root cause of the problem?**

**How often does the issue happen?**
This seems to affect all targets except `bcrm47xx/mips74k` across all versions of OpenWRT

**What are the steps to reproduce the issue?**
Running the following function on a (bcm53xx target)[https://downloads.openwrt.org/releases/19.07.0/targets/bcm53xx/generic/openwrt-19.07.0-bcm53xx-buffalo-wxr-1900dhp-squashfs.trx] and a (bcrm47xx)[https://downloads.openwrt.org/releases/21.02.0-rc1/targets/bcm47xx/mips74k/openwrt-21.02.0-rc1-bcm47xx-mips74k-asus_rt-n10u-b-squashfs.trx] will produce the output below
```python
async def unpack_image(resource):
    resource.add_tag(OpenWrtTrx)
    await resource.save()
    await resource.unpack()

    trx_view = await resource.view_as(OpenWrtTrx)
    trx_header = await trx_view.get_header()
    print(f'loader_offset: {hex(trx_header.trx_loader_offset)}')
    print(f'kernel_offset: {hex(trx_header.trx_kernel_offset)}')
    print(f'rootfs_offset: {hex(trx_header.trx_rootfs_offset)}')

    kernel = await resource.get_only_descendant_as_view(
        OpenWrtTrxKernel, r_filter=ResourceFilter(tags=(OpenWrtTrxKernel,))
    )
    print(f'kernel data length: {await kernel.resource.get_data_length()}')

    print(await resource.summarize_tree())
```
```
openwrt-19.07.0-bcm53xx-buffalo-wxr-1900dhp-squashfs.trx
loader	offset: 0x1c
kernel_offset: 0x400000
rootfs_offset: 0x0
kernel data length: 0
[...]

PR #161, opened 2023-01-06
WIP: Add check for null rootfs offset

RLS-Rylan:
**Link to Related Issue(s)**
#153 

Added logic to skip over rootfs_offset if null, and to repack with null offset if the resource is empty.


ofrak_core/ofrak/core/openwrt.py
```diff
@@ -212,29 +212,32 @@ async def unpack(self, resource: Resource, config=None):
...
-                data=data[trx_header.trx_rootfs_offset :],
+                data=data[trx_header.trx_rootfs_offset:] if trx_header.trx_rootfs_offset > 0 else [],
             )
         else:  # TRX Version 2
             # Create rootfs child
             await resource.create_child(
                 tags=(OpenWrtTrxRootfs,),
-                data=data[trx_header.trx_rootfs_offset : trx_header.trx_binheader_offset],
+                data=data[
+                     trx_header.trx_rootfs_offset: trx_header.trx_binheader_offset] if trx_header.trx_rootfs_offset > 0 else [],
```
EdwardLarson, line 235: The data should be an empty bytestring `b""`, not an empty list `[]`, when rootfs is null

ofrak_core/ofrak/core/openwrt.py
```diff
@@ -433,10 +436,17 @@ async def pack(self, resource: Resource, config=None):
...
+        rootfs_data = await rootfs.resource.get_data()
+        if len(rootfs_data) > 0:
+            repacked_data_l.append(rootfs_data)
+        else:
+            trx_rootfs_offset = 0
 
         if header.get_version() == OpenWrtTrxVersion.VERSION2:
-            trx_binheader_offset = trx_rootfs_offset + len(repacked_data_l[-1])
+            if trx_rootfs_offset > 0:
+                trx_binheader_offset = trx_rootfs_offset + len(repacked_data_l[-1])
```
EdwardLarson, line 447: Are these supposed to be 2 different cases and there's a typo, or should they be merged into one?

EdwardLarson (changes requested):
Thanks for taking this on Rylan! I left a comment on a couple of things that need to be fixed. It also looks like the linter is failing, you can fix those manually or set up the [pre-commit hooks](https://ofrak.com/docs/contributor-guide/getting-started.html) and then run `pre-commit run --all-files` to take care of them all automatically.

We should also add a test case for this. I'll try to get to adding a test case based on the one in #153 today.

ofrak_core/ofrak/core/openwrt.py
RLS-Rylan: I believe that these do need to be two separate cases - `trx_binheader_offset` should always be set to the previous offset + the size of the previous data segment, but this could be either the rootfs or the kernel depending on whether or not the rootfs is included.

PR #163, opened 2023-01-09
Remove unused (duplicated) nginx configuration

rbs-jacob:

There was an unused `nginx.conf` file. This PR removes it.

@EdwardLarson


EdwardLarson:
Looks like that file is actually used in the `proxy.Dockerfile` in that same directory, used to build the `rbs/ofrak-gui-proxy` image which is used in the `compose.yml`. I don't think we can delete just the nginx.conf at this point.

PR #167, opened 2023-01-10
PIE support for Angr

SamL98:
Uses PIE detection for ELF's to fixup the virtual address of CodeRegion's. Support for other filetypes (PE, Mach-O, etc.) should be added as needed.

### DIFF
--- a/disassemblers/ofrak_angr/ofrak_angr/components/angr_analyzer.py
+++ b/disassemblers/ofrak_angr/ofrak_angr/components/angr_analyzer.py
@@ -11,6 +11,9 @@ from ofrak.resource import Resource
 import angr.project
 from ofrak_angr.components.identifiers import AngrAnalysisResource
 from ofrak_angr.model import AngrAnalysis
+from ofrak.component.modifier import Modifier
+from ofrak.core import Program, CodeRegion
+from ofrak import ResourceFilter
 
 
 LOGGER = logging.getLogger(__file__)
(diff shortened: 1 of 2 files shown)


disassemblers/ofrak_angr/ofrak_angr/components/angr_analyzer.py
```diff
@@ -69,3 +73,38 @@ def _create_dependencies(
...
+        if root_resource.has_tag(Elf):
+            elf_header = await root_resource.get_only_descendant_as_view(
+                ElfHeader, r_filter=ResourceFilter(tags=[ElfHeader])
+            )
+
+            if elf_header is not None and elf_header.e_type == ElfType.ET_DYN.value:
+                fixup_address = True
+        else:
+            LOGGER.warning(
+                f"Have not implemented PIE-detection for {root_resource}. The address of {code_region} will likely be incorrect."
```
EdwardLarson, line 102: I'm not sure if we should say "likely" incorrect here. I suppose most of the random ELFs one might come across are PIE, but is the same true of other files.

The uncertainty here again points to why I think it would be ideal to check specifically if the memory map in Angr and the ELF headers are different. While CodeRegions are only executable regions, we could get the lowest `ProgramSection` and compare that to the start address in Angr (`obj.min_addr` right?). This method would 1) not only apply to ELFs, but all programs, 2) not assume that only PIE can cause a memory map mismatch, 3) check _directly_ for the memory map mismatch. Is there a reason that wouldn't work?

SamL98: Yeah I think that would work

SamL98: I tried this and it actually does not quite work because of sections not loaded into the binary (such as .comment or .strtab). These sections have a virtual address of 0 and hence ruin the minimum calculation. As I believe there's no way in OFRAK to distinguish between sections we would and would not want to count in this calculation, I think the PIE-detecting route is our best option.

EdwardLarson: Until we know more about the Angr memory mapping APIs, we'll go with this. Supporting PIE ELF files is better than not supporting PIE files.

disassemblers/ofrak_angr/ofrak_angr/components/blocks/unpackers.py
```diff
@@ -24,14 +24,19 @@ class AngrCodeRegionUnpacker(CodeRegionUnpacker):
...
-        cr_vaddr_range = cr_view.vaddr_range()
 
         ## Run AngrAnalyzer
         root_resource = await resource.get_only_ancestor(
             ResourceFilter(tags=[AngrAnalysisResource], include_self=True)
         )
         angr_analysis = await root_resource.analyze(AngrAnalysis)
 
+        ## Fixup the CodeRegion's virtual address after analyzing with angr.
+        # await resource.run(AngrCodeRegionModifier, AngrCodeRegionModifierConfig(angr_analysis))
```
EdwardLarson, line 35:
```suggestion
```

EdwardLarson (changes requested):
One little fixup then this is g2g

EdwardLarson (approved):

PR #172, opened 2023-01-11
Injection fixes

rbs-afflitto:
Various fixes related to patch maker and binary injections:
- Injecting multiple binaries results in an error from Capstone - this is due to stale `Instruction` children having the memory ranges for the previous instructions. When the `MemoryRange` is auto-analyzed a second time, the new instructions have incorrect lengths, and are parsed as invalid instructions. This fix changes `SegmentInjectorModifier` to delete the stale `Instruction` children after each injection
- Added additional compiler options for the AVR gcc toolchain
- Added ability to set the physical address in the `LiefAddSegmentModifier`
- Fix a null pointer reference bug in Ghidra scripts

@Edward-Larson / @EdwardLarson do we want to use this fix for `SegmentInjectorModifier` or should we fix this another way?
CC @whyitfor @andresito00


ofrak_core/ofrak/core/patch_maker/modifiers.py
```diff
@@ -212,8 +214,20 @@ async def modify(self, resource: Resource, config: SegmentInjectorModifierConfig
...
-            await injected_resource.run(BinaryInjectorModifier, injection_config)
+            result = await injected_resource.run(BinaryInjectorModifier, injection_config)
+            parents_to_delete = list(
+                filter(lambda r: r.get_id() in result.resources_modified, all_decendants)
+            )
+            to_delete = []
+            for parent in parents_to_delete:
+                to_delete.extend(
+                    list(
+                        await parent.get_descendants(r_filter=ResourceFilter.with_tags(Instruction))
```
EdwardLarson, line 227: Shouldn't this also delete `BasicBlock`s at the very least and possibly `ComplexBlock`s as well? Do we even need any filtering here? I feel like anything that is wholly modified should have all of its children deleted.

EdwardLarson:
@rbs-afflitto I don't mind this change/fix for `SegmentInjectorModifier`. I think I mentioned on another channel that OFRAK doesn't track when/how a resource's bounds (or its very existence) depends on some data which can be changed, but those relationships are very real. I'm still thinking about how to handle that more generally, but I like spot-fixing modifiers run head first into this problem.

whyitfor:
@rbs-afflitto, can you add corresponding updates to the "Unreleased" portions of the changelogs for ofrak-patch-maker and ofrak to this PR.

ofrak_core/ofrak/core/patch_maker/modifiers.py
rbs-afflitto: Makes sense, I removed the resource filter as I think it makes sense to delete all the children as well.

andresito00:
@EdwardLarson I've gone ahead and resolved the .md file merge conflicts. If the checks pass and you've completed your review please LGTM :)

andresito00 (approved):

PR #173, opened 2023-01-12
UBIFS packer / unpacker

kibo-redballoon:
This PR contributes a UBIFS packer / unpacker and associated tests for OFRAK. It uses [mtd-utils](http://www.linux-mtd.infradead.org/) and [ubireader](https://github.com/jrspruitt/ubi_reader) to perform resource packing and unpacking.

@whyitfor


ofrak_core/ofrak/core/ubifs.py
```diff
@@ -0,0 +1,176 @@
...
+    orph_lebs: int
+    log_lebs: int
+
+
+@dataclass
+class Ubifs(GenericBinary, FilesystemRoot):
+    """
+    http://www.linux-mtd.infradead.org/doc/ubifs.html
+    Minimum number of bytes per transaction (see http://www.linux-mtd.infradead.org/doc/ubi.html#L_min_io_unit)
+    Size of logical erase blocks and superblock parameters required for repacking.
```
whyitfor, line 49: Can you make this docstring a declarative sentence about what this is? https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/ofrak/core/uimage.py#L193 is a good example.

Also, if you want to describe the dataclasses attributes, you should use the following notation:

```
:ivar min_io_size: minimum number of bytes per transaction (see http://www.linux-mtd.infradead.org/doc/ubi.html#L_min_io_unit)
:ivar leb_size: size of logical erase blocks and superblock parameters required for repacking.
:ivar superblock: A SuberblockNode
``` 

Both of this make the auto-generated docs easier to read, for example: https://ofrak.com/docs/reference/ofrak/core/uimage.html

ofrak_core/ofrak/core/ubifs.py
```diff
@@ -0,0 +1,176 @@
...
+from ubireader.utils import guess_leb_size
+
+LOGGER = logging.getLogger(__name__)
+
+MKFS_UBIFS_TOOL = ComponentExternalTool(
+    "mkfs.ubifs",
+    "http://www.linux-mtd.infradead.org/faq/ubifs.html",
+    install_check_arg="--help",
+    apt_package="mtd-utils",
+    brew_package="",  # This isn't compatible with macos, but there may be an alternative tool to do the bidding.
```
whyitfor, line 30:
```suggestion
    brew_package=None,  # This isn't compatible with macos, but there may be an alternative tool to do the bidding.
```
Should be set to None if we don't have one. See https://ofrak.com/docs/reference/ofrak/model/component_model.html#ofrak.model.component_model.ComponentExternalTool.

whyitfor (changes requested):
Looking good!

See two comments that need to be addressed.

EdwardLarson (changes requested):
Update changelog please

EdwardLarson (approved):

ofrak_core/ofrak/core/ubifs.py
```diff
@@ -0,0 +1,204 @@
...
+    As part of an UBI image, re-packing an UBIFS image requires the 'min_io_size' and 'leb_size' properties that
+    are stored as part of the UBI header (https://elixir.bootlin.com/linux/v6.1.7/source/drivers/mtd/ubi/ubi.h#L441).
+
+    Each UBIFS image has a superblock which is encoded in OFRAK as a SuperblockNode.
+
+    Some documentation about UBIFS layout can also be found here:
+    https://www.kernel.org/doc/Documentation/filesystems/ubifs-authentication.rst
+    http://www.linux-mtd.infradead.org/doc/ubifs.html
+
+    :cvar min_io_size: Minimum number of bytes per I/O transaction (see http://www.linux-mtd.infradead.org/doc/ubi.html#L_min_io_unit)
```
whyitfor, line 73: Have you tested that `cvar` renders correctly in the built docs? I know that `var` and `ivar` both work

I think technically these are not class variables -- it looks like class variables are specifically skipped by dataclass: https://docs.python.org/3/library/dataclasses.html#class-variables. 

Let's just change all of these to `:var ...`?

whyitfor (changes requested):
Docstrings are looking great! 

See one comment about dataclass docstrings.

PR #174, opened 2023-01-12
Enables permission re-mapping from patch to destination program 

andresito00:
**Link to Related Issue(s)**
N/A

Adds a new parameter to `allocate_bom` which specifies possible destination permissions for the current patch segment.
i.e. Developers can specify that .rodata segments of patches may be stored in the .text section of programs.

@EdwardLarson @whyitfor

### DIFF
--- a/ofrak_core/ofrak/core/free_space.py
+++ b/ofrak_core/ofrak/core/free_space.py
@@ -1,7 +1,7 @@
 from collections import defaultdict
 from dataclasses import dataclass
 from itertools import chain
-from typing import List, Tuple, Dict, Optional, Iterable
+from typing import List, Tuple, Dict, Optional, Iterable, Mapping
 
 from immutabledict import immutabledict
 
@@ -114,12 +114,16 @@ class Allocatable(ResourceView):
     async def allocate_bom(
         self,
         bom: BOM,
+        permission_map: Optional[Mapping[MemoryPermissions, Iterable[MemoryPermissions]]] = None,
     ) -> PatchRegionConfig:
         """
         Responsible for allocating the patches if free memory is required and
         providing details about where space was made.
 
         :param bom:
+        :param permission_map: a map that assigns patch segment-local permissions to possible
+        destination permissions. available memory pool is evaluated in order!
+        Ex: a developer wants to enable allocation of RO strings from .rodata in an RX .text section.
 
         :return: information required to generate the linker directive script
         """
@@ -134,16 +138,26 @@ class Allocatable(ResourceView):
         for obj, segment in segments_to_allocate:
             vaddr, final_size = 0, 0
             if segment.length > 0:
-                allocs = await self.allocate(
-                    segment.access_perms,
-                    segment.length,
-                    min_fragment_size=segment.length,
-                    alignment=bom.segment_alignment,
+                possible_perms = permission_map[segment.access_perms] or (segment.access_perms,)
+                for candidate_permissions in permission_map[segment.access_perms]:
+                    try:
+                        allocs = await self.allocate(
+                            candidate_permissions,
+                            segment.length,
+                            min_fragment_size=segment.length,
+                            alignment=bom.segment_alignment,
+                        )
+                        allocation = next(iter(allocs))
+                        vaddr = allocation.start
+                        final_size = allocation.length()
+                        break
+                    except FreeSpaceAllocationError:
+                        continue
+            if vaddr == 0 or final_size == 0:
+                raise FreeSpaceAllocationError(
+                    f"Could not find enough free space for access perms {possible_perms} and "
+                    f"length {segment.length}"
                 )
-                allocation = next(iter(allocs))
-                vaddr = allocation.start
-                final_size = allocation.length()
-
             segments_by_object[obj.path].append(
                 Segment(
                     segment_name=segment.segment_name,
--- a/ofrak_core/ofrak/core/free_space.py
+++ b/ofrak_core/ofrak/core/free_space.py
@@ -135,11 +135,15 @@ class Allocatable(ResourceView):
         # Allocate largest segments first
         segments_to_allocate.sort(key=lambda o_s: o_s[1].length, reverse=True)
         segments_by_object: Dict[str, List[Segment]] = defaultdict(list)
+        possible_perms = None
         for obj, segment in segments_to_allocate:
             vaddr, final_size = 0, 0
             if segment.length > 0:
-                possible_perms = permission_map[segment.access_perms] or (segment.access_perms,)
-                for candidate_permissions in permission_map[segment.access_perms]:
+                if permission_map is not None:
+                    possible_perms = permission_map[segment.access_perms]  # type: ignore
+                else:
+                    possible_perms = (segment.access_perms,)
+                for candidate_permissions in possible_perms:
                     try:
                         allocs = await self.allocate(
                             candidate_permissions,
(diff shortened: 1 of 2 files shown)


ofrak_core/ofrak/core/free_space.py
```diff
@@ -134,16 +138,26 @@ async def allocate_bom(
         for obj, segment in segments_to_allocate:
             vaddr, final_size = 0, 0
             if segment.length > 0:
-                allocs = await self.allocate(
-                    segment.access_perms,
-                    segment.length,
-                    min_fragment_size=segment.length,
-                    alignment=bom.segment_alignment,
+                possible_perms = permission_map[segment.access_perms] or (segment.access_perms,)
+                for candidate_permissions in permission_map[segment.access_perms]:
```
whyitfor, line 142: Should this be something like:
```suggestion
                possible_perms = permission_map[segment.access_perms] if permission_map is not None else (segment.access_perms,)
                for candidate_permissions in possible_perms:
```

ofrak_core/ofrak/core/free_space.py
```diff
@@ -114,12 +114,16 @@ async def allocate(
     async def allocate_bom(
         self,
         bom: BOM,
+        permission_map: Optional[Mapping[MemoryPermissions, Iterable[MemoryPermissions]]] = None,
```
EdwardLarson, line 117: Do we need the full flexibility that this map allows? For example, allowing executable segments to be allocated to (non-executable) read/write space. There seem to be only two legit fallback cases:

* R -> RW
* R -> RX

It would be less flexible, but maybe usage would be more clear and the options more relevant if we just had two optional flags, `allow_r_in_rw` and `allow_r_in_rx` (or similar self-documenting names).

andresito00: I go back and forth.

If someone asked for the ability to map their patch RX into their destination's RW would we say no? 

(considering beyond Linux platforms or ELF formats...)

andresito00: I guess we enable them to do that with your bools.

EdwardLarson: Yeah, the flexibility to deal with unusual architectures or even buggy toolchains is a big bonus. Also being able to set a prioritization of RX over RW or vice versa. I think these come at the disadvantage of the interface being a little harder to grok, so there's the risk that users won't find this option, even when it would be useful. But I don't think that has to block this PR.

Another alternative would be adding 3 new options instead of one: `r_overflow_to`, `rx_overflow_to`, and `rw_overflow_to`. Each one would be an `Iterable[MemoryPermissions]`, essentially unfolding the `permission` map into the function arguments, defaulting to `None` or an empty Tuple.  I think this will be a little easier to understand for a user.

Again, that doesn't have to block this PR. Everything else looks good. I'll leave this comment, if you like the idea, run with it. Otherwise I'll approve and you can merge at your convenience.

EdwardLarson (approved):

ofrak_core/ofrak/core/free_space.py
```diff
@@ -131,19 +135,36 @@ async def allocate_bom(
...
+                    except FreeSpaceAllocationError:
+                        continue
+            if vaddr == 0 or final_size == 0:
+                if possible_perms:
+                    raise FreeSpaceAllocationError(
+                        f"Could not find enough free space for access perms {possible_perms} and "
+                        f"length {segment.length}"
+                    )
+                else:
+                    raise FreeSpaceAllocationError(f"No provided segments longer than 0 bytes.")
```
whyitfor, line 167: The previous implementation allowed for creating segments with vaddr of 0 and final_size of 0. To preserve this, this entire block should be indented so it runs only if `segment.length > 0`.

ofrak_core/ofrak/core/free_space.py
```diff
@@ -131,19 +135,36 @@ async def allocate_bom(
         # Allocate largest segments first
         segments_to_allocate.sort(key=lambda o_s: o_s[1].length, reverse=True)
         segments_by_object: Dict[str, List[Segment]] = defaultdict(list)
+        possible_perms = None
```
whyitfor, line 138:
```suggestion
```
This shouldn't be needed if 160-167 are correctly indented.

ofrak_core/ofrak/core/free_space.py
```diff
@@ -131,19 +135,36 @@ async def allocate_bom(
...
-                    segment.length,
-                    min_fragment_size=segment.length,
-                    alignment=bom.segment_alignment,
-                )
-                allocation = next(iter(allocs))
-                vaddr = allocation.start
-                final_size = allocation.length()
-
+                if permission_map is not None:
+                    possible_perms = permission_map[segment.access_perms]  # type: ignore
```
whyitfor, line 143:
```suggestion
                    possible_perms = permission_map[segment.access_perms]
```

Removing the type ignore does not seem to impact the results of running `make inspect`.

ofrak_core/ofrak/core/free_space.py
whyitfor: All of the relevant changes create the following diff:

```
diff --git a/ofrak_core/ofrak/core/free_space.py b/ofrak_core/ofrak/core/free_space.py
index ab4150a..2bae549 100644
--- a/ofrak_core/ofrak/core/free_space.py
+++ b/ofrak_core/ofrak/core/free_space.py
@@ -135,12 +135,11 @@ class Allocatable(ResourceView):
         # Allocate largest segments first
         segments_to_allocate.sort(key=lambda o_s: o_s[1].length, reverse=True)
         segments_by_object: Dict[str, List[Segment]] = defaultdict(list)
-        possible_perms = None
         for obj, segment in segments_to_allocate:
             vaddr, final_size = 0, 0
             if segment.length > 0:
                 if permission_map is not None:
-                    possible_perms = permission_map[segment.access_perms]  # type: ignore
+                    possible_perms = permission_map[segment.access_perms]
                 else:
                     possible_perms = (segment.access_perms,)
                 for candidate_permissions in possible_perms:
@@ -157,14 +156,14 @@ class Allocatable(ResourceView):
                         break
                     except FreeSpaceAllocationError:
                         continue
-            if vaddr == 0 or final_size == 0:
-                if possible_perms:
-                    raise FreeSpaceAllocationError(
-                        f"Could not find enough free space for access perms {possible_perms} and "
-                        f"length {segment.length}"
-                    )
-                else:
-                    raise FreeSpaceAllocationError(f"No provided segments longer than 0 bytes.")
+                if vaddr == 0 or final_size == 0:
+                    if possible_perms:
+                        raise FreeSpaceAllocationError(
+                            f"Could not find enough free space for access perms {possible_perms} and "
+                            f"length {segment.length}"
+                        )
+                    else:
+                        raise FreeSpaceAllocationError(f"No provided segments longer than 0 bytes.")
             segments_by_object[obj.path].append(
                 Segment(
                     segment_name=segment.segment_name,
```

whyitfor (changes requested):
See comments. Additionally, this PR should update the patch maker change log with an "Added".

ofrak_core/ofrak/core/free_space.py
whyitfor: IMHO `permission_map` seems like the easiest option to me -- it creates only one optional argument. If a user runs into a permission mapping problem, I think they are in a situation where they can figure out how to use it.

I'd vote for leaving it in this PR and returning to this if it causes issues down the line.

andresito00: What Wyatt is outlines here is the original reason for why we've done it this way. Let's go with our first impression of what is "simple" here (fewer parameters, more possible expressiveness), and if we run into to trouble we can adjust to something like Ed's suggestion.

whyitfor (approved):
Looks good!

PR #181, opened 2023-01-20
APIs to integrate GUI more deeply into OFRAK workflows

EdwardLarson:
Add APIs to open GUI in a script or after CLI commands complete.

**Link to Related Issue(s)**

* Add flags to open GUI after `inject` or `unpack` completes
* Add `-r` flag to `unpack` command, to unpack recursively
* Add `get_current_ofrak_context` which... gets the current OFRAK context, so this can be done in any arbitrary place in a script or component and one doesn't have to keep passing the context around. It uses a global variable for this and is therefore fairly brittle, and would break if two contexts are created concurrently. I added a check and error raise for this so it won't be a silent bug.
* Add `open_gui` function to simply start the server and open a browser window. It returns the server, so it's up to the user to do `run_until_complete` if they want to, but it leaves it open for other conditions to exit (not implemented).
* Add a `console_scripts` entrypoint for ofrak so that the CLI can be run a normal CLI tool, not only as a python3 module (i.e. you can run `ofrak unpack ...` instead of `python3 -m ofrak unpack ...`

New options for `unpack`:
```
optional arguments:
  -h, --help            show this help message and exit
  -o OUTPUT_DIRECTORY, --output_directory OUTPUT_DIRECTORY
                        Directory to write unpacked resource tree to. If no
                        directory is given, a new one will be created in the
                        same directory as the file being unpacked.
  --recursive, -r       Unpack recursively: all resources unpacked from the
                        root will be unpack, as well as all resources unpacked
                        from those, and so.
  --gui                 Open the OFRAK GUI after unpacking
  -gH GUI_HOSTNAME, --gui-hostname GUI_HOSTNAME
                        Set GUI server host address.
  -gp GUI_PORT, --gui-port GUI_PORT
                        Set GUI server host port.
...
```

Example usage of `open_gui` etc. in a script:

```
async def main(context: OFRAKContext, outfile):
    root = await load_and_unpack(context)
    await add_program_attributes(root)
    ofrak_gui = await open_gui("0.0.0.0", 80, root)
    source_bundle = await add_source_bundle(root)
    ofrak_gui.open_resource_in_browser(source_bundle.resource)
    await perform_function_replacement(root, source_bundle)
    await root.flush_to_disk(outfile)

    await ofrak_gui.run_until_cancelled()
```

@rbs-jacob @whyitfor

### DIFF
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 - Move GUI server to `ofrak_core`, startup GUI through CLI, add testing for server, make GUI pip installable. [#168](https://github.com/redballoonsecurity/ofrak/pull/168)
   - `python -m ofrak gui` starts the OFRAK GUI server.
 - UBI and UBIFS filesystem analyzers / unpackers / packers added. [#173](https://github.com/redballoonsecurity/ofrak/pull/173), [#177](https://github.com/redballoonsecurity/ofrak/pull/177)
+- Add APIs to open GUI in a script or after CLI commands complete.
 
 ### Changed
 - Refactored `DataService` internals to more efficiently find resources affected by patches [#140](https://github.com/redballoonsecurity/ofrak/pull/140)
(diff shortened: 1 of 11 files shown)


whyitfor (approved):
I really like this feature! I left one minor comment on changelog that you can address as you see fit.

PR #182, opened 2023-01-20
add GenericBinary tag to children of UBI resources if they have data,…

EdwardLarson:
… so auto identifying and unpacking them further works

add GenericBinary tag to children of UBI resources if they have data, so auto identifying and unpacking them further works

**Link to Related Issue(s)**
Without this tag, auto identifying and unpacking them does not work. So if there is e.g. a squashfs filesystem in a UbiVolume, it wouldn't be identified and unpacked.

dd GenericBinary tag to children of UBI resources if they have data

@whyitfor


whyitfor (approved):
LGTM

PR #185, opened 2023-01-24
Ignore .rela sections in FEM created by placeholder

dannyp303:
The same way we do with the .got section, ignore .rela sections in patch maker FEM while injecting.


whyitfor:
Can you update the patchmaker changelog under "Unreleased"? Probably should be "changed" or "fixed" depending on how you see it: https://keepachangelog.com/en/1.0.0/.

PR #187, opened 2023-01-24
Issue 153 openwrt

SamL98:
**Link to Related Issue(s)**
https://github.com/redballoonsecurity/ofrak/issues/153

The OpenWRT documentation [here](https://openwrt.org/docs/techref/headers) does not describe the TRX file format in the general case. The code [here](https://git.openwrt.org/?p=project/firmware-utils.git;a=blob_plain;f=src/otrx.c;hb=HEAD) shows how TRX files are actually created: up to three images (or partitions), possibly padded, are added to the file. These partitions do not necessarily correspond to loader, kernel, etc. described in the wiki or are not necessarily in that order. Therefore the TRX packer/unpacker needs to be much more general than it previously was.


ofrak_core/ofrak/core/openwrt.py
```diff
@@ -115,34 +106,14 @@ def get_header_length(self) -> int:
             raise ValueError(f"Unknown OpenWrt TRX version: {self.trx_version}")
 
 
-class OpenWrtTrxLzmaLoader(GenericBinary):
```
EdwardLarson, line 118: The packer still uses these classes, are they supposed to be deleted?

EdwardLarson (changes requested):
Needs a CHANGELOG entry!

EdwardLarson:
#200 is the fix for this, it's #197 (still waiting on review @whyitfor )

EdwardLarson:
With #197 merged in, one of you should be able to merge in master and make the tests pass. I don't have push access to the fork so I can't do it myself.

EdwardLarson:
With no push access to the main branch, I updated this and made my own PR. Moving this issue once again... #216

PR #188, opened 2023-01-25
Added line to change ghidra's m68k syntax to gnu's at&t syntax for ne…

dannyp303:
Added line to change ghidra's m68k syntax to gnu's at&t syntax for negative imms


andresito00:
@dannyp303 Please update the relevant CHANGELOG.md under a "Fixed" heading. Otherwise this should be good to go.

EdwardLarson (approved):
Andres is right that the changelog should be updated, but there is no existing changelog for ofrak_ghidra at the moment. LGTM.

PR #190, opened 2023-01-26
Feature/pip install ofrak angr capstone

EdwardLarson:
**Link to Related Issue(s)**
ofrak_angr and ofrak_capstone have purely Python dependencies, but weren't pip installable.

There was also a bug with ofrak_angr's complex block unpacker, where it didn't appropriately catch and warn when a basic block would be created outside the parent's range (as the ofrak_binary_ninja unpacker does)

Prep ofrak_angr and ofrak_capstone for pypi release, fix that bug.


whyitfor (approved):
Looks good

PR #194, opened 2023-01-28
Dockerfile optimization

ANogin:
Optimize Dockerfiles for better caching and faster builds.

**Link to Related Issue(s)**
N/A

Install Python requirements in `base.Dockerfile` before OFRAK files are added to the container, to improve cache utilization when container is rebuilt after OFRAK code is modified. Make it possible to double-check that no external requirements  are installed or modified during the `make install`/`make develop` step in `final.Dockerfile` and that all requirements are satisfied (and therefore consistent) at the end of that step.

@whyitfor


build_image.py
```diff
@@ -194,6 +194,8 @@ def create_dockerfile_base(config: OfrakImageConfig) -> str:
 
     dockerfile_base_parts += [
         "FROM python:3.7-bullseye@sha256:338ead05c1a0aa8bd8fcba8e4dbbe2afd0283b4732fd30cf9b3bfcfcbc4affab",
+        "RUN apt-get -y update",
```
whyitfor, line 197: You'll need to install `pre-commit` to run the pre-commit hooks: https://ofrak.com/docs/contributor-guide/getting-started.html#pre-commit.

whyitfor, line 197: The drawback of this proposed change is that the update step is cached separate from the install step.

[Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) generally recommends running update with the install step. The `Dockerstub` approach we have taken generally should group each `apt-get install` in each `Dockerstub` into one step (the exception here is `ofrak-patch-maker`, where we've maintained different steps to make it clear which dependency matches each toolchain).

(And yes, we probably should be running `rm -rf /var/lib/apt/lists/*` to keep the image a little bit smaller).

frontend/Dockerstub
```diff
@@ -1,3 +1,6 @@
+RUN apt-get install --yes nginx
+RUN python3 -m pip install --upgrade aiohttp
```
whyitfor, line 2: This `aiohttp` install can be removed since `ofrak` as of 2.1.0 has a dependency on `aiohttp`.

build_image.py
ANogin: Yes, thanks, saw that afterwards.

ANogin: Makes sense - I reverted this change, and instead made sure that each `apt-get install` is preceded by `apt-get update`

frontend/Dockerstub
ANogin: Removed - with [3a537e8](https://github.com/redballoonsecurity/ofrak/pull/194/commits/3a537e843f3320898576f5104460ca9d932a36be) now the Python dependencies are installed in `base.Dockerfile` based on the requirements files.

build_image.py
```diff
@@ -224,8 +253,20 @@ def create_dockerfile_finish(config: OfrakImageConfig) -> str:
...
+        ### The --network="none" is very helpful in catching all kinds of issues,
+        ### and should probably be enabled eventually.
+        ### Unfortunately it cases problems when a package has a pyproject.toml -
+        ### see https://github.com/pypa/pip/issues/6482#issuecomment-1414208986
+        ### A workaround is to use `pip install --no-build-isolation` or
+        ### `pip install --no-use-pep517` but that starts getting messy.
+        ### Leaving it as commented code here to make it easier for people
+        ### to temporarily enable it.
+        #'# We use --network="none" to ensure all dependencies were installed in base.Dockerfile\n',
+        #'RUN --network="none" make $INSTALL_TARGET\n\n',
```
whyitfor, line 267:
```suggestion
```
This isn't a comment explaining what the code does, but an aside about future improvements. Can you memorialize it as a comment on the PR or, if you think it merits one, an issue on the GitHub tracker?

build_image.py
```diff
@@ -186,25 +187,53 @@ def create_dockerfile_base(config: OfrakImageConfig) -> str:
...
-        dockerfile_base_parts.append(dockerstub)
+        dockerfile_base_parts += [f"### {dockerstage_path}", dockerstub]
 
     dockerfile_base_parts += [
         "FROM python:3.7-bullseye@sha256:338ead05c1a0aa8bd8fcba8e4dbbe2afd0283b4732fd30cf9b3bfcfcbc4affab",
+        "",
     ]
+
+    requirement_suffixes = [""]
+    if config.install_target == InstallTarget.DEVELOP:
```
whyitfor, line 203:
```suggestion
    if config.install_target is InstallTarget.DEVELOP:
```
Enum members are compared by identity: https://docs.python.org/3/howto/enum.html#comparisons.

whyitfor (changes requested):

build_image.py
ANogin: Done

disassemblers/ofrak_capstone/setup.py
```diff
@@ -18,12 +18,24 @@ def run(self):
...
+def read_requirements(requirements_path):
+    python_deps = []
+    with open(requirements_path) as file_handle:
+        for line in file_handle:
+            line = line.strip()
+            if line and line[0] != "#":
+                python_deps.append(line)
+    return python_deps
+
+
```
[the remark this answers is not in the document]
ANogin: I updated the core to use `pkg_resources`, but not `pathlib.Path` (is there any advantage over the plain `open`?). I also kept the function form, for consistency with those `setup.py` files that have to read multiple `requirements*.txt` files.

build_image.py
ANogin: I've made `--network="none"` this a configurable option (off by default).

build_image.py
```diff
@@ -224,8 +257,19 @@ def create_dockerfile_finish(config: OfrakImageConfig) -> str:
...
-    dockerfile_finish_parts.append(f'RUN printf "{develop_makefile}" >> Makefile\n')
-    dockerfile_finish_parts.append("RUN make $INSTALL_TARGET\n\n")
+    dockerfile_finish_parts += [
+        f'RUN printf "{develop_makefile}" >> Makefile\n\n',
+    ]
+    if config.check_python_reqs:
+        dockerfile_finish_parts += [
+            '# We use --network="none" to ensure all dependencies were installed in base.Dockerfile\n',
+            'RUN --network="none" make $INSTALL_TARGET\n',
+            "RUN python3 -m pip check\n\n",
```
whyitfor, line 267: It's not clear to me that this belongs here.

If we decide to enforce that ofrak packages are installed from source (and not PyPI), then that seems like something that belongs in the Makefile.

Similarly, it's not clear to me that checking pip dependencies should happen as a build step -- this feels more like a test/validation step that happens after the image is built (again, if we decide that we want to do that).

The objective of this PR was to optimize Docker build -- I think that is accomplished without this `check_python_reqs` option. What do you think about removing it from this PR and discussing/tackling that as a separate issue?

ANogin: @whyitfor makes sense for `python3 -m pip check`, but `--network="none"` can only be done in a `Dockerfile`. ``python3 -m pip check` only helps catch mismatching dependencies (so yeah, maybe belongs in CI), but `--network="none"` also catches the case where the Docker build is not fully optimized (that is, dependencies that we'd want to install in `base.Dockerfile` were not installed there). But will take it out, and move to a separate PR.

whyitfor (approved):

EdwardLarson (approved):

PR #196, opened 2023-01-30
Enable running the GUI without opening the browser

rbs-jacob:

Enable running the GUI without opening the browser.

Sometimes, a user may not want to open the browser automatically when they start the GUI. This adds a command-line option to enable that feature.

@EdwardLarson

### DIFF
--- a/ofrak_core/ofrak/cli/command/gui.py
+++ b/ofrak_core/ofrak/cli/command/gui.py
@@ -2,7 +2,7 @@ import logging
 from argparse import ArgumentDefaultsHelpFormatter, Namespace
 
 from ofrak.cli.ofrak_cli import OfrakCommandRunsScript
-from ofrak.gui.server import open_gui, start_server
+from ofrak.gui.server import open_gui
 from ofrak.ofrak_context import OFRAKContext
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


ofrak_core/ofrak/cli/command/gui.py
```diff
@@ -31,9 +31,17 @@ def create_parser(self, ofrak_subparsers):
...
+            action="store_true",
+            help="Don't open the browser to the OFRAK GUI",
+        )
         self.add_ofrak_arguments(gui_parser)
         return gui_parser
 
     async def ofrak_func(self, ofrak_context: OFRAKContext, args: Namespace):  # pragma: no cover
-        server = await open_gui(args.hostname, args.port)
+        if not args.no_browser:
+            server = await open_gui(args.hostname, args.port)
```
EdwardLarson, line 44: I'd rather there be one consistent function we call to launch the GUI, and it should probably be `open_gui`. That function could just have an option to try to open it in the browser or not. I also think the print about where the GUI is being served should print regardless of whether the browser is opened or not.

ofrak_core/ofrak/cli/command/gui.py
```diff
@@ -31,9 +31,14 @@ def create_parser(self, ofrak_subparsers):
             help="Set GUI server host port.",
             default=8080,
         )
+        gui_parser.add_argument(
```
EdwardLarson, line 34: Add corresponding arguments to `unpack` and `identify` (GUI arguments are duplicated there for now, perhaps a future refactor can just nest all of these GUI arguments in those commands' arguments)

EdwardLarson (approved):

PR #197, opened 2023-01-31
Bugfix: data service internal state

EdwardLarson:
Fix bug in data service that can cause mangled internal state.

**Link to Related Issue(s)**
The internal data structure that the DataService uses to identify data models affected by a patch was not properly updated in some cases. This could cause strange errors down the line.

* Fix bugs in DataService
* Add a class decorator to make debugging internal state of service etc. easier.

@whyitfor

### DIFF
--- a/ofrak_core/pytest_ofrak/utils.py
+++ b/ofrak_core/pytest_ofrak/utils.py
@@ -11,7 +11,7 @@ def auto_validate_state(validator_function):
 
     Example usage:
     def validator_function(class_instance):
-        # validation logic
+        # validation logic, raising an error if the internal state is invalid
         ...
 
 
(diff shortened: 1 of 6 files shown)


ofrak_core/pytest_ofrak/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    class AutoValidatingImplementation(ClassWithInternalState):
+        pass
+
+    An instance of AutoValidatingImplementation now runs validator_function on itself after every
+    method call.
+
+    :param validator_function:
+
+    :return:
+    """
```
rbs-jacob, line 28: One thing that could be clearer in this docstring is that the validator function has to raise an exception if there is a problem. I could see a situation in which a user might think the validator has to return true or false. This should be an easy, minor clarification to make.

rbs-jacob (commented):
One very, very small comment on a docstring clarification. But otherwise, it looks good to me!

rbs-jacob (approved):

PR #199, opened 2023-02-01
Add GUI file dialog component

rbs-jacob:

This adds a "file browser" dialog component to the GUI.

This PR adds a "file browser" to the GUI and places it on the first screen when the GUI loads. This PR also includes a few small changes such as a warning for large files and a new default message when there are no previously-loaded resources for the start page dropdown.

The new start page looks like this:

image

@whyitfor

### DIFF
--- a/frontend/src/StartView.svelte
+++ b/frontend/src/StartView.svelte
@@ -93,11 +93,14 @@
     preExistingRootsPromise = new Promise(() => {}),
     tryHash = !!window.location.hash;
   let mouseX, selectedAnimal;
+  const warnFileSizeMb = 250;
 
   async function createRootResource(f) {
     if (
-      f.size > 1024 * 1024 * 250 &&
-      !window.confirm("Loading a large file may be slow. Are you sure?")
+      f.size > 1024 * 1024 * warnFileSizeMb &&
+      !window.confirm(
+        `Loading a large file (>${warnFileSizeMb}MB) may be slow. Are you sure?`
+      )
     ) {
       showRootResource = false;
       return;
--- a/frontend/src/StartView.svelte
+++ b/frontend/src/StartView.svelte
@@ -248,7 +248,10 @@
     {:then preExistingRootResources}
       {#if preExistingRootsPromise && preExistingRootsPromise.length > 0}
         <form on:submit|preventDefault="{choosePreExistingRoot}">
-          <select bind:value="{selectedPreExistingRoot}">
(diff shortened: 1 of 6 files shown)


frontend/src/StartView.svelte
```diff
@@ -67,19 +79,30 @@
...
   let dragging = false,
     selectedPreExistingRoot = null,
     preExistingRootsPromise = new Promise(() => {}),
     tryHash = !!window.location.hash;
   let mouseX, selectedAnimal;
 
   async function createRootResource(f) {
+    if (
+      f.size > 1024 * 1024 * 250 &&
+      !window.confirm("Loading a large file may be slow. Are you sure?")
```
whyitfor, line 100: Can the error message define what large means (1024 * 1024 * 250). This will allow a user to know what we consider "large" (they can truncate the file, etc.).

whyitfor (changes requested):
See comment below.

Furthermore, can you combine the two loading options ("Drag file or upload") into one thought.

Right now, there are three options:
1. Drag
2. Browse
3. Use existing

There really should be two high level options:
1. Upload (drag or browse)
2. Use existing

For example, something like https://css-tricks.com/drag-and-drop-file-uploading/,

rbs-jacob:
To address your PR comments @whyitfor, I changed it so that the start page has been reverted to look like the following.

Rather than there being a dedicated button for file uploading, the user can click anywhere (except for the dropdown/button for selecting pre-existing resources) to pop a file browser window.

image

EdwardLarson:
I don't think clicking anywhere will be obvious enough, we should have some text. I like how you've added the big "or" splitting the upload & select existing options. Maybe we can change that top text to "DRAG IN OR BROWSE TO A FILE TO ANALYZE" with the "BROWSE" text being a different-colored button to open the file browser?

rbs-jacob:
> I don't think clicking anywhere will be obvious enough, we should have some text. I like how you've added the big "or" splitting the upload & select existing options. Maybe we can change that top text to "DRAG IN OR BROWSE TO A FILE TO ANALYZE" with the "BROWSE" text being a different-colored button to open the file browser?

I tried adding the "OR BROWSE," but it made the text awkwardly long. Instead I've added a bit of smaller, explanatory text below.

image

EdwardLarson (approved):

whyitfor (approved):
Excited to use this!

PR #200, opened 2023-02-01
Bugfix: more selective criteria for resources to delete when injecting

EdwardLarson:
Delete narrower list of resources when injecting.

**Link to Related Issue(s)**
All descendants of the injected resource itself would be deleted when injecting segments, not just the descendants affected. The injected resource itself was of course modified, and all descendants of modified resources were marked to be deleted.

Only delete resources which are descendants, modified, AND not the injected resource.

@whyitfor @rbs-afflitto

### DIFF
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
 ### Fixed
 - Fix long-broken `OFRAK.set_id_service` [#198](https://github.com/redballoonsecurity/ofrak/pull/198)
-- Delete narrower list of resources when using `SegmentInjectorModifier` [#200](https://github.com/redballoonsecurity/ofrak/pull/200)
+- Fix bug in `SegmentInjectorModifier` that resulted in deleting more resources than necessary [#200](https://github.com/redballoonsecurity/ofrak/pull/200)
 
 ### Changed
 - GUI is much faster, especially for resources with hundreds of thousands of children [#191](https://github.com/redballoonsecurity/ofrak/pull/191)
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 
 ### Changed
 - GUI is much faster, especially for resources with hundreds of thousands of children [#191](https://github.com/redballoonsecurity/ofrak/pull/191)
+- Resources whose data gets modified are now listed in the `resource_modified` field of component results [#200](https://github.com/redballoonsecurity/ofrak/pull/200)
 
 
 ## [2.1.1](https://github.com/redballoonsecurity/ofrak/compare/ofrak-v2.1.0...ofrak-v2.1.1) - 2023-01-25
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
 ### Fixed
 - Fix long-broken `OFRAK.set_id_service` [#198](https://github.com/redballoonsecurity/ofrak/pull/198)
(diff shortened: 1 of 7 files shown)


whyitfor:
@EdwardLarson is there a simple test that we can add to make sure we don't have future regression on this bug?

ofrak_core/CHANGELOG.md
```diff
@@ -6,9 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
 ### Fixed
 - Fix long-broken `OFRAK.set_id_service` [#198](https://github.com/redballoonsecurity/ofrak/pull/198)
+- Delete narrower list of resources when using `SegmentInjectorModifier` [#200](https://github.com/redballoonsecurity/ofrak/pull/200)
```
whyitfor, line 9:
```suggestion
- Fix bug in `SegmentInjectorModifier` that resulted in deleting more resources than necessary [#200](https://github.com/redballoonsecurity/ofrak/pull/200)
```

ofrak_core/ofrak/component/abstract.py
```diff
@@ -149,6 +149,15 @@ async def run(
...
 
+        # Get resources modified by data patches
+        data_ids_to_models = await dependency_handler.map_data_ids_to_resources(
+            patch_result.data_id for patch_result in patch_results
+        )
+        for m in data_ids_to_models.values():
+            modified_resource_ids.add(m.id)
+
+        modified_resource_ids.difference_update(component_context.resources_deleted)
+
```
whyitfor, line 160:
```suggestion
        # Get resourced modified by data patches and add them to 
        # modified_resource_ids
        data_ids_to_models = await dependency_handler.map_data_ids_to_resources(
            patch_result.data_id for patch_result in patch_results
        )
        for m in data_ids_to_models.values():
            modified_resource_ids.add(m.id)
        # Filter deleted resources out of modified_resource_ids
        modified_resource_ids.difference_update(component_context.resources_deleted)
```
If I understand, we are:

1. Adding ids from patch results to `modified_resource_ids`
2. Filtering deleted resource from the modified_resource_ids list.

I think comments could make this clearer,  something like the suggestions. I also don't understand why the patches are not in the modified list already at this point?

ofrak_core/CHANGELOG.md
```diff
@@ -6,9 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 
 ### Changed
 - GUI is much faster, especially for resources with hundreds of thousands of children [#191](https://github.com/redballoonsecurity/ofrak/pull/191)
+- Resources whose data gets modified are now listed in the `resource_modified` field of component results [#200](https://github.com/redballoonsecurity/ofrak/pull/200)
```
whyitfor, line 13: I don't understand this. Does this mean that previously only resource analysis were listed in resource_modified? I think this change could be clearer

ofrak_core/ofrak/core/patch_maker/modifiers.py
```diff
@@ -209,15 +209,19 @@ async def modify(self, resource: Resource, config: SegmentInjectorModifierConfig
 
             injection_tasks.append((region.resource, BinaryInjectorModifierConfig(patches)))
 
-        all_decendants = await resource.get_descendants()
         for injected_resource, injection_config in injection_tasks:
             result = await injected_resource.run(BinaryInjectorModifier, injection_config)
-            parents_to_delete = list(
-                filter(lambda r: r.get_id() in result.resources_modified, all_decendants)
+            allowed_to_delete = result.resources_modified.difference(
```
whyitfor, line 214: I find `allowed_to_delete` a confusing variable name, but I'm struggling to come up with a better idea to express what is happening here. 

Perhaps a descriptive comment could help?

whyitfor (changes requested):
Looks good. I think a couple things could be more clearly explained (see comments).

ofrak_core/CHANGELOG.md
EdwardLarson: Previously only changes to the resource _model_ caused a resource to be add to this "modified" set. Data patches could trigger that indirectly (that is, a data patch invalidates a some attributes that resource has which depends on that data, which is a change to the model) but were not guaranteed to do so. So the change is as it says: Resources whose data gets modified are now *definitely* listed in the `resource_modified` field of component results.

ofrak_core/ofrak/component/abstract.py
EdwardLarson: > I also don't understand why the patches are not in the modified list already at this point?

See my other comment on the nature of the change. The effects of data patches were only used to invalidate attribute dependencies, which *could* result in the resource being "modified", but not *necessarily*.

whyitfor (approved):

PR #202, opened 2023-02-02
Remove music files

rbs-jacob:

Remove music files and player.

The music files were huge, and took up the majority of space in the `pip` package. Also auto-playing music is obnoxious. 

This PR removes the music files from the repo (and thereby from the built frontend, and thus from `pip`), and only opens the audio player as an easter egg, once per year. Hopefully this strikes a good balance between maintaining fun energy and not being bothersome.

@EdwardLarson


EdwardLarson (approved):
Streaming only on April fools is a good compromise

PR #203, opened 2023-02-02
updated llvm to include subsections and rel sections

SamL98:
see title

### DIFF
--- a/ofrak_patch_maker/ofrak_patch_maker/toolchain/llvm_12.py
+++ b/ofrak_patch_maker/ofrak_patch_maker/toolchain/llvm_12.py
@@ -54,7 +54,7 @@ class LLVM_12_0_1_Toolchain(Toolchain):
         )
 
         if self._config.separate_data_sections:
-            raise NotImplementedError("separate sections not supported by LLVM Toolchain yet")
+            self._compiler_flags.append("-fdata-sections")
         if self._config.compiler_cpu:
             self._compiler_flags.append(f"-mcpu={self._config.compiler_cpu}")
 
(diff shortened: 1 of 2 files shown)


ofrak_patch_maker/ofrak_patch_maker/toolchain/llvm_12.py
```diff
@@ -115,6 +115,17 @@ def __init__(
     def name(self) -> str:
         return "LLVM_12_0_1"
 
+    def keep_section(self, section_name: str):
+        if section_name in self._linker_keep_list:
+            return True
+        if self._config.separate_data_sections:
+            for keep_section in self._linker_keep_list:
```
EdwardLarson, line 122: Should this be a more restricted list, limited to just relevant data section prefixes? I think it would be pretty weird if some `.text.blahblah` section is now included if the separate _data_ sections flag is enabled. This should probably only pass if the section name starts with ".rel", maybe ".data" or ".rodata"?

andresito00: Or, @EdwardLarson with the expectation that we will likely want to do something similar with .text at some point... should we make the toolchain config flag more general, i.e. something like `config.wildcard_standard_sections`?

EdwardLarson: That seems fine to me, though maybe an even more functional-sounding name. Some ideas/examples:
`keep_standard_prefixed_sections`
`keep_subsections`
`include_subsections`

But `separate_data_sections` also does cause compiler flags to be added when building object files - are these two separate behaviors that it makes sense to toggle individually? Does having separate data sections always necessarily mean you want to keep sections starting with a standard prefix? It seems like the reverse is not true.

PR #206, opened 2023-02-03
Add input to jump to offset

rbs-jacob:

Add a GUI input to jump to a given data offset.

This PR adds a bit of text above the minimap in the hex view that shows the current file offset. Clicking this text will allow the user to edit it and set it arbitrarily. Doing so and either defocusing the input or pressing "Enter" will apply the change and jump within the file. 

Note that the calculator is enabled such that calculations can also be performed in this quick jump box. For example, a user could enter something like `0x10 * 5 + 0xdeadbeef`, and it would work.

The resulting change looks like this:

image

@EdwardLarson

### DIFF
--- a/frontend/src/JumpToOffset.svelte
+++ b/frontend/src/JumpToOffset.svelte
@@ -1,0 +1,68 @@
+<style>
+  input {
+    margin-bottom: 1em;
+    max-width: 100%;
+    background: inherit;
+    color: inherit;
+    border: none;
+    text-align: center;
+  }
+
+  input:focus {
+    outline: none;
+    box-shadow: inset 0 -1px 0 var(--main-fg-color);
+  }
+</style>
+
+<script>
+  import { calculator } from "./helpers";
+  import { onMount, tick } from "svelte";
+
+  export let dataPromise, scrollY;
+  let startOffset,
+    input,
+    mounted = false;
+  const alignment = 16;
+
+  let dataLength = 0;
+  dataPromise.then((data) => {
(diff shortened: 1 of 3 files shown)


frontend/src/JumpToOffset.svelte
```diff
@@ -0,0 +1,68 @@
...
+      Math.ceil((dataLength * $scrollY.top) / alignment) * alignment,
+      0
+    );
+    input.value = `0x${startOffset.toString(16)}`;
+  }
+</script>
+
+<input
+  type="text"
+  on:focusout="{() => {
```
EdwardLarson, line 47: I don't think jumping on focusout is a good idea. This means there's no way to "cancel" jumping if you start typing then change your mind, right? I think just waiting on an Enter press is perfectly sufficient.

EdwardLarson (changes requested):

frontend/src/JumpToOffset.svelte
rbs-jacob: Well you could write something and <kbd>Ctrl</kbd>+<kbd>z</kbd>, but that you make a fair point about wanting to undo. I could make it only submit on <kbd>Enter</kbd>, but that means if you focus out and scroll a little, what you wrote will be overwritten.

rbs-jacob: While I think I'd personally prefer it to jump on focusout, I've changed it for now. If people express a different preference, we can always do it that way later.

EdwardLarson (approved):

PR #210, opened 2023-02-06
Add keyboard shortcuts to the GUI

rbs-jacob:

Add keyboard shortcuts to the GUI.

This PR adds support for keyboard shortcuts to the GUI. There is an (importable) global object for storing the shortcut keybindings and their callbacks. Any time a key is pressed outside of an input or `contentEditable` element, that object is searched and the respective callback is called.

All of the toolbar buttons with associated shortcuts show their respective shortcut in a tooltip that appears when the button is hovered.

Any component can import and update the shortcuts object. Thus callbacks can close over reactive variables local to the component. 

This design has a few nice properties:

- It's pretty simple
- Keyboard shortcuts can be created within the components that implement functionality, removing code duplication and indirection, and reducing maintenance burden
- The importable list of shortcuts is viewable and modifiable by any component, so it is possible to list all of the bound shortcuts, as well as rebind them later
- If we eventually support plugins in the form of components, they will be able to set their own keyboard shortcuts by modifying this object

For this first pass, I set a couple of shortcuts that can always be overridden later:

- <kbd>i</kbd> – identify the current resource
- <kbd>u</kbd> – unpack the current resource
- <kbd>a</kbd> – analyze the current resource
- <kbd>p</kbd> – pack the current resource
- <kbd>n</kbd> – new resource
- <kbd>Shift</kbd>+<kbd>u</kbd> – recursively unpack the current resource
- <kbd>Shift</kbd>+<kbd>p</kbd> – recursively pack the current resource
- <kbd>g</kbd> – focus on the "go to offset" input
- <kbd>h</kbd> OR <kbd>←</kbd> – collapse the current resource tree node
- <kbd>l</kbd> OR <kbd>→</kbd> – expand the current resource tree node
- <kbd>k</kbd> OR <kbd>↑</kbd> – focus one resource up in the tree
- <kbd>j</kbd> OR <kbd>↓</kbd> – focus one resource down in the tree

@whyitfor @EdwardLarson


SamL98:
The shortcuts work great but I will say that not having j and k (or up and down) sort of defeats the purpose of having them imo.

EdwardLarson:
> The shortcuts work great but I will say that not having j and k (or up and down) sort of defeats the purpose of having them imo.

@SamL98 do you mean hotkeys to navigate around, selecting different resources? I sort of agree, because with that the GUI is nearly fully usable with hotkeys so the feature is "complete". But @rbs-jacob has said this is more just adding the infrastructure for hotkeys themselves, I think adding navigation will require a bit more code besides just adding a button. I think we could do without navigation, at least initially.

rbs-jacob:
I'll add <kbd>j</kbd> and <kbd>k</kbd>, and then this will be ready to merge.

whyitfor:
@rbs-jacob this looks cool! Would be nice to have a key bindings page in the docs, no?

EdwardLarson (approved):

PR #212, opened 2023-02-06
Fix `ofrak-minimal` build, make `make -C ofrak_core develop` more robust

ANogin:

**Link to Related Issue(s)**

Make `make -C ofrak_core develop` more robust:
- Make it work when `/bin/sh` is `dash` and does not have `pushd`/`popd`
- Make it work when there is no `frontend` (such as in the `ofrak-minimal` container)

@jstrieb

### DIFF
--- a/ofrak_core/Makefile
+++ b/ofrak_core/Makefile
@@ -24,9 +24,9 @@ dependencies:
 ofrak/gui/public:
 	if [ -d /ofrak_gui ] ; then \
 		cp -r /ofrak_gui ofrak/gui/public ; \
-	else \
-		pushd ../frontend ; \
-		npm run build ; \
-		popd ; \
+	elif [ -d ../frontend ]; then \
+		cd ../frontend && \
+		npm run build && \
+		cd ../ofrak_core && \
 		cp -r ../frontend/public ofrak/gui/public ; \
 	fi


rbs-jacob (commented):
I wonder if an alternative strategy of setting `SHELL := bash` at the top would also work for all of the cases we care about? 

If you think that would be cleaner, we can try that. If not, this looks good. Your call.

ANogin:
Zero preference. The `if [ -d frontend ]` part is needed either way (unless we want to always include frontend, including in `minimal`).

rbs-jacob:
> Zero preference. The `if [ -d frontend ]` part is needed either way (unless we want to always include frontend, including in `minimal`).

I'll leave as-is for now.

I just ran some quick tests to confirm locally that importing and using OFRAK without the `ofrak_core/ofrak/gui/public` directory will work as expected unless the user tries to start the GUI (such as when the `--gui` flag is passed to the `ofrak` CLI). As such, it should be fine for a minimal container to not have the front end.

That being said, the only additional dependency from the `frontend` folder is a few MB of static web assets (since we removed the music a few PRs back).

ANogin:
Well, `frontend` also add a bunch of steps to the container build process, so probably some value to leaving it out...

rbs-jacob (approved):

PR #215, opened 2023-02-08
add view and backend support to add tags to resources through the GUI

EdwardLarson:
Add a button to the GUI to add a new tag to a resource.

**Link to Related Issue(s)**
There was no way to manually tag a resource, which means an RE will dead-end when trying to use the GUI as soon as something can't be automatically identified.

* `ResourceTag` tracks all the tags that are created
* An endpoint for the server that gets and returns all the tags
* An endpoint for the server to add a tag to a resource
* A new button and "modifier view" in GUI with interface to select a tag (from a drop-down) and add it to the resource

@rbs-jacob


frontend/src/AddTagView.svelte
```diff
@@ -0,0 +1,211 @@
...
+  function chooseTag() {
+    if (selectedTag) {
+      modifierView = undefined;
+      $selectedResource.add_tag(selectedTag);
+    }
+  }
+
+  onMount(async () => {
+    await fetch(`/get_all_tags`).then((r) => {
+      r.json().then((ofrakTags) => {
```
rbs-jacob, line 153: This will not properly handle non-200 status code responses from the server. For an example of the error handling to add here, see: https://github.com/redballoonsecurity/ofrak/blob/b68c632ae0f5b9d936ecef4bbcff600ddbcf8d88/frontend/src/ofrak/remote_resource.js#L169-L176

Moreover, it should be inside a `try {} catch {}`, and update the `errorMessage` variable in the `catch` block. See: https://github.com/redballoonsecurity/ofrak/blob/b68c632ae0f5b9d936ecef4bbcff600ddbcf8d88/frontend/src/CarveView.svelte#L105-L123

frontend/src/AddTagView.svelte
```diff
@@ -0,0 +1,211 @@
...
+</script>
+
+<div class="container">
+  <div class="inputs">
+    <p>Select tag to add to resource.</p>
+    New tag:
+    {#await ofrakTagsPromise}
+      <LoadingText />
+    {:then ofrakTags}
+      {#if ofrakTagsPromise && ofrakTagsPromise.length > 0}
```
rbs-jacob, line 174:
```suggestion
      {#if ofrakTags && ofrakTags.length > 0}
```

frontend/src/ResourceTreeToolbar.svelte
```diff
@@ -89,6 +90,14 @@
         },
       },
 
+      {
+        text: "Add tag",
```
rbs-jacob, line 94:
```suggestion
        text: "Add Tag",
```

ofrak_core/ofrak/gui/server.py
```diff
@@ -456,6 +459,23 @@ async def search_for_vaddr(self, request: Request) -> Response:
         except NotFoundError:
             return json_response([])
 
+    @exceptions_to_http(SerializedError)
+    async def add_tag(self, request: Request) -> Response:
+        """
+        Expected POST body is a comment in the form Tuple[Optional[Range], str] (serialized to JSON).
+        """
```
rbs-jacob, line 466:
```suggestion
```

Does this docstring actually apply, or is it fake news?

frontend/src/AddTagView.svelte
```diff
@@ -0,0 +1,211 @@
...
+          </select>
+
+          <button
+            on:click|stopPropagation="{() => undefined}"
+            disabled="{!selectedTag}"
+            type="submit">Add</button
+          >
+        </form>
+      {:else}
+        Failed to get the list of OFRAK tags!
```
rbs-jacob, line 195: I think the following is more descriptive for this path.

```suggestion
        No tags found!
```

rbs-jacob (changes requested):

ofrak_core/ofrak/gui/server.py
EdwardLarson: ☝️

frontend/src/AddTagView.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+  .nowrap {
+    white-space: nowrap;
+  }
+
+  .error {
+    margin-top: 2em;
+  }
+
+  h1 {
```
rbs-jacob, line 68: There are a bunch of unused CSS selectors that cause warnings during the `npm run build`, and that should be removed.

```
(!) Plugin svelte: Unused CSS selector "label"
src/AddTagView.svelte
47:   }
48:
49:   label {
      ^
50:     margin-bottom: 1em;
51:     display: flex;
(!) Plugin svelte: Unused CSS selector ".nowrap"
src/AddTagView.svelte
58:   }
59:
60:   .nowrap {
      ^
61:     white-space: nowrap;
62:   }
(!) Plugin svelte: Unused CSS selector "h1"
src/AddTagView.svelte
66:   }
67:
68:   h1 {
      ^
69:     text-transform: uppercase;
70:     font-weight: bold;
(!) Plugin svelte: Unused CSS selector ".center"
src/AddTagView.svelte
81:   }
82:
83:   .center {
      ^
84:     display: flex;
85:     flex-direction: column;
(!) Plugin svelte: Unused CSS selector "input[type="file"]"
src/AddTagView.svelte
114:   }
115:
116:   input[type="file"] {
       ^
117:     display: none;
118:   }
created public/build/bundle.js in 7.2s
```

frontend/src/AddTagView.svelte
```diff
@@ -0,0 +1,222 @@
...
+<div class="container">
+  <div class="inputs">
+    <p>Select tag to add to resource.</p>
+    New tag:
+    {#await ofrakTagsPromise}
+      <LoadingText />
+    {:then ofrakTags}
+      {#if ofrakTags && ofrakTags.length > 0}
+        <form on:submit|preventDefault="{chooseTag}">
+          <select
```
rbs-jacob, line 187: I think this would look cleaner if the `<form>` was a row flexbox with the `<select>` having `flex-grow: 1`. That way the label, dropdown, and button would all be in a row, and would take up the full width of the containing element. When doing this, removing the `max-width: 50%` from the `<form>` will help as well.

```suggestion
    {#await ofrakTagsPromise}
      <LoadingText />
    {:then ofrakTags}
      {#if ofrakTags && ofrakTags.length > 0}
        <form on:submit|preventDefault="{chooseTag}">
          New tag:
          <select
```

If you'd rather, I can just make those changes and push them to this branch.

rbs-jacob (changes requested):

frontend/src/AddTagView.svelte
EdwardLarson: I tweaked it slightly so the "New Tag" text is inline with that and there's no more max width. But I don't understand the flex grow (can't even add it anyway as the parent's `display: block` style is apparently incompatible) and it looks fine.

rbs-jacob: I just pushed a change that addresses the remainder of this comment.

rbs-jacob (approved):

PR #217, opened 2023-02-09
re-pin xattr version to version that works with Python 3.9

EdwardLarson:
re-pin xattr version to version that works with Python 3.9

**Link to Related Issue(s)**
The xattr version we had pinned it to used a filter for classifiers in its setup.py, and Python 3.9 needs to that be a list, so installing that version of xattr raises an error in Python 3.9.

Re-pin to latest version of xattr

@whyitfor


whyitfor (approved):
LGTM

PR #218, opened 2023-02-09
Add `--check_python_reqs` option to build_image.py

ANogin:

**Link to Related Issue(s)**

### DIFF
--- a/build_image.py
+++ b/build_image.py
@@ -31,6 +31,7 @@ class OfrakImageConfig:
     build_finish: bool
     # Whether to supply --no-cache to docker build commands
     no_cache: bool
+    check_python_reqs: bool
     extra_build_args: Optional[List[str]]
     install_target: InstallTarget
     cache_from: List[str]
@@ -140,6 +141,7 @@ def parse_args() -> OfrakImageConfig:
     parser.add_argument("--base", action="store_true")
     parser.add_argument("--finish", action="store_true")
     parser.add_argument("--no-cache", action="store_true")
+    parser.add_argument("--check-python-reqs", action="store_true")
     parser.add_argument(
         "--target",
         choices=[InstallTarget.DEVELOP.value, InstallTarget.INSTALL.value],
@@ -157,6 +159,7 @@ def parse_args() -> OfrakImageConfig:
         args.base,
         args.finish,
         args.no_cache,
+        args.check_python_reqs,
         config_dict.get("extra_build_args"),
         InstallTarget(args.target),
         args.cache_from,
@@ -233,6 +236,7 @@ def create_dockerfile_base(config: OfrakImageConfig) -> str:
 def create_dockerfile_finish(config: OfrakImageConfig) -> str:
     full_base_image_name = "/".join((config.registry, config.base_image_name))
     dockerfile_finish_parts = [
+        "# syntax = docker/dockerfile:1.3\n\n",
         f"FROM {full_base_image_name}:{GIT_COMMIT_HASH}\n\n",
         f"ARG OFRAK_SRC_DIR=/\n",
     ]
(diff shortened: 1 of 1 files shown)


ANogin:
@whyitfor could we please land this one and use the new option in our CI? As I am working on #416 and #417 I am went to go back to this manually to make sure all the dependencies across ofrak components are consistent (only remembered to do it because I saw the docker build output for #417 and noticed it's uninstalling/reinstalling some packages in weird way in `make develop`) - and now discovering that even the main branch has incorrectly specified dependencies (will fill in separate PRs on those)?

whyitfor:
@ANogin, can you complete the PR descripion for this?

whyitfor (changes requested):
I've reflected on this one a bit, and discussed with @rbs-jacob.

Let's make the following changes.

1. Remove the `check-python-reqs` argument from `build-image.py`
2. `build-image.py` should add an `inspect` target to the Makefile it generates that runs `python3 -m pip check`. This will enable users who want to check for this consistency to do so.
3. Does `--network="none"` work for all use cases? If so, let's make it default. If not, we should remove it.

ANogin:
> I've reflected on this one a bit, and discussed with @rbs-jacob.
> 
> Let's make the following changes.
> 
> 1. Remove the `check-python-reqs` argument from `build-image.py`
> 2. `build-image.py` should add an `inspect` target to the Makefile it generates that runs `python3 -m pip check`. This will enable users who want to check for this consistency to do so.
> 3. Does `--network="none"` work for all use cases? If so, let's make it default. If not, we should remove it.

Done, should be good to go once #419 is fixed.

build_image.py
```diff
@@ -145,6 +145,7 @@ def parse_args() -> OfrakImageConfig:
     parser.add_argument("--base", action="store_true")
     parser.add_argument("--finish", action="store_true")
     parser.add_argument("--no-cache", action="store_true")
+    parser.add_argument("--check-python-reqs", action="store_true")
```
whyitfor, line 148:
```suggestion
```

disassemblers/ofrak_angr/Makefile
```diff
@@ -7,7 +7,9 @@ install:
 
 .PHONY: develop
 develop:
-	$(PIP) install -e .[test] --config-settings editable_mode=compat
+	# Legacy-edinable is needed to allow mypy to find packages.
+	# See https://github.com/python/mypy/issues/13392
+	SETUPTOOLS_ENABLE_FEATURES="legacy-editable" $(PIP) install -e .[test]
```
whyitfor, line 12: I don't understand why this change is needed.

#561 updated these targets to enable `editable_mode=compat` -- is this not the same issue?

cc @alchzh

alchzh: yes this change is no longer needed (the command line flag is preferred over the environment variable)

ANogin: No, with command line flag it insists on reaching out to pypi over network, but with the environment variable it does not.

whyitfor: I've looked into this a bit and reflected.

If I understand correctly,`SETUPTOOLS_ENABLE_FEATURES` results in using a pre-PEP 517 backend. `editable_mode=compat` uses a PEP660 backend that provides some backwards compatibility. Switching back to `SETUPTOOLS_ENABLE_FEATURES` does not make sense to me: is moving the repo in the wrong direction, away from PEP660.

It looks like there are ways to get `pip` to not reach out to the network (a combination of `--no-index` and `--find-links`) which we might want to eventually look into.

For now though, I'd like to prioritize getting the `pip check` functionality this MR adds in, without the `--network="none"` changes.

PR #219, opened 2023-02-09
swap order that errors are raised when auto-running fails

EdwardLarson:
the component's error is now at the bottom of the stack trace, so the user sees it first

Tweak how errors are raised when auto-running components, so the actual root cause is not buried

**Link to Related Issue(s)**
If a component is auto-run (which is common, every `analyze` or `unpack` are auto-runs for example) and it raises an error `E`, then the job service also generates it's own error `F` with info about why the component was chosen by the auto-run. Previously the errors were raised as `raise F from E`, which resulted in the stack trace printing error info in the order:

```
E
F
```
So what a user sees most obviously is some junk about the auto-run, and the root cause is buried halfway up. This also translates to the GUI when there is a backend error: Since `F` is the most proximate error, that auto-run info error is the one shown in a pop-up when there is a backend error.

Do `raise E from F` instead, so errors are printed in the order:
```
F
E
```
And the most recent stuff printed is about the root cause that the user cares about. The info about why the component was selected is still there, if it will be useful for debugging.

### DIFF
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -4,14 +4,16 @@ All notable changes to `ofrak` will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
+### Fixed
+- Fix bug where initially loaded GUI resource has collapsed children
 
 ### Added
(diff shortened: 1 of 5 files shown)


rbs-jacob (approved):

EdwardLarson:
As an example of the effect in the GUI, here's how an error is displayed before and after this PR:
Screen Shot 2023-02-09 at 1 37 14 PM
Screen Shot 2023-02-09 at 1 45 22 PM

rbs-jacob:
I like that this is a fairly simple fix that will have a major usability impact! Both in the GUI and for API users working at the command line.

PR #220, opened 2023-02-10
Sort by data {length, offset}

EdwardLarson:
Add mechanism to sort & filter by data length or offset like other indexable attributes (e.g. VirtualAddress)

**Link to Related Issue(s)**
There wasn't a way to format a query (like `get_descendants`) so that OFRAK would filter or sort by aspects of the resource's binary data. Users would have to grab that info themselves and implement some filtering or sorting logic locally.

* Add `RootOffset` and `Length` indexable attributes that can be used to sort of filter by these aspects of resource's binary data:
`r_sort=ResourceSort(RootOffset),`
* Have the GUI sort resources by data offset by default
* Implement this with a semi-magical special ResourceAttributes known as `DataAttributes` which OFRAK specially handles updating when resources are created with data or get patched

### DIFF
--- a/ofrak_core/ofrak/model/resource_model.py
+++ b/ofrak_core/ofrak/model/resource_model.py
@@ -897,7 +897,7 @@ def _validate_indexed_type(getter_func: Callable[[Any], X]):
 
 
 @dataclasses.dataclass(**ResourceAttributes.DATACLASS_PARAMS)
-class DataBytes(ResourceAttributes):
+class Data(ResourceAttributes):
     """
     Special attributes class for accessing info about a resource's binary data.
     Users should never access or modify this directly! Changing the fields of this data structure
--- a/ofrak_core/ofrak/model/resource_model.py
+++ b/ofrak_core/ofrak/model/resource_model.py
@@ -897,23 +897,19 @@ def _validate_indexed_type(getter_func: Callable[[Any], X]):
 
 
 @dataclasses.dataclass(**ResourceAttributes.DATACLASS_PARAMS)
-class _DataAttributes(ResourceAttributes):
+class DataBytes(ResourceAttributes):
     """
     Special attributes class for accessing info about a resource's binary data.
     SHOULD NEVER BE USED OUTSIDE OF INTERNAL OFRAK MECHANISMS.
     """
 
-    root_offset: int
-    length: int
+    _offset: int
+    _length: int
 
     @index
-    def RootOffset(self) -> int:
-        return self.root_offset
+    def Offset(self) -> int:
+        return self._offset
 
     @index
     def Length(self) -> int:
-        return self.length
-
-
-RootOffset = _DataAttributes.RootOffset
-Length = _DataAttributes.Length
+        return self._length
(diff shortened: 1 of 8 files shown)


EdwardLarson:
Couple things I want to polish off before I mark this is ready:

[ ] Are the names `RootOffset` and `Length` good?
[ ] Are things well-organized?
[ ] Does this have a noticeable performance impact on heavier workloads? It adds one extra call to `DataService.get_by_ids` and a loop, so I'm not concerned, just want to double check.

ofrak_core/ofrak/model/resource_model.py
```diff
@@ -894,3 +894,22 @@ def _validate_indexed_type(getter_func: Callable[[Any], X]):
...
+
+@dataclasses.dataclass(**ResourceAttributes.DATACLASS_PARAMS)
+class DataBytes(ResourceAttributes):
+    """
+    Special attributes class for accessing info about a resource's binary data.
+    SHOULD NEVER BE USED OUTSIDE OF INTERNAL OFRAK MECHANISMS.
+    """
+
+    _offset: int
+    _length: int
```
whyitfor, line 907: These attributes are going to be useful in the GUI. With that in in mind, I think there are three things I would like to know:

1. Is the data mapped?
2. What is the offset in the root?
3. What is the size of the data?

I'm wondering if this attribute could contain all three pieces of information. Something like:

```
@dataclass
class DataBytes

    offset_in_root: Optional[int]
    length: int
```

What do you think?

EdwardLarson: It's an interesting idea. What are your thoughts on how extensively we should rely on/use these sort-of-magic attributes? I think the more visible we make this class, the more protections we need against manual (or accidental) modification of these attributes.

whyitfor: FWIW this will show up in the GUI.

I think we make the dataclass frozen and update the docstring to say explicitly that setting these fields will not update the underlying data can go a long way.

EdwardLarson: The dataclass was already frozen, and I elaborated in the docstring as you suggested.

Per our quick offline chat about this, showing this in the GUI is helpful since it's the only way to see data offset and length, so I won't add code to filter it out.

ofrak_core/ofrak/service/dependency_handler.py
```diff
@@ -65,6 +66,19 @@ async def handle_post_patch_dependencies(self, patch_results: List[DataPatchesRe
...
         )
 
+        all_data_ids = [patch_result.data_id for patch_result in patch_results]
+        models_by_data_id: Dict[bytes, DataModel] = {
+            data_id: model
+            for data_id, model in zip(
+                all_data_ids, await self._data_service.get_by_ids(all_data_ids)
+            )
+        }
+        # First go through and update all models' Data
```
rbs-jacob, line 76:
```suggestion
        # Go through and update all models' Data
```

"First" sort of implies that there will be a "next" in a subsequent comment...

rbs-jacob (changes requested):
Besides the one minor comment, should the `data_index.py` file still be there? It is empty.

Pending those two changes, this looks good to me. Really excited to get this merged into the GUI! It will be very useful for making the tree clearer.

rbs-jacob (approved):

PR #234, opened 2023-02-22
Add PPC GNU 10 Linux toolchain (fix #204)

### LINKED ISSUE #204: PowerPC PatchMaker Support
**What is the use case for the feature?**
Patching PPC binaries with PatchMaker requires support for PPC builds. The LLVM toolchain may be sufficient, but we should also have a GCC toolchain targeting PPC. For example, I'm not sure how good LLVM's support for the [VLE](https://www.nxp.com/docs/en/supporting-information/VLEPIM.pdf) encoding is.

**Does the feature contain any proprietary information about another company's intellectual property?**
Nope.

**How would you implement this feature?**
A Toolchain & tests that targets PPC. The same test could also test the PPC target for LLVM.

Just relying on LLVM is probably fine, but GCC toolchains give more flexibility.

Not as the main contributor, but I'll definitely help out with blockers!

rbs-jacob:

Add PPC GNU 10 Linux toolchain support to OFRAK PatchMaker. Addresses #204.

@andresito00

### DIFF
--- a/ofrak_patch_maker/Dockerstub
+++ b/ofrak_patch_maker/Dockerstub
@@ -70,3 +70,8 @@ fi;
 
 #AVR GCC
 RUN apt-get update && apt-get install -y gcc-avr binutils-avr avr-libc
+
+#PPC GNU 10 Linux
+RUN if [ "$TARGETARCH" = "amd64" ]; then \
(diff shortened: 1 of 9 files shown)


andresito00:
As discussed in person. We will be creating an issue for implementation of the asm test :) Nothing to see here!

PR #245, opened 2023-03-06
Adding necessary compiler flags for non-relocatable arm64 patches

rbs-alexr:
Adding the -fno-pic compiler flag to the GNU Toolchain allows non-relocatable arm64 patches to succeed

**Link to Related Issue(s)**
N/A

This pull request adds the -fno-pic compiler flag to the GNU Toolchain

@whyitfor

### DIFF
--- a/ofrak_patch_maker/CHANGELOG.md
+++ b/ofrak_patch_maker/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to `ofrak-patch-maker` will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
+### Added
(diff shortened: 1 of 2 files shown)


andresito00:
@alexreuter Let's provide some more detailed context here.

Do you mind providing more details on why not having this flag resulted in patch creation failures?

rbs-alexr:
Without the -fno-pic compiler flag, the compiler tries to generate a GOT to access "all imported or exported read-write data". If we are making a patch to an ELF where having a GOT is not an option, for example where the assembly is ripped out of the ELF to be used on a system with no dynamic linker, then we want relocatable=False to actually not use a GOT

whyitfor (approved):

andresito00 (approved):

andresito00:
It should be noted that this change will apply to all GNU toolchains, not just ARM64. I believe we want it in all of them. 

I think we need to explore a rename of this configuration flag to make it clear that we are preventing or enabling certain types of relocations. We may find ISA-specific constraints as this type of configuration sees more usage.

whyitfor:
@rbs-alexr, can you update the PatchMaker changelog to document these changes? Once that's done I think we are good to go.

PR #249, opened 2023-03-09
Slurp up contents of patch source files, rather than expected them as resources

EdwardLarson:
Remove need to create Resources to pass source code and headers to `PatchFromSourceModifier` and `FunctionReplaceModifier`

**Link to Related Issue(s)**
Creating a resource was awkward and resulted in unnecessary boilerplate.

Add recursive slurping and dataclass post-init so that a user can just pass the path to the source directory and OFRAK will read out the contents of the files behind the scenes. This keeps a very simple interface for the benefit of the user, and does not assume the filesystem is the same between the caller and a component.


ofrak_core/ofrak/core/patch_maker/modifiers.py
```diff
@@ -53,41 +101,23 @@ class PatchFromSourceModifier(Modifier):
...
         header_dirs = []
-        if config.header_directory_resource_ids:
-            for header_directory_id in config.header_directory_resource_ids:
-                header_dirs.append(
-                    await self._flush_source_bundle_to_disk(header_directory_id, resource)
-                )
+        for header_directory in config.header_directories:
+            header_tmp_dir = tempfile.mkdtemp()
+            header_directory.dump(header_tmp_dir)
+            header_dirs.append(header_tmp_dir)
```
rbs-jacob, line 120: Unless I'm misunderstanding something, the call to `tempfile.mkdtemp` leaks temporary directories, right? The [docs for `mkdtemp`](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) specify that the caller is responsible for cleaning up, but I don't see logic to do that anywhere.

ofrak_core/ofrak/core/patch_maker/modifiers.py
```diff
@@ -3,47 +3,95 @@
...
     defining where to inject one or more of the .text, .data, and .rodata from the build file
     :var toolchain_config: configuration for the
     [Toolchain][ofrak_patch_maker.toolchain.abstract.Toolchain] to use
-    :var toolchain: the type of which [Toolchain][ofrak_patch_maker.toolchain.abstract.Toolchain] to use
+    :var toolchain: the type of which [Toolchain][ofrak_patch_maker.toolchain.abstract.Toolchain]
+      to use to build patch
     :var patch_name: Optional name of patch
-    :var header_directory_resource_ids: Optional additional FilesystemRoot resources with header
-    directories
+    :var header_directories: (Optional) paths to directories to search for header files in
```
rbs-jacob, line 86: Should the order here be swapped to match the order below?

rbs-jacob (changes requested):
One and a half minor comments, and this is ready to go.

ofrak_core/ofrak/core/patch_maker/modifiers.py
EdwardLarson: I don't think we do want to clean it up by deleting it. I often find it useful for debugging to re-run the commands that PatchMaker invoked, which operate on files in those temp directories. Even when the component succeeds, it's useful to still have the files there if I want to inspect them or try modifying the commands for a similar result. 

Over time it leaves old folders in /tmp, but that's kind of what /tmp is for.

rbs-jacob (approved):

PR #243, opened 2023-03-06
Fix bug in GUI where "jump to offset" 

whyitfor:
**One sentence summary of this PR (This should go in the CHANGELOG!)**
The GUI "jump to offset" feature would round up to nearest 16-byte-aligned address. This fixes this behavior to correctly round down.

**Please describe the changes in your request.**
See above.

PR #254, opened 2023-03-15
Fix bug when jumping to a multiple of 0x10

rbs-jacob:

Fix bug where jumping to a multiple of `0x10` in the GUI went to the previous line.

The changes introduced in #243 caused a different bug where jumping to an exact multiple of `0x10` rounded down the address to jump to and went to the previous line in the hex view of the GUI. This change fixes that issue.

@whyitfor

### DIFF
--- a/frontend/src/JumpToOffset.svelte
+++ b/frontend/src/JumpToOffset.svelte
@@ -55,7 +55,7 @@
     if (e.key === 'Enter') {
       input.blur();
       try {
-        let result = calculator.calculate(input.value);
+        let result = calculator.calculate(input.value) + 1;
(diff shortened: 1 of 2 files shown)


whyitfor:
The description suggests there is a rounding error. Is that really the case?

What seems to be happening is if, say, you try to navigate to 0x20 -- the GUI is displaying **byte** 0x20 (which is on the previous line), and not the 0x20 heading.

If this is what is happening, I wonder if it makes sense to have the GUI start at byte 0 (instead of 1). If not, this change looks good.

whyitfor (approved):

PR #258, opened 2023-03-18
General support for separate_data_sections, and new flag for keeping more sections

EdwardLarson:
- Add separate data sections support to LLVM toolchain, and add general flag for including subsections

**Link to Related Issue(s)**
Extends and replaces https://github.com/redballoonsecurity/ofrak/pull/203

The addition of the `separate_data_sections` functionality to the LLVM toolchain aligns it with the GNU toolchain. In both toolchains, enabling `separate_data_sections` requires broadening the pattern for the sections we keep from the BOM to the FEM. Per the discussion in #203, I added a separate flag that just enables that broader keep list without actually making data sections.

### DIFF
--- a/ofrak_patch_maker/ofrak_patch_maker/toolchain/abstract.py
+++ b/ofrak_patch_maker/ofrak_patch_maker/toolchain/abstract.py
@@ -82,7 +82,7 @@ class Toolchain(ABC):
         # The keep_list should only contain FUNCTIONALLY important sections
         # (not empty .got.plt, for instance).
         # TODO: Come up with a better system to handle this...
-        self._linker_keep_list = [".data", ".rodata", ".text"]
+        self._linker_keep_list = [".data", ".rodata", ".text", ".rel"]
         self._linker_discard_list = [
             ".gnu.hash",
             ".comment",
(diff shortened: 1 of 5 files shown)


ofrak_patch_maker/ofrak_patch_maker/toolchain/abstract.py
```diff
@@ -357,11 +357,16 @@ def link(self, o_files: List[str], exec_path: str, script: str = None):
     def linker_include_filter(symbol_name: str) -> bool:
         return "." in symbol_name or "_DYNAMIC" in symbol_name
 
-    def keep_section(self, section_name: str) -> bool:
-        if self._config.separate_data_sections:
-            raise NotImplementedError("you must override keep_section() in your Toolchain sublass")
```
rbs-jacob, line 362: Out of curiosity, do you know why this was done this way in the first place?

rbs-jacob (commented):
Depending on the discussion below, this is probably good to go.

ofrak_patch_maker/ofrak_patch_maker/toolchain/abstract.py
EdwardLarson: I believe it is because flexible keep sections was only allowed because of the `separate_data_sections` option, which would have to be implemented in a toolchain-dependent way. Each toolchain could use a different flag and generate its own pattern of data section names. But it turns out the pattern is pretty standard for the two major families we have (LLVM and GNU). Overriding this is still possible if a toolchain breaks the pattern.

rbs-jacob (approved):

PR #264, opened 2023-03-22
Better view analysis

EdwardLarson:
Choose Analyzer components which output the entirety of a view, rather than piece by piece, which would choose the wrong Analyzer sometimes.

**Link to Related Issue(s)**
ResourceViews are composed of multiple attributes types, and when running something like `view_as` OFRAK would previously make a request for each of those attributes types. With limited visibility into what the user actually wants, the job service would sometimes end up choosing Analyzers which do output one or more of the requested attributes but actually require the analysis of the other attributes, resulting in a deadlock.

* Auto-analyze attributes requests now can take multiple attributes types at once (and `view_as` uses this)
* The component filter for analyzer output looks for analyzers which output all of the requested attributes (and possibly other attributes as well), instead of any analyzers which output one or more of the requested attributes
* When auto-analyzing attributes, the job service first looks for an Analyzer outputting all of the requested attributes, and if such an Analyzer is not found, falls back to looking for Analyzers for each requested attribute individually (the previous behavior)

@rbs-jacob

### DIFF
--- a/ofrak_core/ofrak/resource.py
+++ b/ofrak_core/ofrak/resource.py
@@ -421,7 +421,7 @@ class Resource:
         """
         attributes = self._check_attributes(resource_attributes)
         if attributes is None:
-            await self._analyze_attributes(resource_attributes)
+            await self._analyze_attributes((resource_attributes,))
             return self.get_attributes(resource_attributes)
         else:
             return attributes
@@ -533,13 +533,13 @@ class Resource:
 
         destination.write(await self.get_data())
 
-    async def _analyze_attributes(self, attribute_type: Type[ResourceAttributes]):
+    async def _analyze_attributes(self, attribute_types: Tuple[Type[ResourceAttributes], ...]):
         job_context = self._job_context
         components_result = await self._job_service.run_analyzer_by_attribute(
             JobAnalyzerRequest(
                 self._job_id,
                 self._resource.id,
-                attribute_type,
+                attribute_types,
                 tuple(self._resource.tags),
             ),
             job_context,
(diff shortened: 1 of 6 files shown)


rbs-jacob:
I'm a little confused about whether this reliably solves the problem based on this last bullet:

> - When auto-analyzing attributes, the job service first looks for an Analyzer outputting all of the requested attributes, and if such an Analyzer is not found, falls back to looking for Analyzers for each requested attribute individually (the previous behavior)

This makes it sound like there is still a risk of deadlock in certain circumstances?

EdwardLarson:
> This makes it sound like there is still a risk of deadlock in certain circumstances?

That's correct. The true solution to the root cause would be pretty complicated. The approach I have in mind is to flesh out the job context to include a "stack" of the components or at least analyzers being run, and raise an error when the same analyzer appears twice in the stack. However that would take me a couple days at *least* to do, this took me an hour or so.

And besides this does solve a real problem: The wrong analyzer was being chosen in a test that was deadlocking. The job service was choosing the `ComplexBlockSymbolAnalyzer` for one of the attributes in a `ComplexBlock` view, and using the `ComplexBlockAnalyzer` for other attributes. Since the former indirectly calls the latter, the set of analysis tasks was deadlocking.

rbs-jacob:
Thanks for the clarification!

ofrak_core/ofrak/resource.py
```diff
@@ -747,15 +747,17 @@ def _view_as(self, viewable_tag: Type[RV]) -> Union[RV, Awaitable[RV]]:
...
+                attrs_t
+                for attrs_t, existing in zip(composed_attrs_types, existing_attributes)
+                if not existing
+            )
+        )
 
         # Only if analysis is absolutely necessary is an awaitable created and returned
         async def finish_view_creation() -> RV:
-            await asyncio.gather(*analysis_tasks)
+            await asyncio.gather(analysis_task)
```
rbs-jacob, line 760: If there is just a single task, does it make sense to `asyncio.gather` it instead of just `await`ing it?

```suggestion
            await analysis_task
```

rbs-jacob (changes requested):
Two very small changes, and this should be good to go!

PR #265, opened 2023-03-22
Generate script in GUI based on user actions

marczalik:
While using the GUI, dynamically generate runnable OFRAK scripts based on corresponding user actions and allow user to view and download the current script.

**Link to Related Issue(s)**
N/A

Add `ScriptBuilder` class to handle generation of scripts. 
Add server endpoint for getting current script. 
Add `ScriptView` component to frontend to display the script.

@whyitfor  @Edward-Larson 

**Suggestions for Post-MVP Improvements**

- Determine at run-time which backend a user is running and modify the script to import that specific backend.
- Create a test that runs the generated script and compares it to expected results.
- Add action to flush to disk in script when a user downloads the script from the GUI.
- Improve the way ScriptBuilder handles multi-line action strings; in particular make it less brittle to small changes and remove the concern about whitespace from the route handlers.
- Reconsider naming of "Show Script" button given that not all users are concerned with or aware that a script is being generated in the background at all times.
- Consider moving many of `ScriptBuilder`'s methods into `ScriptSession`.
- Normalize the scripts in the test cases by comparing them as strings (not lists) and using regex to replace inconsistent or changeable values.


ofrak_core/ofrak/service/script_builder.py
rbs-jacob, line 1: @whyitfor and @EdwardLarson may have thoughts on this, but this should probably *not* be a service/in the `service` folder, in my opinion.

ofrak_core/ofrak/service/script_builder.py
```diff
@@ -0,0 +1,268 @@
...
+
+class ActionType(IntEnum):
+    UNPACK = 0
+    MOD = 1
+    PACK = 2
+    UNDEF = 3
+
+
+@dataclass
+class ScriptAction:
```
rbs-jacob, line 30: Would appreciate some doc comments for this and the classes below

frontend/public/code.css
```diff
@@ -0,0 +1,87 @@
...
+
+code.hljs {
+  padding: 3px 5px;
+}
+
+.hljs {
+  color: #e0e2e4;
+}
+
+.hljs-keyword,
```
rbs-jacob, line 15: We could probably stand to break a few of these up, and add some more colors. It's worth doing some combination of:

- [ ] Opening an issue to document that this should be done – we shouldn't block this PR for it
- [ ] Generating some themes ([possibly with HueMint](https://huemint.com/illustration-3/))
- [ ] Syncing with Neil about this if he has time to put together a palette

frontend/src/App.svelte
```diff
@@ -61,7 +62,11 @@
     rootResourceLoadPromise = new Promise((resolve) => {}),
     resourceNodeDataMap = {},
     resources = {};
-  let carouselSelection, currentResource, rootResource, modifierView;
+  let carouselSelection,
+    currentResource,
+    rootResource,
+    modifierView,
+    scriptView;
```
rbs-jacob, line 69: `modifierView` ended up becoming a bit of a misnomer. I don't want the same fate for `scriptView` that will result in a refactor later. Probably worth renaming to something like `bottomLeftPane` or something. Someone will refactor the `modifierView` name into something clearer at a later date.

Moreover, it's confusing to have a variable named `scriptView` and a class named `ScriptView` – seems ripe for unexpected errors.

frontend/src/ResourceTreeToolbar.svelte
```diff
@@ -212,6 +213,14 @@
           modifierView = SearchView;
         },
       },
+
+      {
+        text: "Show Script",
+        iconUrl: "/icons/identify.svg",
```
rbs-jacob, line 219: Search view directly above this uses this icon. Identify also uses the same icon. It's worth modifying one of the existing ones. Perhaps the "new" icon, but change the plus to some horizontal lines?

If there isn't time for that, I'd at least use a different icon from this one.

frontend/src/ScriptView.svelte
```diff
@@ -0,0 +1,125 @@
...
+  export let scriptView;
+
+  onMount(async () => {
+    await $selectedResource.get_script();
+  });
+</script>
+
+<link rel="stylesheet" href="./code.css" />
+
+<div class="close">
```
rbs-jacob, line 87: Dan mentioned renaming this to `buttonbar` or some other class name. `close` has become a misnomer with the addition of the download button.

frontend/src/ScriptView.svelte
```diff
@@ -0,0 +1,125 @@
...
+  <button
+    class="download"
+    on:click="{async (e) => {
+    const lines = $script.join("\n");
```
rbs-jacob, line 91: Did you remember to run `make lint`???

ofrak_core/ofrak/gui/server.py
```diff
@@ -166,6 +168,27 @@ def __init__(
...
+                        allow_headers="*",
+                    )
+                },
+            )
+
+            # Configure CORS on all routes.
+            for route in list(self._app.router.routes()):
+                cors.add(route)
+        except ImportError:
+            pass
```
rbs-jacob, line 191: This `try/except` block should probably be inside a conditional/flag. Wrapping all the routes with CORS is great for local debugging, but runs the risk of causing problems when the server is subclassed and routes are added. 

Also, if we're going to rely on `aiohttp_cors`, we need to make sure to add the dependency to the list.

ofrak_core/ofrak/gui/server.py
```diff
@@ -258,6 +288,8 @@ async def batch_get_range(self, request: Request) -> Response:
         else:
             raise ValueError("No IP address found for the remote request!")
 
+        # TODO: replace request.json()
```
rbs-jacob, line 291: Still to do?

ofrak_core/ofrak/gui/server.py
```diff
@@ -410,39 +457,66 @@ async def queue_patch(self, request: Request) -> Response:
         start = int(start_param) if start_param is not None else 0
         end_param = request.query.get("end")
         end = int(end_param) if end_param is not None else (await resource.get_data_length())
-
+        # TODO: There has to be a better way
+        new_data_string = "/x" + "/x".join(
```
rbs-jacob, line 461: Maybe I'm misunderstanding, but should this be a backslash?

```suggestion
        new_data_string = "\\x" + "\\x".join(
```

rbs-jacob (changes requested):
I just did a very quick pass, and will do another, slightly more comprehensive review tomorrow. In the meantime, I've requested a few changes.

ofrak_core/ofrak/service/script_builder.py
marczalik: Any suggestions on where to put it? Is it a part of `core`?

ofrak_core/ofrak/gui/server.py
marczalik: Added the dependency to `requirements.txt`.

What would the use of `aiohttp_cors` be conditional upon? Check whether the type of self is `AiohhtpOFRAKServer`?

ofrak_core/ofrak/service/script_builder.py
rbs-jacob: I think `gui`. When will the script builder be used outside of the GUI?

ofrak_core/ofrak/gui/server.py
rbs-jacob: That could be one option. Or there could be an additional argument `use_cors` passed in `__init__` with a default value of `False` that a user could turn on when debugging. I'm open to other options, but I don't think CORS should be on by default.

marczalik: Is CORS just for debugging though? At this point, I'm not able to run the GUI without CORS on any of Firefox, Safari, or Chrome. Perhaps that is indicative of something else needing to be fixed?

ofrak_core/ofrak/service/script_builder.py
marczalik: Fair point, I cannot think of a use case for it outside of the GUI right now.

frontend/public/code.css
marczalik: Created issue #267

ofrak_core/ofrak/gui/server.py
```diff
@@ -20,8 +20,7 @@
     Any,
 )
 
-import orjson
```
dannyp303, line 23: Change back to orjson

ofrak_core/ofrak/gui/server.py
rbs-jacob: Yes, CORS should just be for debugging. You should be able to run the GUI without CORS as long as the GUI is built and copied into the folder that the server serves static files from. CORS is only necessary when running a separate GUI server on a different port. This should never happen in "production" and should only be for debugging.

marczalik: Should this option be made visible as a new CLI argument for the gui?

ofrak_core/ofrak/gui/server.py
dannyp303: yes...

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+        :return script:
+        """
+        root_resource = await self._get_root_resource(resource)
+        return self._get_script(root_resource.get_id())
+
+    def _get_script(self, resource_id: bytes, target_type: Optional[ActionType] = None) -> str:
+        script = []
+        script.append(self.script_sessions[resource_id].boilerplate_header)
+        for script_action in self.script_sessions[resource_id].hashed_actions.values():
+            if target_type is None or target_type == script_action.action_type:
```
EdwardLarson, line 240: Seems like `UNDEF` actions should be allowed here too, since that includes things like assigning values to variables, variables which might be used in later e.g. `UNPACK` actions (or whichever match the `target_type`). `target_type` seems more like it should be a set of allowed types, and then `get_all_of_type` determines what that set is.

Though tbh `get_all_of_type` is unused and I'm not sure it belongs in MVP anyway, so maybe `get_all_of_type` should just raise `NotImplementedError` and we don't try to filter by action type in this function.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+    def _dedent(self, s):
+        split = list(s.splitlines())
+        prefix = os.path.commonprefix([line for line in split if line != ""])
+        indent_matches = re.findall(r"^\s+", prefix)
+        if not indent_matches:
+            return s
+        indent_end_index = len(indent_matches[0])
+        return [line[indent_end_index:] for line in split if line.startswith(prefix) and line != ""]
+
+    def get_all_of_type(self, resource_id: bytes, target_type: ActionType) -> str:
```
EdwardLarson, line 258: Unused

EdwardLarson (changes requested):

frontend/src/ofrak/remote_resource.js
```diff
@@ -82,7 +83,7 @@ export class RemoteResource extends Resource {
     super(resource_id, data_id, parent_id, tags, caption, attributes);
 
     this.resource_list = resource_list;
-    this.uri = `/${this.resource_id}`;
+    this.uri = this.resource_id;
```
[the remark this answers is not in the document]
rbs-jacob: @EdwardLarson See [this discussion](https://github.com/redballoonsecurity/ofrak/pull/255#discussion_r1137816449).

frontend/src/ofrak/remote_resource.js
```diff
@@ -435,19 +463,36 @@ export class RemoteResource extends Resource {
...
+    ).then(async (r) => {
+      if (!r.ok) {
+        throw Error(JSON.stringify(await r.json(), undefined, 2));
+      }
+      return await r.json();
+    });
+    return remote_models_to_resources(matching_models);
+  }
+
+  async get_script() {
```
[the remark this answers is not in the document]
rbs-jacob: Perhaps `fetch_script`?

ofrak_core/ofrak/gui/server.py
rbs-jacob: Might as well be a new argument, I think that's a good idea.

frontend/src/ofrak/remote_resource.js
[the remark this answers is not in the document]
EdwardLarson: I considered that, I didn't suggest it because "fetch" still carries some connotation of returning a value, though less so than "get." I feel "update" is the most clear verb in this case.

frontend/src/ofrak/remote_resource.js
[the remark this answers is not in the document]
EdwardLarson: That discussion doesn't really explain why the change was made the way it was made or why one would even want to do it that way. We made a variable so that if we make general changes to the backend request URL, we change only that one place. Then for some reason made the change in every place the variable is used and turned the "uri" variable into just the resource ID without even changing the name.

frontend/src/App.svelte
```diff
@@ -164,13 +169,18 @@ Answer by running riddle.answer('your answer here') from the console.`);
...
               rootResource="{rootResource}"
+              bind:bottomLeftPane="{bottomLeftPane}"
               bind:resourceNodeDataMap="{resourceNodeDataMap}"
               bind:modifierView="{modifierView}"
             />
           {/if}
         </Pane>
         <Pane slot="second" paddingVertical="{'1em'}">
-          <AttributesView resource="{currentResource}" />
+          {#if bottomLeftPane}
```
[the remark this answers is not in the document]
rbs-jacob: It was not supposed to be a boolean based on what @marczalik and I discussed. It is supposed to be the actual "class" of component that is displayed in the bottom left pane (similar to `modifierView`). This if statement should check if that variable is set, or if it is `undefined` or `null`. Upon a second pass, it seems that it wasn't actually implemented this way.

When `bottomLeftPane` is implemented the same way as `modifierView`, the interfaces for `AttributesView` and `ScriptView` don't need to be interchangeable because `bottomLeftPane` will never have a value of `AttributesView`, it will only be `ScriptView`, `undefined`, or something that conforms to the same interface as `ScriptView`. 

For an example in the code base where there are panes that can optionally occupy the same space, but which have different interfaces, see the hex view/text view/disassembly view area.

https://github.com/redballoonsecurity/ofrak/blob/c08eebf3d7f99d16da0d0500373c237e84f19959/frontend/src/App.svelte#L181-L192

frontend/src/App.svelte
```diff
@@ -164,13 +169,18 @@ Answer by running riddle.answer('your answer here') from the console.`);
...
+            <ScriptView bind:bottomLeftPane="{bottomLeftPane}" />
```
rbs-jacob, line 180: As per [this discussion](https://github.com/redballoonsecurity/ofrak/pull/265/files/b2dda78d083b1800cb4003bc3011d18b550f60b9#r1147805129) in the preceding comment, this should be changed to the following, with a corresponding change in the `ResourceTreeView` component.

```suggestion
            <svelte:component this="{bottomLeftPane}" />
```

frontend/src/App.svelte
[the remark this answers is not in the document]
rbs-jacob: Specifically, see this change in `ResourceTreeToolbar` that sets that bound variable:

https://github.com/redballoonsecurity/ofrak/blob/b2dda78d083b1800cb4003bc3011d18b550f60b9/frontend/src/ResourceTreeToolbar.svelte#L221

It should be exactly like the `modifierView` variable. In [this comment](https://github.com/redballoonsecurity/ofrak/pull/265/files/b2dda78d083b1800cb4003bc3011d18b550f60b9#r1147959815) I make a suggestion to change the code accordingly.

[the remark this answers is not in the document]
rbs-jacob: One additional note is that (whether for `modifierView` or for `bottomLeftPane`), Svelte is very flexible about passing extra (or not enough) parameters to a component. The component will be instantiated and initialized just fine, it will just have some variables that are `undefined`. As long as the component checks this case (which it should anyway!) there won't actually be any problems.

That's a long way of saying that it doesn't actually matter if they implement the same interface. It will still work.

frontend/src/ofrak/remote_resource.js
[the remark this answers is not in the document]
rbs-jacob: I'm just pointing out that I suggested it, and that Dan presumably had some reason for choosing to do it his way :)

In general, I agree with your proposed change.

ofrak_core/ofrak/gui/script_builder.py
marczalik: The intention is for this to be used when we need to support selectively creating scripts based on action types (e.g., "get all steps that involved unpacking down to this resource so that I can repeat this unpacking process again").

frontend/src/App.svelte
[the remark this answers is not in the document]
marczalik: Just a quick clarification here:
>  It is supposed to be the actual "class" of component that is displayed in the bottom left pane (similar to modifierView). This if statement should check if that variable is set, or if it is undefined or null. Upon a second pass, it seems that it wasn't actually implemented this way.
`bottomLeftPane` is not a boolean and it does contain the class of the component selected as seen here
https://github.com/redballoonsecurity/ofrak/blob/b2dda78d083b1800cb4003bc3011d18b550f60b9/frontend/src/ResourceTreeToolbar.svelte#L221

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        var_name = await self.add_variable(resource)
+        qualified_action = action.replace("$resource", var_name)
+        await self._add_action_to_session(resource, qualified_action, action_type)
+
+    async def _add_action_to_session(self, resource, action, action_type):
+        root_resource = await self._get_root_resource(resource)
+        session = self._get_session(root_resource.get_id())
+        # TODO: actions are duplicated if page is refreshed, is this reasonable?
+        session.hashed_actions[session.actions_counter] = ScriptAction(action_type, action)
```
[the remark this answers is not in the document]
marczalik: My original thought was that a dict would be better for supporting deletions from the actions list; however, we've fixed that up by not recording actions for unnecessary fetches (e.g., get_data). Also, after thinking about it a bit further, an OrderedDict with int indexes wouldn't offer any improvement for arbitrary deletions (if we needed that for any additional reason) over lists. I will go ahead and make this change.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+                r_filter=ResourceFilter(
+                    tags=resource.get_most_specific_tags(),
+                    attribute_filters=[
+                        ResourceAttributeValueFilter(attribute=attribute, value=attribute_value)
+                    ],
+                )
+            )
+        except:
+            raise SelectableAttributesError(
+                f"Resource with ID {resource.get_id()} does not have a selectable attribute."
```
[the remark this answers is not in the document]
marczalik: > (this LoC will probably be too long so re-lint if you apply this suggestion)

Black strongly prefers not to add trailing `\` to break up long strings into multiple lines. So in this case, it doesn't automatically get split apart. Which is preferable: manually splitting up the LOC with `\`, splitting it up into 2 strings on different lines, or leaving it as-is?

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+            f"Resource with ID {resource.get_id()} does not have a selectable attribute."
+        )
+
+    async def _generate_name(self, resource: Resource) -> str:
+        root_resource = await self._get_root_resource(resource)
+        # Find the most specific tag and use that with a number
+        most_specific_tag = list(resource.get_most_specific_tags())[0].__name__.lower()
+        _, selectable_attribute_value = await self._get_selectable_attribute(resource)
+        name = f"{most_specific_tag}_{selectable_attribute_value}"
+        name = re.sub(r"[\-\.\/\\]", "_", name)
```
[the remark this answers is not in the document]
marczalik: @EdwardLarson @dannyp303 it looks like the appropriate regex for capturing whitespace would be either `[ \-\.\/\\]` (note the leading literal space character) or `[\s\-\.\/\\]` depending on whether we want to capture more whitespace than just spaces. Thoughts on which is more appropriate for our use case?

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+                print(e)
+                continue
+            return attribute, attribute_value
+        raise SelectableAttributesError(
+            f"Resource with ID {resource.get_id()} does not have a selectable attribute."
+        )
+
+    async def _generate_name(self, resource: Resource) -> str:
+        root_resource = await self._get_root_resource(resource)
+        # Find the most specific tag and use that with a number
```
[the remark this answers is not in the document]
dannyp303: Yeah, this is an artifact from an older implementation, please remove thank you :)

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,280 @@
...
+    boilerplate_header: str = r"""
+    from ofrak import *
+    from ofrak.core import *
+
+    async def main(ofrak_context: OFRAKContext):
+    """
+    boilerplate_footer: str = r"""
+    if __name__ == "__main__":
+
+        # Use ANGR Backend
```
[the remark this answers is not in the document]
EdwardLarson: In the `OFRAK` instance (not `OfrakContext` unfortunately) there is a list `_discovered_modules` which should have what we want, e.g. `[ofrak, ofrak_angr, ofrak_capstone]`

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,273 @@
...
+        self, resource: Resource
+    ) -> Tuple[ResourceIndexedAttribute, any]:
+        for attribute in self.selectable_indexes:
+            try:
+                await resource.analyze(attribute.attributes_owner)
+                attribute_value = attribute.get_value(resource.get_model())
+            except Exception as e:
+                print(e)
+                continue
+            return attribute, attribute_value
```
EdwardLarson, line 137:
```suggestion
        for attribute in self.selectable_indexes:
            if resource.has_attributes(attribute.attributes_owner):
                attribute_value = attribute.get_value(resource.get_model())
                return attribute, attribute_value
```

ofrak_core/ofrak/gui/script_builder.py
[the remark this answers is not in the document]
EdwardLarson: @marczalik Just go with whatever `black` wants to do, I don't think it ever uses backslashes to wrap strings and letting python auto-concat string literals on consecutive lines does the job.

@dannyp303 I see, that will be a good future extension. I suppose the reason why the chosen selector doesn't work is not of interest to the user in that context, so we don't really need to refine & define the reason. But, I still think this code should be refactored a little to form the loop I described. It seems logical, even for MVP, that if the resource has a selectable attribute that for some reason doesn't work, it will try the next one. And only if all selectable attributes either fail to be found, or don't form a successful query, do we then prompt the user.

ofrak_core/ofrak/gui/server.py
```diff
@@ -300,63 +329,77 @@ async def get_resource_range(resource_id):
...
         result = await resource.run(DataSummaryAnalyzer)
-        response_pjson = await self._serialize_component_result(result)
-        return json_response(response_pjson)
+
+        return json_response(await self._serialize_component_result(result))
 
     @exceptions_to_http(SerializedError)
     async def analyze(self, request: Request) -> Response:
         resource = await self._get_resource_for_request(request)
+        script_str = """
```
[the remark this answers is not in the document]
marczalik: I agree that the way the script strings are currently constructed relies on non-obvious behavior (by requiring a leading newline) and that it would be better to handle that in `ScriptBuilder`. However, in order to remove the leading newline the resulting docstrings end up much worse, in my opinion.

For example, in order to get the alignment correct the docstrings for the script header and `create_root_resource` need to look like:
```
    boilerplate_header: str = (
r"""    from ofrak import *
    from ofrak.core import *

    async def main(ofrak_context: OFRAKContext):
    """
    )
```

```
        script_str = (
f"""        with open("{name}", "rb") as fh:
            resource_data = fh.read()
        root_resource = await ofrak_context.create_root_resource_from_file("{name}")"""

```

ofrak_core/ofrak/gui/script_builder.py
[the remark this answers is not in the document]
marczalik: As discussed offline, replaced with `if False` blocks for now until we can determine a good way to provide ScriptBuilder with the the current backend.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+
+        if False:
+            import ofrak_binary_ninja
+            import ofrak_capstone
+            ofrak.discover(ofrak_capstone)
+            ofrak.discover(ofrak_binary_ninja)
+
+        if False:
+            import ofrak_ghidra
+            ofrak.discover(ofrak_ghidra)
```
rbs-jacob, line 71: Running `black` changes this, so we probably want to have the version here match the version `black` outputs.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        :return script:
+        """
+        root_resource = await self._get_root_resource(resource)
+        return self._get_script(root_resource.get_id())
+
+    async def get_all_of_type(self, resource: Resource, target_type: ActionType) -> str:
+        """
+        :param target_type:
+        :return script:
```
rbs-jacob, line 151: In my opinion, these should either be filled in, or removed. Same goes for all instances of the parameters in other doc comments throughout the PR.

ofrak_core/ofrak/gui/server.py
```diff
@@ -410,39 +458,75 @@ async def queue_patch(self, request: Request) -> Response:
...
+        # TODO: There has to be a better way
+        new_data_string = "\\x" + "\\x".join(
+            [new_data.hex()[i : i + 2] for i in range(0, len(new_data.hex()), 2)]
+        )
+        script_str = (
+            r"""
+        {resource}.queue_patch"""
+            rf"""(Range({start}, {end}), b"{new_data_string}")"""
+        )
+        """await {resource}.save()"""
```
rbs-jacob, line 470: Is this string supposed to be dead code?

```suggestion
```

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        root_resource = await self._get_root_resource(resource)
+        return self._get_script(root_resource.get_id(), target_type)
+
+    async def _get_root_resource(self, resource: Resource) -> Resource:
+        if resource.get_id() in self.root_cache:
+            return self.root_cache[resource.get_id()]
+        while len(list(await resource.get_ancestors())) != 0:
+            resource = await resource.get_parent()
+        self.root_cache[resource.get_id()] = resource
```
rbs-jacob, line 161: This does not properly update the cache. It uses the ID of the root and sets its value to the root. The ID should be stored in a variable before the destructive loop above.

At the moment, the cache is effectively unused.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+        script = []
+        script.append(self.script_sessions[resource_id].boilerplate_header)
+        for script_action in self.script_sessions[resource_id].actions:
+            # Always include UNDEF actions like variable assignments
+            if (
+                target_type is None
+                or script_action.action_type == target_type
+                or script_action.action_type == ActionType.UNDEF
+            ):
+                script.append(f"{script_action.action}")
```
rbs-jacob, line 282:
```suggestion
                script.append(script_action.action)
```

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+        script.append(self.script_sessions[resource_id].boilerplate_footer)
+        script = "\n".join(script)
+        script = self._dedent(script)
+        res = format_str("\n".join(script), mode=FileMode())
```
rbs-jacob, line 286: The script is already joined?

```suggestion
        res = format_str(script, mode=FileMode())
```

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+        script = res.split("\n")
```
rbs-jacob, line 287: What is the benefit of splitting the script into lines? I don't really see one, and if there isn't one, it would obsolete and obviate error-prone code in several places if we just passed around a string.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+    async def delete_action(self, resource: Resource, action: str) -> None:
+        """
+        :param action:
+        """
+        root_resource = await self._get_root_resource(resource)
+        session = self._get_session(root_resource.get_id())
+        # TODO: do we really need to delete an action from the script?
+        for idx, script_action in enumerate(session.actions):
+            if script_action.action == action:
+                session.actions.pop(idx)
```
rbs-jacob, line 139: Should this break after, or should it remove multiple? If multiple, I suggest that the function be renamed to `delete_actions`.

```suggestion
        for idx, script_action in enumerate(session.actions):
            if script_action.action == action:
                session.actions.pop(idx)
                break
```

ofrak_core/ofrak/gui/server.py
```diff
@@ -485,6 +575,11 @@ async def get_all_tags(self, request: Request) -> Response:
             self._serializer.to_pjson(self._ofrak_context.get_all_tags(), Set[ResourceTag])
         )
 
+    @exceptions_to_http(SerializedError)
+    async def get_script(self, request: Request) -> Response:
+        resource = await self._get_resource_for_request(request)
+        return json_response(await self.script_builder.get_script(resource))
```
rbs-jacob, line 581: I have not debugged why, but this is currently returning one single script for all root resources. It's supposed to be returning a different script for each root resource.

In other words, this is currently not working.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+
+    async def _var_exists(self, resource: Resource):
+        root_resource = await self._get_root_resource(resource)
+        session = self._get_session(root_resource)
+        return resource.get_id() in session.resource_variable_names
+
+    async def _add_action_to_session(self, resource, action, action_type):
+        root_resource = await self._get_root_resource(resource)
+        session = self._get_session(root_resource.get_id())
+        # TODO: actions are duplicated if page is refreshed, is this reasonable?
```
rbs-jacob, line 178: This is not reasonable in my opinion. Why does this happen?

I also don't see this happening in practice. Is this still true?

rbs-jacob (changes requested):
A bunch of changes that I still think need to be made before this is ready.

In particular, it feels like there were a couple of issues that would have been caught by actually running this a bunch, and generating lots of different scripts for each of the different actions. I did this a bit, but wouldn't be comfortable giving this PR the okay until this is run and tested a bit more.

frontend/src/ResourceTreeToolbar.svelte
```diff
@@ -212,6 +213,14 @@
           modifierView = SearchView;
         },
       },
+
+      {
+        text: "Show Script",
```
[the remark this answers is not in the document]
marczalik: "Build" and "Generate" both seem to imply one-time actions - "Build the script at this point in time" - except the script is continually updated. Furthermore, "Show" (or something like "Display") is more accurate for what is actually happening. That is, the script is regularly being sent to the frontend whenever a user clicks a button, and the frontend always has the most up-to-date script stored. Clicking the button displays the script, it doesn't build it (or, as the button was previously labeled, get the script).

ofrak_core/ofrak/gui/script_builder.py
marczalik: `dedent` expects a string and returns a list. The second join could be moved inside of `dedent` so that it takes a string and returns a string, although that's just shifting the operation around rather than removing it. I did a quick experiment and it doesn't appear that it's straightforward to remove the first join and have `dedent` take a list. The reason being is that the the `splitlines` call in `dedent` removes '\n' chars that are present in the `script` list before the first call to `join`, so the two lists are not identical.

marczalik: `dedent` expects a string and returns a list. The second join could be moved inside of `dedent` so that it takes a string and returns a string, although that's just shifting the operation around rather than removing it. I did a quick experiment and it doesn't appear that it's straightforward to remove the first join and have `dedent` take a list. The reason being is that the the `splitlines` call in `dedent` removes '\n' chars that are present in the `script` list before the first call to `join`, so the two lists are not identical.

ofrak_core/ofrak/gui/script_builder.py
marczalik: The primary benefit is that it simplifies line numbering on the frontend. Rather than having to parse the script on the frontend to increment the line number each time there's a newline, line numbering is easily handled by setting the current line number to the index in the list plus one.

ofrak_core/ofrak/gui/script_builder.py
marczalik: I did not see it happening as of yesterday, I just hadn't committed the change to remove it since it was the only change I had at the time and expected that more would need to be made.

marczalik: To answer the first question: it is likely, although I don't recall exactly which actions I was seeing duplicated at the time that I made this comment, that the actions being duplicated were the ones that occur strictly for the frontend to work and don't need to be recorded in the script (e.g., get_data, get_children, etc). Once we removed those, this issue was addressed as well.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,297 @@
...
+        else:
+            msg = []
+            for collision, value in attribute_collisions.items(): 
+                msg.append(f"Resource with ID {resource.get_id()} cannot be uniquely identified by attribute {attribute.__name__} (resource has value {attribute_value}).")
+            raise SelectableAttributesError("\n".join(msg))
+
+    async def _generate_name(self, resource: Resource) -> str:
+        root_resource = await self._get_root_resource(resource)
+        most_specific_tag = list(resource.get_most_specific_tags())[0].__name__.lower()
+        _, selectable_attribute_value = await self._get_selectable_attribute(resource)
```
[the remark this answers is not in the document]
marczalik: Won't this change work as intended, without causing issues to the value used in `ResourceAttributeValueFilter`? This change will only affect the string representation in the new variable name; it doesn't alter the value used in `_get_selector` or `_get_selectable_attribute` where the instances of `ResourceAttributeValueFilter` occur.

I believe this change can be made as written. See this sample output:
```
    coderegion_0x8090 = await root_resource.get_only_child(
        r_filter=ResourceFilter(
            tags={CodeRegion, ElfSection},
            attribute_filters=[
                ResourceAttributeValueFilter(attribute=Data.Offset, value=32912)
            ],
        )
    )
```

ofrak_core/ofrak/gui/script_builder.py
rbs-jacob: Cool. If this is no longer an issue, feel free to remove the `TODO` and resolve this thread. Thanks!

ofrak_core/ofrak/gui/script_builder.py
rbs-jacob: I'm not sure that is worth passing everything around as a list of lines, since the string can just as easily be split on the front end. But I also think at this point it doesn't need to be changed.

ofrak_core/ofrak/gui/script_builder.py
rbs-jacob: No need for `dedent` to take a list. I do think it would be cleaner if `dedent` took a string and returned a string. That interface just feels a little more natural to me.

ofrak_core/ofrak/gui/server.py
marczalik: This should actually be included in the script so that the queued patch is saved into the script, no? I had accidentally left this out when adjusting the script string resource name replacements from `$resource` to `{resource}`.

Unfortunately, as yet another example of the fragility of how the script builder handles whitespace for scripts, there's no elegant way to add this onto the existing script string in this route handler, which is using auto-concatenation to build the `queue_patch()` call. Black isn't able to parse the input script if I simply include variations of `"""\nawait {resource}.save()"""` to the existing string. In order to handle thing, I need to create and add a second action within this route handler, which doesn't seem ideal. It works though.

```
        script_str = (
            r"""
        {resource}.queue_patch"""
            rf"""(Range({start}, {end}), b"{new_data_string}")"""
        )
        await self.script_builder.add_action(resource, script_str, ActionType.MOD)
        script_str = """
        await {resource}.save()"""
        await self.script_builder.add_action(resource, script_str, ActionType.MOD)
```

ofrak_core/ofrak/gui/server.py
marczalik: This should be resolved now. The issue was that the same instances of ScriptActions and ScriptSessions were being shared by multiple files.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,348 @@
...
+        root_resource = await self._get_root_resource(resource)
+        for ancestor in await resource.get_ancestors():
+            if (
+                ancestor.get_id()
+                in self.script_sessions[root_resource.get_id()].resource_variable_names
+            ):
+                break
+        attribute, attribute_value = await self._get_selectable_attribute(resource)
+        try:
+            result = await ancestor.get_only_descendant(
```
EdwardLarson, line 231: I'm a little confused why the code is searching up through the ancestors here - the code chunk returned by the function only ever contains a `get_only_child` call, and my memory is that that is the intent - we only  select resources one "level" at a time. What is the purpose of this loop?

ofrak_core/ofrak/gui/server.py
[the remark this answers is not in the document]
EdwardLarson: Multiline strings that are tricky to format can of course still have leading newlines when it helps, redundant newlines shouldn't harm the generated script, but generally it seems like the server code will look a lot nicer if the script builder always adds newlines (reminds me of a difference between C's `printf` and Python's `print`, I often forget I need a trailing newline when I go back to C and once it bit me badly!).

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,348 @@
...
+    a session.
+    """
+
+    def __init__(self):
+        self.root_cache: Dict[bytes, Resource] = {}
+        self.script_sessions: Dict[bytes, ScriptSession] = {}
+        self.selectable_indexes: List[ResourceIndexedAttribute] = [
+            FilesystemEntry.Name,
+            Data.Offset,
+        ]
```
rbs-jacob, line 89: I think we should try using virtual addresses as identifiers before we use offsets. Not only will virtual addresses be unique more than offsets will, but they will make the generated code more closely correspond with the displayed info in the GUI. 

For example, without this, a complex block in the GUI might be shown as: `0xd80: main`, but in the generated code it is `complexblock_0x2e0` because that is its offset. I find this confusing, and think using virtual addresses when possible would address it.

```suggestion
        self.selectable_indexes: List[ResourceIndexedAttribute] = [
            FilesystemEntry.Name,
            Addressable.VirtualAddress,
            Data.Offset,
        ]
```

ofrak_core/ofrak/gui/server.py
```diff
@@ -475,6 +563,10 @@ async def search_for_vaddr(self, request: Request) -> Response:
     async def add_tag(self, request: Request) -> Response:
         resource = await self._get_resource_for_request(request)
         tag = self._serializer.from_pjson(await request.json(), ResourceTag)
+        script_str = """
+        {resource}.add_tag(tag)
```
rbs-jacob, line 567: Should this be something like the following? Right now it adds a tag from a variable that doesn't exist.

```suggestion
        {resource}.add_tag({tag.__name__})
```

ofrak_core/ofrak/gui/script_builder.py
marczalik: I think the issue is that the code chunk returned needs to be updated from `get_only_child` to `get_only_descendant`. @dannyp303 please correct me if I'm wrong, but I think that this loop uses the nearest ancestor which has been previously uniquely ID'd as the starting point from which to uniquely ID the current resource, allowing us to select resources at any level (i.e., enabling `unpack_recursively`).

frontend/src/ofrak/remote_resource.js
```diff
@@ -271,6 +282,8 @@ export class RemoteResource extends Resource {
     ingest_component_results(analyze_results, this.resource_list);
     this.flush_cache();
     this.update();
+
+    await this.update_script();
```
rbs-jacob, line 286: The script should update after `queue_patch` below, but doesn't. Please do a pass over the back end routes and make sure that all routes that change the script have a corresponding call to `update_script` in the front end!

ofrak_core/ofrak/gui/server.py
```diff
@@ -410,39 +458,77 @@ async def queue_patch(self, request: Request) -> Response:
...
         end = int(end_param) if end_param is not None else (await resource.get_data_length())
-
+        # TODO: There has to be a better way
+        new_data_string = "\\x" + "\\x".join(
+            [new_data.hex()[i : i + 2] for i in range(0, len(new_data.hex()), 2)]
+        )
+        script_str = (
+            r"""
+        {resource}.queue_patch"""
+            rf"""(Range({start}, {end}), b"{new_data_string}")"""
```
rbs-jacob, line 468: The `r` preceding the string is not necessary. And the comment modifier does ranges using hex, so this should probably be consistent with that. In fact, any generated script ranges should be consistently using hex offsets. Please do a pass over the generated script strings looking for places with `Range` that may not be using hex, but should.

```suggestion
            f"""(Range({hex(start)}, {hex(end)}), b"{new_data_string}")"""
```

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,348 @@
...
+        action_type: ActionType,
+    ) -> None:
+        """
+        Adds an action to the script session to which the selected resource belongs. An action is
+        a string representing the code that is being run on the resource based on an action that
+        has occurred in the GUI.
+
+        :param resource: Resource upon which the action is being taken.
+        :param action: A string describing the code being run based on a GUI action.
+        :param action_type: An instance of `ActionType` categorizing the action.
```
rbs-jacob, line 104: I really like the docstrings! They're quite helpful. 

Please make sure they match with the [OFRAK contributor guide style conventions](https://ofrak.com/docs/contributor-guide/getting-started.html#functions-and-methods)!

rbs-jacob (changes requested):
Still some outstanding changes that I think would have been caught by running the GUI for a while and clicking around. Please do this before this PR gets merged! I am confident it will catch other issues Edward or I may have missed.

ofrak_core/ofrak/gui/script_builder.py
marczalik: I've added `raises` to `add_variable`, removed the periods from non-general descriptions, and added backticks are what I consider to be "named code artifacts" (slightly unsure of what 'artifacts' means in this context, since nothing is being left behind).

rbs-jacob:
> @rbs-jacob @EdwardLarson one potential issue I'm seeing with the script is that, if the user does something that throws an error within OFRAK, that action is still recorded in the script. And because an error is returned, the script isn't updated automatically (the command will show up in the script when the next valid action runs). The easiest way to see this in action is to try and run `StringFindReplaceModifier` and replacing the first string with a longer string.

Marc and I talked offline, and we agreed that the solution is to make sure the script builder is the absolute last thing that runs in each route before returning a response. That way, if there are any OFRAK actions that raise exceptions, the script builder will not add lines for them to the script.

rbs-jacob (changes requested):
The script builder is causing issues when unpacking children of DTB files and raising errors such as the following.

image

Even though the OFRAK action succeeds, the script builder fails with an error. This behavior *must* be fixed before this PR is merged, because it is preventing unpacking DTB children (and likely other file types) in the GUI, even though there is nothing stopping OFRAK itself from doing so.

My understanding is that this exception was added to give the user the opportunity to pick an indexable attribute for making the resource selection. While we should come back to making this possible, this should not raise an exception that blocks the route. Rather, it should log a warning that the user can see somewhere to know that their script may have errors. 

In general, we should do a pass over this code and make sure that exceptions with the script builder will not accidentally break regular OFRAK functionality in other ways.

EdwardLarson:
@rbs-jacob Thanks for catching these issues. As a general solution to the problem of script generation errors blocking working GUI actions, we can refactor this code to only actually generate the script (including getting selectors etc.) when the user presses the "Generate(/View/Build) Script" button. I believe actions already are basically handled by this - it's `add_variable` that seems most error prone, and would need the most work to refactor.

ofrak_core/ofrak/gui/server.py
```diff
@@ -205,11 +231,18 @@ async def create_root_resource(self, request: Request) -> Response:
         root_resource = await self._ofrak_context.create_root_resource(name, resource_data, (File,))
         if request.remote is not None:
             self._job_ids[request.remote] = root_resource.get_job_id()
+
+        script_str = rf"""
+        with open("{name}", "rb") as fh:
+            resource_data = fh.read()
+        root_resource = await ofrak_context.create_root_resource_from_file("{name}")"""
```
EdwardLarson, line 238:
```suggestion
        root_resource = await ofrak_context.create_root_resource_from_file("{name}")"""
```
`resource_data` is unused, don't need to open and read the file if we are going to use `create_root_resource_from_file`

marczalik:
> @rbs-jacob Thanks for catching these issues. As a general solution to the problem of script generation errors blocking working GUI actions, we can refactor this code to only actually generate the script (including getting selectors etc.) when the user presses the "Generate(/View/Build) Script" button. I believe actions already are basically handled like this - it's add_variable that seems most error prone, and would need the most work to refactor.

@EdwardLarson @rbs-jacob @dannyp303  we discussed something like this offline and the issue with this approach is that, if the script is generated only at the time that the user clicks "generate/show script" then the script generation will be created on a version of the state of the internal resources that is not consistent with its state at the time of the actions originally taking place. In particular, if you perform a series of actions that modify the attributes of a resource, then by the time you generate the script there may be selectable attributes on that resource that wouldn't have existed at the time that you perform the action.

marczalik:
@dannyp303 @rbs-jacob @EdwardLarson I have removed the unused methods `delete_action` and `get_all_of_type` from ScriptBuilder since they are not being used in the MVP. If/when we need that functionality we can add it back in.

ofrak_core/ofrak/gui/server.py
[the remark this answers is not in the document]
marczalik: @EdwardLarson  As discussed offline, resolving this for MVP and recording it as a future improvement.

ofrak_core/ofrak/gui/script_builder.py
marczalik: @EdwardLarson resolving based on changes made to refactor this method.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,405 @@
...
+
+    async def _get_selector(self, resource: Resource) -> str:
+        root_resource = await self._get_root_resource(resource)
+        session = self._get_session(root_resource.get_id())
+        parent = await resource.get_parent()
+        attribute, attribute_value = await self._get_selectable_attribute(resource)
+        await self._test_selectable_attributes(parent, resource, attribute, attribute_value)
+
+        if isinstance(attribute_value, str) or isinstance(attribute_value, bytes):
+            attribute_value = f'"{attribute_value!s}"'.rstrip()
```
EdwardLarson, line 293:
```suggestion
            attribute_value = f"{attribute_value!r}"
```
The ["!r"](https://docs.python.org/3.7/tutorial/inputoutput.html) seems like a slightly simpler way to accomplish what we want here, automatically adding quotations if necessary. Though I had to look up "!s" and that's how I found "!r" so maybe you've experimented here a bit more.

In fact I think we can always format the attribute value this way, not just for `str` and `bytes`. Selectable attributes are always indexable attributes which are always primitive types. It has the same result for the other primitives as they are formatted now, but it removes an extra check.

And is there any reason for the `.rstrip()`? At best it seems like it should have no effect because the string is ended with non-whitespace.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,405 @@
...
+        {name} = {selector}""",
+                ActionType.UNDEF,
+            )
+            await self._add_variable_to_session_queue(resource, name)
+        except SelectableAttributesError as e:
+            name = await self._generate_missing_name(resource, e)
+            LOGGER.exception("Could not find selectable attributes for resource")
+            return name
+        except:
+            LOGGER.exception("Exception raised in add_variable")
```
rbs-jacob, line 201: This should probably print out the exception that gets raised.

ofrak_core/ofrak/gui/server.py
```diff
@@ -300,63 +337,103 @@ async def get_resource_range(resource_id):
...
+            await self.script_builder.clear_script_queue(resource)
+            raise e
+        return json_response(await self._serialize_component_result(result))
 
     @exceptions_to_http(SerializedError)
     async def pack(self, request: Request) -> Response:
         resource = await self._get_resource_for_request(request)
+        script_str = """
+
+        await {resource}.pack()"""
```
rbs-jacob, line 370:
```suggestion
        script_str = """
        await {resource}.pack()"""
```

This spacing is inconsistent with other component calls, but also with `pack_recursively` below. If the intention was to space out the `pack` call, then `pack_recursively` should be spaced out as well. Otherwise, this space should probably be removed.

ofrak_core/ofrak/gui/server.py
```diff
@@ -300,63 +337,103 @@ async def get_resource_range(resource_id):
...
+        await self.script_builder.add_action(resource, script_str, ActionType.PACK)
         result = await resource.pack()
-        response_pjson = await self._serialize_component_result(result)
-        return json_response(response_pjson)
+        await self.script_builder.commit_to_script(resource)
+        return json_response(await self._serialize_component_result(result))
```
rbs-jacob, line 374: Does this not need the `try`/`except` structure of the other route handlers?

rbs-jacob (changes requested):
Some final, small changes.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,353 @@
...
+                r_filter=ResourceFilter(
+                    tags=resource.get_most_specific_tags(),
+                    attribute_filters=[
+                        ResourceAttributeValueFilter(attribute=attribute, value=attribute_value)
+                    ],
+                )
+            )
+        except:
+            raise SelectableAttributesError(
+                f"Resource with ID {resource.get_id()} cannot be uniquely identified by attribute "
```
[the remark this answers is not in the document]
marczalik: @rbs-jacob Are you looking at the most recent changes? This has already been applied.

ofrak_core/ofrak/gui/script_builder.py
```diff
@@ -0,0 +1,353 @@
...
+                        )
+                    )
+                )
+                if len(children) > 1:
+                    attribute_collisions[attribute.__name__] = attribute_value
+                    continue
+                return attribute, attribute_value
+        if len(attribute_collisions) == 0:
+            raise SelectableAttributesError(
+                f"Resource with ID {resource.get_id()} does not have a selectable attribute."
```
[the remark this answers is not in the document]
marczalik: @rbs-jacob  same as above

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -379,3 +394,267 @@ async def test_add_tag(ofrak_client: TestClient, hello_world_elf):
...
+
+
+async def test_update_script(ofrak_client: TestClient, hello_world_elf):
+    create_resp = await ofrak_client.post(
+        "/create_root_resource", params={"name": "hello_world_elf"}, data=hello_world_elf
+    )
+    root = await create_resp.json()
+    root_id = root["id"]
+
+    # Perform some actions that should hit add_action, add_variable, and all their subcalls
```
EdwardLarson, line 406: So there are no unit tests that actually explicitly call `add_action`, `add_variable`, etc.?

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -46,6 +48,19 @@ async def test_resource(ofrak_context, hello_world_elf):
...
 
 
+def dicts_are_similar(d1, d2, attributes_to_skip=None):
+    if attributes_to_skip is None:
+        attributes_to_skip = {"id", "data_id"}
+    for key, value in d1.items():
+        if key in attributes_to_skip:
+            continue
+        if isinstance(value, list) and set(value) != set(d2[key]):
+            return False
```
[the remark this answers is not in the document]
marczalik: @rbs-jacob The list of tags returned for each child in the response to `unpack` is unordered, which necessitated this case. I can add that as a comment.

ofrak_core/ofrak/gui/script_builder.py
marczalik: `exception` prints out the message argument as well as the stack trace of the exception that raised it, including the error message of the handled exception. For example:
image

Please let me know if this still needs to be changed.

ofrak_core/ofrak/gui/script_builder.py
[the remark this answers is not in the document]
jstrieb: It says suggested 5 days ago. No idea why GitHub is displaying it here

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -379,3 +394,267 @@ async def test_add_tag(ofrak_client: TestClient, hello_world_elf):
...
+        "    raise RuntimeError(",
+        '        "Resource with ID 0x00000002 cannot be uniquely identified by attribute Data.Offset (resource has value 0)."',
+        "    )",
+        "",
+        "    await root_resource_MISSING_RESOURCE_0.unpack()",
+        "",
+        "    # Resource with parent root_resource is missing, could not find selectable attributes.",
+        "    raise RuntimeError(",
+        '        "Resource with ID 0x00000003 cannot be uniquely identified by attribute Data.Offset (resource has value 0)."',
+        "    )",
```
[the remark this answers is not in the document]
marczalik: @rbs-jacob that makes sense. One issue I thought of in particular is what happens if the resource ID changes. At the moment they appear to be stable across runs, but will that always be the case?

ofrak_core/test_ofrak/unit/test_ofrak_server.py
EdwardLarson: They don't have to be, and testing GUI interaction with them is necessary too.

EdwardLarson (approved):

ofrak_core/test_ofrak/unit/test_ofrak_server.py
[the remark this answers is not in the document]
rbs-jacob: Adding that as a comment would be appreciated. Feel free to resolve this when that's done.

ofrak_core/test_ofrak/unit/test_ofrak_server.py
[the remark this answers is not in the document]
rbs-jacob: That's a good point – they may not be stable across runs.

I'd add a comment that links to this thread, and suggests adding a "normalization" step wherein the test case reference and `resp_body` are canonicalized into a consistent format. 

Normalization/canonicalization would involve both the reference string and `resp_body` being joined into strings (instead of lists of strings) and then some regular expression replacements could be performed. For example, in this case, the replacement would be something like:

``` python
def normalize(list_of_strs):
    in_str = "\n".join(list_of_strs)
    return re.sub(r"RuntimeError(\n.*\n)", "RuntimeError(err)", in_str, re.M)
```

rbs-jacob (approved):

PR #269, opened 2023-03-28
add command line arg to import and discover additional code when star…

EdwardLarson:
…ting OFRAK from cli

* Add option to the CLI to import and discover additional OFRAK Python packages when starting OFRAK.

**Link to Related Issue(s)**
This is similar to the existing `--backend` option which lets users choose one of a static set of backends, but that option didn't allow users to easily extend OFRAK and load their extensions in the CLI.

* Add `-i`/`--import` option to dynamically import and discover additional Python modules
* These don't have to be installed modules; they can be a local file (e.g. `-i ./my_extension.py`), detected with the presence of the OS's path separator (e.g. "/" for Mac or Linux) or the ".py" file extension at the end. Importing these requires adding their directory to the system path.
* Limit the values passed to the "--backend" argument to only the valid disassemblers
* Change how handing the "--backend" argument is handled to use the same mechanism as `-i`/`--import`

@rbs-jacob 
Unfortunately doing these dynamic imports by package name adds a few seconds to the startup time.

### DIFF
--- a/ofrak_core/ofrak/cli/ofrak_cli.py
+++ b/ofrak_core/ofrak/cli/ofrak_cli.py
@@ -21,7 +21,7 @@ BACKEND_PACKAGES = {
     "angr": {"ofrak_angr", "ofrak_capstone"},
     "binary-ninja": {"ofrak_binary_ninja", "ofrak_capstone"},
     "ghidra": {"ofrak_ghidra"},
-    None: {},
+    None: set(),
 }
 
 
@@ -169,7 +169,7 @@ class OfrakCommandRunsScript(OfrakCommand, ABC):
         ofrak_pkgs = set(args.imports)
         ofrak_pkgs.update(BACKEND_PACKAGES.get(args.backend))
 
-        if not any(pkgs.issubset(ofrak_pkgs) for pkgs in BACKEND_PACKAGES.values()):
+        if not any(pkgs and pkgs.issubset(ofrak_pkgs) for pkgs in BACKEND_PACKAGES.values()):
             logging.warning("No disassembler backend specified, so no disassembly will be possible")
 
         for ofrak_pkg_name in ofrak_pkgs:
--- a/ofrak_core/ofrak/cli/ofrak_cli.py
+++ b/ofrak_core/ofrak/cli/ofrak_cli.py
@@ -7,7 +7,7 @@ from abc import ABC, abstractmethod
 from argparse import ArgumentParser, _SubParsersAction, Namespace
 from inspect import isabstract
 from types import ModuleType
-from typing import Dict, Iterable, List, Optional, Sequence, Type
+from typing import Dict, Iterable, List, Optional, Sequence, Type, Set
 
 from importlib_metadata import entry_points
 
@@ -16,8 +16,7 @@ from ofrak.model.component_model import ComponentExternalTool
 from ofrak.ofrak_context import OFRAKContext, OFRAK
 from synthol.injector import DependencyInjector
 
-
-BACKEND_PACKAGES = {
+BACKEND_PACKAGES: Dict[Optional[str], Set[str]] = {
     "angr": {"ofrak_angr", "ofrak_capstone"},
     "binary-ninja": {"ofrak_binary_ninja", "ofrak_capstone"},
     "ghidra": {"ofrak_ghidra"},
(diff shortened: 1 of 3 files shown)


ofrak_core/ofrak/cli/ofrak_cli.py
```diff
@@ -14,6 +16,13 @@
 from ofrak.ofrak_context import OFRAKContext, OFRAK
 from synthol.injector import DependencyInjector
 
+BACKEND_PACKAGES: Dict[Optional[str], Set[str]] = {
+    "angr": {"ofrak_angr", "ofrak_capstone"},
+    "binary-ninja": {"ofrak_binary_ninja", "ofrak_capstone"},
+    "ghidra": {"ofrak_ghidra"},
+    None: set(),
+}
```
rbs-jacob, line 24: This is pretty slick, and I like it.

ofrak_core/ofrak/cli/ofrak_cli.py
```diff
@@ -143,29 +165,25 @@ def run(self, ofrak_env: OFRAKEnvironment, args: Namespace):
...
 
-                ofrak.discover(ofrak_ghidra)
+        if not any(pkgs and pkgs.issubset(ofrak_pkgs) for pkgs in BACKEND_PACKAGES.values()):
+            logging.warning("No disassembler backend specified, so no disassembly will be possible")
 
-            elif args.backend.lower() == "angr":
-                import ofrak_capstone  # type: ignore
-                import ofrak_angr  # type: ignore
+        for ofrak_pkg_name in ofrak_pkgs:
+            if os.path.sep in ofrak_pkg_name or ofrak_pkg_name.endswith(".py"):
```
rbs-jacob, line 175: Is it worth checking instead for `os.path.exists` and then cleaning up the path after that? This would enable importing directories with an `__init__.py` as well as DLLs that end in `.dll`, `.so`, or `.dylib`. 

Might be able to use some combination of these two recipes to import from files without path manipulations:

- <https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly>
- <https://docs.python.org/3/library/importlib.html#approximating-importlib-import-module>

rbs-jacob (changes requested):
One minor suggested change. 

Otherwise, this looks like a great little feature! I love the slick use of the `BACKEND_PACKAGES`.

ofrak_core/ofrak/cli/ofrak_cli.py
EdwardLarson: I tried out a few things and did clean this up a little bit, but it's not clear to me what the correct Python procedure is to import files as you've described - the most correct way is not to do that, and make them modules instead!

I can keep looking into it if you can define the use case a little more, but at the moment it easily supports loading a Python file so you can quickly include some new scratch OFRAK extension as you're developing it. For everything else, I think we should just recommend installing it as a Python module we can import normally (i.e. by the module name).

rbs-jacob: I really like where it's at right now, and don't think it's worth blocking on this.

rbs-jacob (approved):

ofrak_core/ofrak/cli/ofrak_cli.py
EdwardLarson: I had to revert the changes to imports and went back to the jankier-feeling `sys.path.append` method. The import technique shown [here](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly) breaks synthol discovery: the contents of the imported file are considered "builtins" by inspect and so get filtered out by discovery. I believe this is because that way of importing defines the modules in the top-level `__main__`, and [this](https://stackoverflow.com/a/56177941/16690095) comment says that classes defined in `__main__` (e.g. defined in interpreter) are considered "builtins". I haven't found clear info in the Python docs to verify all that, but it could explain what's going on.

Anyway I found this by adding a unit test for this import feature to the CLI to get full function coverage, so another +1 to that 100% function coverage policy! 

The informal quick check I used to see if it worked was "working" because it contained a `MagicDescriptionIdentifier.register(...)` call, and the file was successfully identified. But in fact, nothing in that file was being "discovered" - it is __imported__ though, and so the class was correctly added as a tag by the `MagicDescriptionIdentifier` which sort of backdoored a scratch file class into OFRAK without it being discovered. On a closer at that scratch file, I found that the components in that file were indeed not discovered and would not be run.

Issue #275, opened 2023-04-05
CPIO Unpacker no work on local Linux (not container, local).


gskip17:
**What is the problem? (Here is where you provide a complete Traceback.)**

OFRAK relies on the `cpio` utility to unpack CPIO filesystems. The utility is called with `-id` flags. When unpacking the filesystem, if there are any absolute filepaths, the utility will crash because it should (rightly) not try and overwrite a user's system files. 

```
ofrak.service.job_service_i.ComponentAutoRunFailure: Component CpioUnpacker failed when running on e1348bce32ac4798a453715762d4898b. Component was chosen because it matched filters (((ComponentTypeFilter(Unpacker) or ComponentTypeFilter(Identifier))) and ((ComponentTypeFilter(Identifier) and ComponentTargetFilter(FilesystemRoot, CpioFilesystem)) or (not ComponentTypeFilter(Identifier) and (ComponentTargetFilter(CpioFilesystem) then ComponentTargetFilter(FilesystemRoot)))))

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/ofrak/gui/server.py", line 95, in wrapper
    return await func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/ofrak/gui/server.py", line 303, in unpack
    result = await resource.unpack()
  File "/usr/local/lib/python3.10/dist-packages/ofrak/resource.py", line 412, in unpack
    return await self.auto_run(all_identifiers=True, all_unpackers=True)
  File "/usr/local/lib/python3.10/dist-packages/ofrak/resource.py", line 385, in auto_run
    components_result = await self._job_service.run_components(
  File "/usr/local/lib/python3.10/dist-packages/ofrak/service/job_service.py", line 262, in run_components
    individual_component_results = await self._auto_run_components(
  File "/usr/local/lib/python3.10/dist-packages/ofrak/service/job_service.py", line 456, in _auto_run_components
    raise component_run_error from ComponentAutoRunFailure(
  File "/usr/local/lib/python3.10/dist-packages/ofrak/service/job_service.py", line 133, in _run_component
    result = await component.run(
  File "/usr/local/lib/python3.10/dist-packages/ofrak/component/abstract.py", line 103, in run
    raise ComponentSubprocessError(e)
ofrak.component.abstract.ComponentSubprocessError: Command '['cpio', '-id']' returned non-zero exit status 2.
Stderr: None.
[...]

PR #276, opened 2023-04-05
Update CPIO unpacker to use 7z utility instead of cpio utility issue …

gskip17:
…#275 in public repo

Reference to issue #275 in master repository.

This change modifies `CpioUnpacker` to instead use the `7z` utility (which is a utility used by other OFRAK components) to extract the CPIO filesystem. 

**Link to Related Issue(s)**

https://github.com/redballoonsecurity/ofrak/issues/275

CPIO unpacker fails on native Linux due to how OFRAK handles process error codes in the `ofrak_core/ofrak/core/cpio.py`

Specifically, during unpacking, when `cpio` utility tries to extract files with absolute paths, the OS will (correctly) raise an error for that file, which is reflected in the return code. 

One potential fix was to add the `--no-absolute-flags` to the `cpio` command invoked by the unpacker - this however means you will lose certain files when you go to repack. 

Instead, by using `7z`, we can unpack CPIO filesystems without a) putting the OS in a position where it will try to overwrite system files, b) work cleanly with root, and c) handle subprocess return codes more cleanly.

### DIFF
--- a/ofrak_core/ofrak/core/cpio.py
+++ b/ofrak_core/ofrak/core/cpio.py
@@ -104,6 +104,7 @@ class CpioUnpacker(Unpacker[None]):
                 *cmd,
             )
             await proc.wait()
+            # Raise when a specific or non-fatal error occurs
             if proc.returncode and proc.returncode != 2:
                 raise CalledProcessError(returncode=proc.returncode, cmd=cmd)
 
--- a/ofrak_core/ofrak/core/cpio.py
+++ b/ofrak_core/ofrak/core/cpio.py
@@ -1,4 +1,5 @@
 import asyncio
+import os
 import logging
 import tempfile
 from dataclasses import dataclass
@@ -100,7 +101,7 @@ class CpioUnpacker(Unpacker[None]):
 
             # use 7z utility to unpack cpio temp file to temp_flush_dir
             cmd = [
-                "7z",
+                "7zz",
                 "x",
                 f"-o{temp_flush_dir}",
                 temp_file_path,
@@ -112,7 +113,7 @@ class CpioUnpacker(Unpacker[None]):
 
             await proc.communicate()
             await proc.wait()
-            if proc.returncode:
+            if proc.returncode and proc.returncode != 2:
                 raise CalledProcessError(returncode=proc.returncode, cmd=cmd)
 
             # before initializing cpio resource, remove the temp cpio file
--- a/ofrak_core/ofrak/core/cpio.py
+++ b/ofrak_core/ofrak/core/cpio.py
@@ -85,23 +85,39 @@ class CpioUnpacker(Unpacker[None]):
     external_dependencies = (CPIO_TOOL,)
 
     async def unpack(self, resource: Resource, config=None):
+
+        resource_data = await resource.get_data()
         cpio_v = await resource.view_as(CpioFilesystem)
-        resource_data = await cpio_v.resource.get_data()
+
         with tempfile.TemporaryDirectory() as temp_flush_dir:
+
+            temp_file_path = os.path.join(temp_flush_dir, "temp_cpio")
+
+            # write cpio data to a temp file in temp_flush_dir
+            with open( temp_file_path, "wb") as t:
+                t.write(resource_data)
+                t.close()
+
+            # use 7z utility to unpack cpio temp file to temp_flush_dir
             cmd = [
-                "cpio",
-                "-id",
+                "7z",
+                "x",
+                f"-o{temp_flush_dir}",
+                temp_file_path,
             ]
+
             proc = await asyncio.create_subprocess_exec(
                 *cmd,
-                stdin=asyncio.subprocess.PIPE,
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
-                cwd=temp_flush_dir,
             )
-            await proc.communicate(input=resource_data)
+
+            await proc.communicate()
+            await proc.wait()
             if proc.returncode:
                 raise CalledProcessError(returncode=proc.returncode, cmd=cmd)
+
+            # before initializing cpio resource, remove the temp cpio file
+            os.remove(temp_file_path)
+
             await cpio_v.initialize_from_disk(temp_flush_dir)
 
 
--- a/ofrak_core/ofrak/core/cpio.py
+++ b/ofrak_core/ofrak/core/cpio.py
@@ -1,5 +1,4 @@
 import asyncio
-import os
 import logging
 import tempfile
 from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


ofrak_core/ofrak/core/cpio.py
```diff
@@ -85,23 +86,39 @@ class CpioUnpacker(Unpacker[None]):
     external_dependencies = (CPIO_TOOL,)
```
rbs-jacob, line 86: This should be changed to also list the 7zip tool as an external dependency.

ofrak_core/ofrak/core/cpio.py
```diff
@@ -85,23 +86,39 @@ class CpioUnpacker(Unpacker[None]):
...
 
     async def unpack(self, resource: Resource, config=None):
+
+        resource_data = await resource.get_data()
         cpio_v = await resource.view_as(CpioFilesystem)
-        resource_data = await cpio_v.resource.get_data()
+
         with tempfile.TemporaryDirectory() as temp_flush_dir:
+
+            temp_file_path = os.path.join(temp_flush_dir, "temp_cpio")
```
rbs-jacob, line 95: Instead of putting the CPIO in the directory we're going to extract to, we should probably use `tempfile.TemporaryFile` to make a new one that is already opened with `'wb'`, and will delete automatically.

ofrak_core/ofrak/core/cpio.py
```diff
@@ -85,23 +86,39 @@ class CpioUnpacker(Unpacker[None]):
...
                 *cmd,
-                stdin=asyncio.subprocess.PIPE,
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
-                cwd=temp_flush_dir,
             )
-            await proc.communicate(input=resource_data)
-            if proc.returncode:
+
+            await proc.communicate()
```
rbs-jacob, line 114:
```suggestion
```

ofrak_core/ofrak/core/cpio.py
```diff
@@ -85,23 +86,39 @@ class CpioUnpacker(Unpacker[None]):
...
+            await proc.wait()
+            if proc.returncode and proc.returncode != 2:
                 raise CalledProcessError(returncode=proc.returncode, cmd=cmd)
+
+            # before initializing cpio resource, remove the temp cpio file
+            os.remove(temp_file_path)
```
rbs-jacob, line 120: See comment above about using a tempfile for this. Should be deleted automatically.

ofrak_core/ofrak/core/filesystem.py
```diff
@@ -352,13 +352,22 @@ async def initialize_from_disk(
...
-                            relative_path,
-                            fh.read(),
-                            file_attributes_stat,
-                            file_attributes_xattr,
-                        )
+                    try:
+                        with open(absolute_path, "rb") as fh:
+                            file_data = fh.read()
+
+                    except Exception as e:
```
rbs-jacob, line 359: We only want to catch the specific exception that happens here, not all exceptions.

ofrak_core/ofrak/core/filesystem.py
```diff
@@ -352,13 +352,22 @@ async def initialize_from_disk(
...
+                        os.chmod(absolute_path, stat.S_IRUSR)
```
rbs-jacob, line 360: Is there a reason we shouldn't just try to automatically add this permission? Right now, we're waiting for there to be an exception before trying to add the permission, but I don't see a benefit to that.

rbs-jacob (changes requested):
Requested some changes. Happy to make them myself to help get this over the line. We can talk offline about it.

rbs-jacob (approved):

ofrak_core/ofrak/core/cpio.py
```diff
@@ -82,26 +84,30 @@ class CpioUnpacker(Unpacker[None]):
...
                 *cmd,
-                stdin=asyncio.subprocess.PIPE,
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
-                cwd=temp_flush_dir,
             )
-            await proc.communicate(input=resource_data)
-            if proc.returncode:
+            await proc.wait()
+            if proc.returncode and proc.returncode != 2:
```
whyitfor, line 107: What does return code 2 mean? Is this something we could document in a comment?

whyitfor (approved):
Looks good, see one comment.

EdwardLarson:
This only changes the Unpacker, and the Packer still uses cpiotool. What happens when a CPIO which cpiotool _can't_ handle is unpacked, then repacked? Does the Packer using cpiotool repack it correctly?

@whyitfor mentioned already, but this needs a test exposing the behavior that cpiotool was failing for. That would also serve to test that repacking with cpiotool works, even for cases it can't unpack.

whyitfor (changes requested):
@marczalik:

- We'll want to address https://github.com/redballoonsecurity/ofrak/pull/276/files#r1162002399
- Also, this change will need a changelog entry.

rbs-jacob (approved):
Left two comments, both of which I have just addressed. Looks good to go to me!

ofrak_core/test_ofrak/components/test_filesystem_component.py
```diff
@@ -397,3 +407,25 @@ def diff_directories(dir_1, dir_2, extra_diff_flags):
             second_type = " ".join(second.split(" ")[4:])
 
             assert first_type == second_type
+
+
+class TestCPIOFilesystem:
+    async def test_unpack(self, cpio_filesystem_root):
```
whyitfor, line 413: Can we have a docstring that descriptes what this tests? (/dev/console) seems to be it.

whyitfor: Also, can we move this test to `test_cpio_component.py` to match all other compressed filesystem tests?

ofrak_core/test_ofrak/components/test_filesystem_component.py
```diff
@@ -397,3 +407,25 @@ def diff_directories(dir_1, dir_2, extra_diff_flags):
...
+        children_names = [child.get_name() for child in children]
+        assert "dev" in children_names
+        dev = await cpio_filesystem_root.get_only_child(
+            r_filter=ResourceFilter(
+                attribute_filters=(ResourceAttributeValueFilter(FilesystemEntry.Name, "dev"),)
+            )
+        )
+        grandchildren = list(await dev.get_children_as_view(FilesystemEntry))
+        grandkid_names = [grandkid.get_name() for grandkid in grandchildren]
+        assert "console" in grandkid_names
```
whyitfor, line 431: could all of this be rewritten with something like:
```suggestion
        filesystem_view = await cpio_filesystem_root.view_as(FilesystemRoot)
        dev_console = await cpio_filesystem.get_entry("dev/console")
        assert dev_console is not None
```

whyitfor (changes requested):

marczalik (approved):

EdwardLarson (approved):
Thanks for updating the test, that addresses all of my feedback

PR #282, opened 2023-04-11
Strings unpacker

EdwardLarson:

Unpack the ASCII strings in program sections as resources.

**Link to Related Issue(s)**
We have some string analyzer and string modifier, but strings are not represented as resources so it's hard to do analysis and queries around them.

An AsciiString view and AsciiStringUnpacker component targeting program sections. Other targets could be added as sensible, but I thought restricting it to sections of executables that get loaded into memory was a reasonably useful constraint.

I used a regex expression to find strings rather than the `strings` utility. Two reasons for this:
1. Using `strings` required flushing to disk, I found this caused linux resource issues when running a big `unpack_recursively` (too many files)
2. Avoid a dependency on `strings` which may not be installed on all systems (important as we step up Windows support). The existing `StringsAnalyzer` could also be changed to use this regex to avoid this dependency in the code base entirely - it actually does not already declare it's dependency on `strings`...

The actual process of searching for strings is pretty simple to expressing in regex, so I don't think we lose confidence by not using `strings`.

@marczalik could maybe help adding tests to this. We should make sure that expected strings are unpacked from a binary. The string length restrictions should also be exercised by this test - i.e. a short string in non-code section should be found, a short string in code section should not be found, and a long string in either one should be found.


ofrak_core/ofrak/core/strings.py
```diff
@@ -77,3 +86,71 @@ async def modify(self, resource: Resource, config: StringFindReplaceConfig) -> N
...
+            return
+        if resource.has_tag(CodeRegion):
+            # code is less likely to have strings so more likely to have false positives
+            min_length = 8
+        else:
+            min_length = 2
+
+        # match sequences of printable characters of at least `min_length` ending with null byte
+        # printable characters defined as: ASCII between ' ' and '~', tab, newline, carriage return
+        pattern = rb"([ -~,\n,\t,\r]{%d,})\x00" % min_length
```
rbs-jacob, line 145: Two comments on this:

- Usually, you don't need commas to separate characters in the character class brackets – this regular expression will match literal commas (see suggested fix below)
- This should probably be compiled with [`re.compile`](https://docs.python.org/3/library/re.html#re.compile) and stored as a class attribute, instead of rebuilding it every time the unpacker is run – this can have significant speed implications if this runs on lots of resources

```suggestion
        pattern = rb"([ -~\n\t\r]{%d,})\x00" % min_length
```

ofrak_core/test_ofrak/components/test_string.py
```diff
@@ -1,16 +1,58 @@
...
+__asm__(".global shortString\n\t"
+    ".type shortString, @function\n\t"
+    "push %r15\n\t"
+    "and 0, %r15\n\t"
+);
+
+int main() {
+    printf("O");
+    printf("h, hi");
+    printf(" Mark!\n");
```
rbs-jacob, line 50: Suggest changing to "Marc" in honor of our very own Marc Zalik!

```suggestion
    printf(" Marc!\n");
```

Note: will also need to be changed in the actual tests below.

ofrak_core/test_ofrak/components/test_string.py
```diff
@@ -1,16 +1,58 @@
...
+    "push %r15\n\t"
+    "and 0, %r15\n\t"
+);
+
+// Generate bytes that look like a short ascii string (7 bytes)
+__asm__(".global shortString\n\t"
+    ".type shortString, @function\n\t"
+    "push %r15\n\t"
+    "and 0, %r15\n\t"
+);
```
rbs-jacob, line 45: The implication is that these shouldn't match as strings, but will, right?

Might be worth clarifying that in the comments?

rbs-jacob (changes requested):
Some very minor suggestions. 

For the record, we're assuming that the user has `strings` installed for the [`StringsAnalyzer`](https://github.com/redballoonsecurity/ofrak/blob/cbbdca26498328656b2d6ca9b5d2275df3f273fa/ofrak_core/ofrak/core/strings_analyzer.py#L38-L43), but your argument about speed is well-taken.

ofrak_core/test_ofrak/components/test_string.py
EdwardLarson: The only thing stopping these being matched as a string should be the min length requirement (8) for strings intermixed with code.

ofrak_core/ofrak/core/strings.py
EdwardLarson: Yeah compiling is a good idea. I suppose we'll need two compiled versions for the two possible `min_length` values.

rbs-jacob: I wonder if it makes sense to have a class attribute that caches compiled versions. Then, it compiles on the first run of `unpack`, but has the regular expression cached for later. Or maybe, a method with `@lru_cache` decorator to return a compiled regular expression given the minimum number of characters?

Either way, instead of having two versions, I like the idea of generalizing. That way, if we want to make an optional component config for this to let the user control the minimum size, the code will not need to be refactored.

EdwardLarson: I don't think some sort of caching is called for, if we want to add an optional config for minimum length we can just compile the appropriate pattern on the fly when a config is given.

rbs-jacob (approved):
Looks good to me! Thanks for making those changes

PR #284, opened 2023-04-12
Add big-endian support in the patchmaker. Also add ARM BE8 support

paulnoalhyt:

Add support for compiling/assembling big-endian code with the patchmaker. Same for ARM BE8.

**Link to Related Issue(s)**

- Added big-endian flags in the GNU toochain for the compiler, linker and assembler.
- Added a `GENERIC_ARM_BE8` ProcessorType
- Added the corresponding BE8 flags to the compiler and linker

@whyitfor could maybe help adding tests to this.


alchzh:
Two questions:
- Should we process the endian flag on an architecture by architecture basis instead of in abstract, since it is architecture dependent? I believe that PowerPC defaults to big endian and should add `-mlittle-endian` if little endian is specified? For other architectures, I think that we should probably raise an error if an unsupported endianness is used.
- Should the flag be added to llvm as well?

ofrak_patch_maker/ofrak_patch_maker/toolchain/gnu.py
```diff
@@ -46,7 +46,6 @@ def __init__(
                 #      if sections contain more than one function =/
                 "-fno-merge-constants",  # avoids sections like .rodata.cst16, .rodata.str1.1 etc
                 "-fno-reorder-functions",
-                "-Wall",
```
alchzh, line 49: What is the reason for this change? I don't see any additional warnings generated by the big endian test cases you added. If there's a specific spurious warning you ran into consistently, it should be disabled by a conditional `-Wno` in `gnu_arm.py`

alchzh (dismissed):

ofrak_patch_maker/ofrak_patch_maker/toolchain/gnu.py
paulnoalhyt: I just removed one because the `-Wall` option was present twice in this list

alchzh (approved):

PR #287, opened 2023-04-14
Run any OFRAK Component in the GUI

dannyp303:

Added a new view to the OFRAK GUI to run any component on a resource. 

**Link to Related Issue(s)**

Added 4 new routes to the server to get component information and 3 new svelte components to select components and provide the required input to them.
New Routes:
- `get_components`: provides a list of all components, with filtering options.
- `get_config_for_component`: Gets the config and its fields for the component selected.
- `get_tags_and_num_components`: Gets the tags associated with the selected resource and the number of components targeting that tag.
- `run_component`: Provides config information to the server and runs the component on the selected resource.

New Svelte Components:
- `ComponentsView.svelte`: Provides the user a dropdown of all components available with filtering options.
- `ComponentConfig.svelte`: Provides the user a form to provide information required for the ComponentConfig.
- `ComponentConfigNode.svelte`: A single input form for a parameter of the ComponentConfig. Recursively runs through each parameter. 

@rbs-jacob the svelte master.


rbs-jacob (changes requested):
I will review this shortly. In the meantime, can you merge `master` in and resolve the merge conflicts? That way we can run the tests.

Thanks!

frontend/src/stores.js
```diff
@@ -11,4 +11,4 @@ export const script = writable([]);
  * Points the OFRAK frontend to a seperate backend server
  * When empty, uses the same host and port as the frontend
  */
-export const backendUrl = "";
+export const backendUrl = "//localhost:1666";
```
rbs-jacob, line 14: This is probably not supposed to be committed here...

```suggestion
export const backendUrl = "";
```

ofrak_core/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Generate dynamic, runnable script based on GUI actions and display the script in the GUI. ([#265](https://github.com/redballoonsecurity/ofrak/pull/265))
 - Add `-f`/`--file` option to `ofrak gui` command to pre-load some files into OFRAK before opening the GUI, so they can be explored right away ([#266](https://github.com/redballoonsecurity/ofrak/pull/266))
 - Add `-i`/`--import` option to the CLI to import and discover additional OFRAK Python packages when starting OFRAK. [#269](https://github.com/redballoonsecurity/ofrak/pull/269)
+- Add a new view to the OFRAK GUI to run any component on a resource. [#287](https://github.com/redballoonsecurity/ofrak/pull/287)
```
rbs-jacob, line 11:
```suggestion
- Let the OFRAK GUI run any component [#287](https://github.com/redballoonsecurity/ofrak/pull/287)
```

disassemblers/ofrak_capstone/ofrak_capstone/components.py
```diff
@@ -143,7 +144,7 @@ def __init__(
         super().__init__(resource_factory, data_service, resource_service)
         self._disassembler_service = disassembler_service
 
-    async def analyze(self, resource: Resource, config: None) -> RegisterUsage:
+    async def analyze(self, resource: Resource, config: ComponentConfig = None) -> RegisterUsage:
```
rbs-jacob, line 147: This is an incorrect type change made throughout this PR. If you must delete the type variable, it should be the following. But setting the variable type to `ComponentConfig` when the value is `None` is definitely not correct.

```suggestion
    async def analyze(self, resource: Resource, config: Optional[ComponentConfig] = None) -> RegisterUsage:
```

Why was it necessary to get rid of the type variable in the first place? The type variable allows the resource method to have the same type as the component, which is a generic subclass of `AbstractComponent` that uses that type variable.

I'm surprised mypy did not complain about this, but maybe we'll see it when the tests get run in CI?

ofrak_core/ofrak/gui/server.py
```diff
@@ -156,7 +170,16 @@ def __init__(
...
+                web.post(
+                    "/{resource_id}/get_components",
+                    self.get_components,
+                ),
+                web.get("/{resource_id}/get_config_for_component", self.get_config_for_component),
+                web.post("/{resource_id}/run_component", self.run_component),
                 web.get("/", self.get_static_files),
+                web.post(
+                    "/{resource_id}/get_tags_and_num_components", self.get_tags_and_num_components
+                ),
```
rbs-jacob, line 182: The static routes must be last. This should be moved above the preceding line.

ofrak_core/requirements.txt
```diff
@@ -6,7 +6,6 @@ importlib-metadata>=1.4
 intervaltree==3.1.0
 keystone-engine==0.9.2
 lief==0.12.3
-orjson~=3.8.7
```
rbs-jacob, line 9: Is removing the `orjson` dependency supposed to be part of this PR?

If anything, we should try and import it, and fall back on the regular `json` library if we can't. But I don't think removing it entirely makes sense.

frontend/src/App.svelte
```diff
@@ -109,9 +113,13 @@
...
   function handleShortcut(e) {
     // Don't handle keypresses from within text inputs.
+    // Disable shortcuts in views with text inputs, otherwise misclicking outside of a text area may
+    // cause users to accidentally run shortcuts.
     if (
       ["input", "textarea"].includes(e.target?.tagName.toLocaleLowerCase()) ||
-      e.target.isContentEditable
+      e.target.isContentEditable ||
+      modifierView === ComponentConfig ||
+      modifierView === FindReplaceView
```
rbs-jacob, line 122: Probably makes sense for this to just check that the `modifierView` variable is set at all.

```suggestion
      modifierView
```

frontend/src/ResourceTreeNode.svelte
```diff
@@ -76,6 +76,18 @@
     color: var(--main-bg-color);
   }
 
+  .modified {
+    text-decoration-line: underline;
+    text-decoration-color: #dc4e47;
```
rbs-jacob, line 81: We want to avoid hard-coding CSS colors inside components if possible. This makes them less portable, and means we can't change the colors programmatically. This should probably be a CSS variable set at the root in `global.css`.

frontend/src/ResourceTreeNode.svelte
```diff
@@ -129,9 +144,17 @@
       resourceNodeDataMap[self_id].commentsPromise =
         rootResource.get_comments();
     }
+    if (resourceNodeDataMap[self_id].modified === undefined) {
+      resourceNodeDataMap[self_id].modified = false;
+    }
+    if (resourceNodeDataMap[self_id].prevModified === undefined) {
+      resourceNodeDataMap[self_id].prevModified = false;
+    }
```
rbs-jacob, line 152: The names `modified` and `prevModified` are not clear. What is the difference? Presumably anything modified has been previously modified?

frontend/src/App.svelte
```diff
@@ -66,7 +69,8 @@
     currentResource,
     rootResource,
     modifierView,
-    bottomLeftPane;
+    bottomLeftPane,
+    selectedComponent;
```
rbs-jacob, line 73: Instead of having `selectedComponent` be a variable of `App.svelte`, it should probably be a variable of `ComponentsView.svelte`. Then `ComponentsView` can wrap the other components that make up the view, rather than having `modifierView` be set with one of the `ComponentConfig` views. 

Let me know if this is unclear, and I can elaborate.

rbs-jacob (changes requested):
Did a very quick first pass. I didn't look too closely at much of the core of the PR yet, but already have some comments.

I'll do some more review passes, but in the meantime, here is some stuff that needs fixing.

Also, getting the tests running will help move this PR along significantly!

disassemblers/ofrak_capstone/ofrak_capstone/components.py
whyitfor: Probably best is just:
```python
    async def analyze(self, resource: Resource, config=None) -> RegisterUsage:
```

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,199 @@
...
+        >
+          <option value="{null}">{node["enum"][0]}</option>
+          {#each Object.entries(node["enum"]) as [name, value]}
+            <option value="{node['type']}.{name}">
+              {name}
+            </option>
+          {/each}
+        </select>
+      </form>
+    {:else if node["fields"] != null}
```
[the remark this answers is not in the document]
dannyp303: dataclasses have fields, I can add a comment, but idk of a way to make that less confusing.

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,199 @@
...
+        <p>Value</p>
+        <svelte:self node="{node['args'][1]}" bind:element="{dictValue}" />
+      </li>
+      {#each element as elements}
+        <li>{elements}</li>
+      {/each}
+    {:else if node["type"] == "typing.Union"}
+      <p>Select Type</p>
+      {#each node["args"] as type}
+        <button
```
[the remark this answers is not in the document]
dannyp303: for sure

frontend/src/App.svelte
dannyp303: This makes sense, implemented

frontend/src/ComponentConfig.svelte
```diff
@@ -0,0 +1,174 @@
...
+          const results = await $selectedResource.run_component(
+            selectedComponent,
+            ofrakConfig['type'],
+            config
+          );
+          resourceNodeDataMap[$selected] = {
+            collapsed: false,
+            childrenPromise: $selectedResource.get_children(),
+          };
+          for (const result in results) {
```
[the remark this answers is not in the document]
dannyp303: There is already a helper function that does this in `remote_resource.js` called `ingest_component_results` we should maybe move the highlighting there.

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,199 @@
...
+      </li>
+    {:else if node["type"] == "builtins.int"}
+      <li>
+        <input type="number" bind:value="{element}" />
+      </li>
+    {:else if node["type"] == "typing.List"}
+      <li>
+        {#each node["args"] as arg}
+          <svelte:self node="{arg}" bind:element="{listElement}" />
+          <button on:click="{addElementToArray}">Add Element</button>
```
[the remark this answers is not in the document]
dannyp303: This is a good idea but not super straight forward to implement for reasons such as:
- displaying the value of the element when the value is another list or other larger structure
- we are not currently tracking items in objects, just kind of plopping them in there
- im sure others...
I think this will be easier to implement after I attack the CSS with @rbs-jacob and find a nice way of displaying the current config inside the view.

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,199 @@
...
+        {/each}
+      </li>
+      {#each element as elements}
+        <li>{elements}</li>
+      {/each}
+    {:else if node["type"] == "typing.Tuple"}
+      <li>
+        {#each node["args"] as arg, i}
```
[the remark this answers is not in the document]
dannyp303: Implemented server side, check out `_modify_elipsis` in `gui/server.py`

frontend/src/ComponentConfig.svelte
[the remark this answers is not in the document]
EdwardLarson: I think we talked about this the other day, always highlighting modified resources is a good idea.

frontend/src/ComponentConfigNode.svelte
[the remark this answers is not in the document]
EdwardLarson: @dannyp303 is this more doable now or still complicated?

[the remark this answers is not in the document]
dannyp303: oh sorry didnt comment, but its done
Screen Shot 2023-05-04 at 2 31 50 PM

frontend/public/global.css
```diff
@@ -3,6 +3,7 @@
   --main-fg-color: white;
   --selected-bg-color: #86e3ed;
   --comment-color: #eb8e5b;
+  --highlight-color: #dc4e47;
```
rbs-jacob, line 6: Why is this highlight color added here when it is overridden by the original two lines below?

frontend/src/App.svelte
```diff
@@ -109,9 +111,14 @@
...
     // Don't handle keypresses from within text inputs.
+    // Disable shortcuts in views with text inputs, otherwise misclicking outside of a text area may
+    // cause users to accidentally run shortcuts.
     if (
-      ["input", "textarea"].includes(e.target?.tagName.toLocaleLowerCase()) ||
-      e.target.isContentEditable
+      (["input", "textarea"].includes(e.target?.tagName.toLocaleLowerCase()) ||
+        e.target.isContentEditable ||
+        modifierView === ComponentsView ||
+        modifierView === FindReplaceView) &&
```
rbs-jacob, line 120: Shortcuts should probably be disabled if any modifier view is set.

```suggestion
        modifierView) &&
```

The linter will probably fail after this suggestion is applied.

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,360 @@
...
+    return element;
+  }
+
+  const getFileContents = async () => {
+    await slurpSourceBundle(files);
+  };
+
+  $: if (files) {
+    getFileContents(files);
+  }
```
rbs-jacob, line 176:
```suggestion

  $: if (files) {
    slurpSourceBundle(files);
  }
```

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,360 @@
...
+
+  // $: if (node["type"] == "builtins.int") {
+  //   try {
+  //     element = calculator.calculate(_element);
+  //     intInput?.setCustomValidity("");
+  //   } catch {
+  //     element = undefined;
+  //     intInput?.setCustomValidity("Invalid expression.");
+  //   }
+  // }
```
rbs-jacob, line 160: Why is this commented it out? If it's causing a problem, we should probably address it a different way.

frontend/src/ComponentConfigNode.svelte
```diff
@@ -0,0 +1,360 @@
...
+
+  async function slurpSourceBundle(files) {
+    for (const file of files) {
+      let text = await file.text();
+      element = [...element, [file.name, text]];
+    }
+    return element;
+  }
```
rbs-jacob, line 168:
```suggestion
  async function slurpSourceBundle(files) {
    element = [];
    for (const file of files) {
      let text = await file.text();
      element.push([file.name, text]);
    }
    return element;
  }
```

frontend/src/ComponentsView.svelte
```diff
@@ -0,0 +1,336 @@
...
+  import Checkbox from "./Checkbox.svelte";
+
+  export let modifierView, resourceNodeDataMap;
+  let errorMessage,
+    all_components = true,
+    incl_analyzers = false,
+    incl_modifiers = false,
+    incl_packers = false,
+    incl_unpackers = false,
+    target_filter = null,
```
rbs-jacob, line 140: JavaScript style is camelCase instead of Python's snake_case. It's kind of annoying to switch, and I'm not going to be a stickler about letting that block the PR. Just a heads up for the future.

frontend/src/FileBrowser.svelte
```diff
@@ -68,7 +68,7 @@
   {#if !dragging}
     <label class="filelabel">
       <slot />
-      <input type="file" bind:this="{input}" bind:files="{files}" />
+      <input type="file" multiple bind:this="{input}" bind:files="{files}" />
```
rbs-jacob, line 71: I don't think we always want uses of the `FileBrowser` component to have multiple file selection enabled. Probably best to do this as a boolean component parameter. Then `multiple` can be true or false based on that variable.

rbs-jacob (changes requested):

frontend/src/App.svelte
dannyp303: Given that, I think the entire conditional can be simplified to `if (modifierView)`...

rbs-jacob: I disagree, I think we still need to check for whether we're in `input`, `textarea`, or content editable elements because it is possible for those to exist outside of modifier view panes. For example, we have the jump to offset input that exists in the hex view.

frontend/src/ComponentConfigNode.svelte
dannyp303: I'm not sure of the root reason, but when clearing an element of a list or dict with an int value, the value that was being cleared would propagate to the the new nth element in the list/dict.

For instance if i try to remove element 2 of [(n1, k1), (n2,k2), (n3,k3)] where n is an int it will become  [(n1, k1), (n2,k3)].

I think it has something to do with the reactivity of the calculator line, but was kind of difficult to figure out.

rbs-jacob (approved):

PR #296, opened 2023-04-28
Bump importlib-metadata version

rbs-jacob:

Running `pip install ofrak` installs version `3.0.0` and, for some reason, seems to install the lowest possible version of `importlib-metadata` (1.4.0). This old version does not actually support all of the features of the library we are using. This PR bumps the version.

@whyitfor


whyitfor:
@rbs-jacob, shouldn't we have a bugfix in the changelog? This could end up being `3.0.1`.

whyitfor:
> I wasn't sure about updating the changelog, because I'm not sure if this qualifies as a `3.0.0` post release, or whether it's a minor version bump of its own. Sounds like you want me to update it, though.

@rbs-jacob, this seems like something that should go in "Fixed" in a change log. It isn't a minor version but a patch version bum, so it would be released as `3.0.1`. Please add a "Fixed" entry in the change log.

whyitfor (approved):

PR #299, opened 2023-05-01
Add dependencies for tools which were missing them

EdwardLarson:

Mark dependencies on libmagic and `strings` utility.

**Link to Related Issue(s)**
These are tools which OFRAK depends on. On some platforms, like Mac, they are not necessarily installed. They should have been explicitly marked as dependencies so the `ofrak deps` command can find and install them when necessary.


ofrak_core/ofrak/core/magic.py
```diff
@@ -22,18 +21,47 @@ class Magic(ResourceAttributes):
...
+        import magic
+    except ImportError:
+        magic = None
+
+    def __init__(self):
+        super().__init__(
+            "libmagic",
+            "https://linux.die.net/man/3/libmagic",
+            install_check_arg="",
+            brew_package="libmagic",
```
rbs-jacob, line 35: Should this also have `apt_package="libmagic1"`?

ANogin: I think the proper URL is https://www.darwinsys.com/file/. And yes, `apt_package="libmagic1"`

rbs-jacob (approved):

ofrak_patch_maker/ofrak_patch_maker/toolchain/utils.py
```diff
@@ -5,15 +5,19 @@
...
+_magic = None
+
 
 def get_file_format(path):
-    result = magic.from_file(path).split(" ")[0].lower()
+    global _magic
+    if _magic is None:
+        import magic as _magic
+
+    result = _magic.from_file(path).split(" ")[0].lower()
```
whyitfor, line 20: Why is this change needed? I don't understand what is different.

ofrak_core/ofrak/core/magic.py
```diff
@@ -22,19 +22,54 @@ class Magic(ResourceAttributes):
     descriptor: str
 
 
+class _LibmagicDependency(ComponentExternalTool):
+    try:
+        import magic
+
+        magic_: ModuleType = magic
+    except ImportError:
+        magic_ = None
```
whyitfor, line 31: Instead of duplicating this try-catch logic, why not have a global `MAGIC_INSTALLED` variable like we are already doing with [binwalk](https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/ofrak/core/binwalk.py)?

ofrak_patch_maker/ofrak_patch_maker/toolchain/utils.py
EdwardLarson: If it is left as a global import, it doesn't matter if a user (without libmagic installed) wants to exclude components missing dependencies because a hard error will be raised at import time when PatchMaker code is imported. The expected behavior is that as a user without an external tool dependency installed, I should still be able to launch and use OFRAK without needing to install all those tools, working instead with a limited set of components.

Since we only use magic in this one place, I thought it acceptable to sort of localize the magic import.

ofrak_core/ofrak/core/magic.py
whyitfor: @EdwardLarson

ofrak_patch_maker/ofrak_patch_maker/toolchain/utils.py
whyitfor, line 20: I see. What about deleting lines 12-13, and something like this:

```suggestion
    try:
        import magic
    except ImportError:
        # ImportError is likely raise because libmagic cannot be found on the system. See error message.
        raise
    result = magic.from_file(path).split(" ")[0].lower()
```

ofrak_patch_maker/ofrak_patch_maker/toolchain/utils.py
```diff
@@ -5,15 +5,19 @@
...
 from typing import Optional, Dict, Mapping, Tuple
 
-import magic
-
 from ofrak_patch_maker.toolchain.model import BinFileType, Segment
 from ofrak_type.error import NotFoundError
 from ofrak_type.memory_permissions import MemoryPermissions
 
+_magic = None
+
```
whyitfor, line 13: In favor of deleting this in conjunction with https://github.com/redballoonsecurity/ofrak/pull/299/files#r1194275565.:

```suggestion
```

ofrak_core/ofrak/core/magic.py
```diff
@@ -22,19 +28,47 @@ class Magic(ResourceAttributes):
...
     outputs = (Magic,)
+    external_dependencies = (LIBMAGIC_DEP,)
 
     async def analyze(self, resource: Resource, config=None) -> Magic:
         data = await resource.get_data()
-        magic_mime = magic.from_buffer(data, mime=True)
-        magic_description = magic.from_buffer(data)
-        return Magic(magic_mime, magic_description)
+        if not MAGIC_INSTALLED:
+            raise ImportError("libmagic does not seem to be installed!")
```
whyitfor, line 67: Would it be better to check for (and handle) this error inside of [AbstractResource.run](https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/ofrak/component/abstract.py#L100), where other `ComponentMissingDependencyError`s are raised?

whyitfor (commented):
Looking good! See one comment about raising a `ComponentMissingDependencyError`.

whyitfor (approved):

whyitfor:
@EdwardLarson, can you add a "fixed" changelog to OFRAK core for the Magic issue? Otherwise I think this looks good!

PR #304, opened 2023-05-10
better check for THUMB mode, and try CB addr as Thumb if not found in…

EdwardLarson:
…itially

**Link to Related Issue(s)**
For some Thumb binaries there were 2 errors:
1. The complex block could not be found by address. The problem was that the address that angr had for the complex block was the "thumbified" address, with LSB set to zero, while the address we stored in OFRAK for the complex block was the "de-thumbified" address with LSB cleared. It is this "de-thumbified" address that the AngrComplexBlockUnpacker was looking for and could not find.
2. Basic blocks were not correctly identified as being in Thumb mode because the architecture that angr analyzed for the binary did not include "thumb" in the name. So the capstone disassembly was nonsense, disassembling Thumb instructions as ARM.

Fix the first issue by checking for a complex block with the address's LSB set to 1 if the complex block could not be found ith LSB set to 0.

Fix the second issue by checking the address of the basic block that angr gives us is "thumbified" with LSB set 1, and use that to determine if the basic block is Thumb, rather than the name of the global architecture.

Add a new PIE Thumb test case for all disassembler backends.

Fix another ofrak_angr issue with data word extraction - the CB bounds could be over-eager and include data words which would be erroneously unpacked, so now data words are only extracted if they are referenced by basic blocks which have actually been unpacked from the CB.


disassemblers/ofrak_angr/ofrak_angr/components/blocks/unpackers.py
```diff
@@ -125,6 +125,10 @@ def _angr_get_basic_blocks(
         LOGGER.debug(f"Getting basic blocks from function {cb_vaddr_range.start:#x}")
 
         angr_complex_block = angr_analysis.project.kb.functions.function(addr=cb_vaddr_range.start)
+        if not angr_complex_block:
+            thumb_cb_vaddr = cb_vaddr_range.start | 0x1
+            angr_complex_block = angr_analysis.project.kb.functions.function(addr=thumb_cb_vaddr)
```
whyitfor, line 130: Shouldn't this attempt only be made if the architecture is ARM?

whyitfor:
@EdwardLarson, this looks like it could use an angr chaangelog entry?

disassemblers/ofrak_angr/ofrak_angr/components/blocks/unpackers.py
```diff
@@ -226,3 +238,43 @@ def _angr_get_dword_blocks(
...
+    fallthrough_vaddr = get_real_address_if_arm(
+        angr_analysis.project.arch, angr_cb_basic_blocks[current_bb_idx + 1].addr
+    )
+    successors_generator = angr_complex_block.graph.succ.get(current_angr_bb.codenode)
+
+    if successors_generator is None:
+        LOGGER.warning(
+            f"angr did not find any successors for BB 0x{current_angr_bb.addr:x}, but since it has a BB after it, assuming that it still can fallthrough to the next BB."
+        )
+        return True, fallthrough_vaddr
```
whyitfor, line 270:
```suggestion
    successor_vaddrs = [
        get_real_address_if_arm(angr_analysis.project.arch, succ_codenode.addr)
        for succ_codenode, edge_info in
        angr_complex_block.graph.succ.get(current_angr_bb.codenode).items()
    ]
    fallthrough_vaddr = get_real_address_if_arm(
        angr_analysis.project.arch, angr_cb_basic_blocks[current_bb_idx + 1].addr
    )
    # Check if fallthrough_vaddr is in list of successor. 
    # If the basic block ends in an unconditional branch, 
    # the fallthrough vaddr will not be the exit_addr.
    # In this case, use the first successor address
    exit_addr = fallthrough_vaddr if fallthrough_vaddr in successor_vaddrs else successor_vaddrs[0]
    return False, exit_addr
```

whyitfor: I tried making the intention here clearer

whyitfor (changes requested):
Looks good! I made a couple comments with an eye towards making the code clearer for future readers.

disassemblers/ofrak_angr/ofrak_angr/components/blocks/unpackers.py
```diff
@@ -226,3 +238,43 @@ def _angr_get_dword_blocks(
...
+
+    successor_vaddrs = [
+        get_real_address_if_arm(angr_analysis.project.arch, succ_codenode.addr)
+        for succ_codenode, edge_info in successors_generator.items()
+    ]
+
+    if fallthrough_vaddr in successor_vaddrs:
+        return False, fallthrough_vaddr
+    else:
+        return False, successor_vaddrs[0]
```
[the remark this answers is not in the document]
EdwardLarson: If the fallthrough (which is the next basic block) isn't a successor, it means that the basic block can't actually exit to the next basic block. For example, the last instruction is an unconditional branch (and not a linking branch, i.e. a call that returns to the same place). So it just checks if the fallthrough address is actually a successor, if it is, then it should always be the successor. Otherwise, it chooses the first successor, which is a bit arbitrary, but is probably the only successor (in the unconditional branch case) and if it isn't, then the control flow is "weird" (possibly jump table).

Actually, is that the expected behavior when the BB doesn't fall through?

disassemblers/ofrak_angr/ofrak_angr/components/blocks/unpackers.py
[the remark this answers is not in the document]
EdwardLarson: It's not that the basic block isn't a successor (though the first block is often not a successor) but that it doesn't seem to have any successors. Honestly this is as much for catching errors in angr or problems with how we are getting the successors, and handling them somewhat elegantly.

whyitfor (approved):

PR #309, opened 2023-05-15
Add a settings pane to the GUI with the ability to change colors

rbs-jacob:

Add a settings pane to the GUI with the ability to change colors.

Add a `settings` store that can be imported from `stores.js`. This object stores all of the settings values. If exposed, these values can be edited by a user via the Settings pane, but it is possible to have stored settings with no corresponding UI for the user to edit them.

Update the minimap views and hex view to automatically change colors when the settings change.

The only outstanding problem is that the back end can be changed in the dev settings, and if incorrect, the user won't be able to reach the settings menu to change the value back. This can be fixed by clearing the local storage via the developer tools. Eventually, we should make the settings menu accessible from the "start view" before a resource exists, but for now, I don't believe it should block the PR. 

@EdwardLarson @dannyp303 @whyitfor


whyitfor:
@rbs-jacob , can we add an entry in the changelog? Perhaps a quick Docs page as well explaining how to use the feature?

whyitfor (commented):
LGTM! See comments about documenting these changes!

whyitfor (approved):

PR #310, opened 2023-05-19
Feature/load and run script

EdwardLarson:

Add a Modifier and associated GUI button/view to upload a Python file and run a function in that file on the selected Resource

**Link to Related Issue(s)**

TODO:
[ ] Write a test for the Modifier
[ ] Custom icon for the "Run Script" button
[ ] Clean up the CSS to make the "Run Script" view tolerable


frontend/src/ResourceTreeToolbar.svelte
```diff
@@ -240,7 +241,13 @@
...
         },
       },
-
+      {
+        text: "Run Script",
+        iconUrl: "/icons/settings.svg",
+        onclick: async (e) => {
+          modifierView = RunScriptView;
+        },
+      },
```
rbs-jacob, line 250:
```suggestion

      {
        text: "Run Script",
        iconUrl: "/icons/settings.svg",
        onclick: async (e) => {
          modifierView = RunScriptView;
        },
      },
      
```

rbs-jacob (changes requested):
Minor spacing fix

EdwardLarson:
@whyitfor What do you think of gating this behind the "Experimental Features" checkbox? We were talking about it at standup today, and all of us could go either way.

On the one hand, in some ways it is simpler than running any component, since we only need one simple config to be properly serialized/deserialized. On the other hand, there could be a lot of "weird" side effects running scripts.

marczalik:
> @whyitfor What do you think of gating this behind the "Experimental Features" checkbox? We were talking about it at standup today, and all of us could go either way.
> 
> On the one hand, in some ways it is simpler than running any component, since we only need one simple config to be properly serialized/deserialized. On the other hand, there could be a lot of "weird" side effects running scripts.

@EdwardLarson As mentioned offline, if we know that the UI/UX will change significantly for this feature that may be a good reason to set it behind "Experimental Features" for now. Regarding the fact that it's not an API-level change, I still think it's a good idea to inform users ahead of time if we know that something is being shipped "unfinished" and that seems like what "Experimental Features" should be.

whyitfor:
@EdwardLarson, if we are on the fence then let's put it behind the "experimental" config.

PR #316, opened 2023-05-25
Fix AARCH64 Keystone Vector Instructions bug

rbs-alexr:

When capstone disassembles the bytes for the following instruction "movi v0.2d #0" it represents it as "movi v0.0x2d #0" which causes keystone to throw an "Unknown Operand" error. 

**Link to Related Issue(s)**
N/A

I added some processing to the assembler_service_keystone to work around this capstone bug

@whyitfor

### DIFF
--- a/disassemblers/ofrak_capstone/ofrak_capstone/disassembler_service_capstone.py
+++ b/disassemblers/ofrak_capstone/ofrak_capstone/disassembler_service_capstone.py
@@ -213,7 +213,7 @@ def _asm_fixups(base_mnemonic: str, base_operands: str, isa: InstructionSet) -> 
     """
     In AARCH64, vector instructions use numerical constants to reference vectors to perform operations on
     These numerical constants shouldn't be converted to hex because keystone will be unable to reassemble them
-    In every other case we want to to the constants to hex conversion
+    In every other case we want to do the hex conversion
     """
     hex_operand_fixup = lambda base_operands: re.sub(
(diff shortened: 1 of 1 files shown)


whyitfor (changes requested):
@rbs-alexr, can you add a tests case that fails before this change is made?

rbs-alexr:
To provide some context to the most recent commits. The erroneous behavior we were observing was not due to keystone, it was due to our own _asm_fixups function in disassembler_service_capstone. We had regex to convert constants to their hex values, which breaks for vector instructions. The new code skips this conversion if the mnemonic is from a list of known AARCH64 vector instructions

disassemblers/ofrak_capstone/ofrak_capstone/disassembler_service_capstone.py
```diff
@@ -199,7 +199,35 @@ def _get_cs_mode_flag(cs_disam: CapstoneDisassemblerType):
...
+        "fmls",
+        "fmov",
+        "fmul",
+        "fmulx",
+        "fneg",
+    ]
+    """
+    In AARCH64, vector instructions use numerical constants to reference vectors to perform operations on
+    These numerical constants shouldn't be converted to hex because keystone will be unable to reassemble them
+    In every other case we want to to the constants to hex conversion
```
EdwardLarson, line 216:
```suggestion
    In every other case we want to do the hex conversion
```

PR #321, opened 2023-06-01
Highlight fix

dannyp303:

Small update to highlighting.

**Link to Related Issue(s)**

### DIFF
--- a/frontend/public/global.css
+++ b/frontend/public/global.css
@@ -3,8 +3,8 @@
   --main-fg-color: white;
   --selected-bg-color: #86e3ed;
   --comment-color: #eb8e5b;
-  --last-modified-color: #dc4e47;
-  --all-modified-color: #fdb44e;
+  --last-modified-color: #df463e;
+  --all-modified-color: #fff259;
   --font: "cartograph";
   --highlight-color: #fff259; /* Snake color */
   --accent-text-color: #a9ed96; /* Caterpillar color */
--- a/frontend/public/global.css
+++ b/frontend/public/global.css
@@ -3,7 +3,8 @@
   --main-fg-color: white;
   --selected-bg-color: #86e3ed;
   --comment-color: #eb8e5b;
-  --modified-color: #dc4e47;
+  --last-modified-color: #dc4e47;
+  --all-modified-color: #fdb44e;
   --font: "cartograph";
   --highlight-color: #fff259; /* Snake color */
   --accent-text-color: #a9ed96; /* Caterpillar color */
--- a/frontend/src/BinaryModifyView.svelte
+++ b/frontend/src/BinaryModifyView.svelte
@@ -181,7 +181,9 @@
         await selectedResource.queue_patch(patchData, startOffset, endOffset);
       }
 
-      resourceNodeDataMap[$selected].modified = true;
+      resourceNodeDataMap[$selected] = {
+        lastModified: true,
+      };
       modifierView = undefined;
       refreshResource();
     } catch (err) {
--- a/frontend/src/BinaryModifyView.svelte
+++ b/frontend/src/BinaryModifyView.svelte
@@ -180,10 +180,10 @@
       if (selectedResource) {
         await selectedResource.queue_patch(patchData, startOffset, endOffset);
       }
-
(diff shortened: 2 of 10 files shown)


frontend/public/global.css
```diff
@@ -3,7 +3,8 @@
   --main-fg-color: white;
   --selected-bg-color: #86e3ed;
   --comment-color: #eb8e5b;
-  --modified-color: #dc4e47;
+  --last-modified-color: #dc4e47;
+  --all-modified-color: #fdb44e;
```
rbs-jacob, line 7: These colors also should be set in `frontend/src/stores.js` in the default settings object. https://github.com/redballoonsecurity/ofrak/blob/eb0c5d6d2a3942abc11054f2b46f3a99757b37bd/frontend/src/stores.js#L15-L30

The corresponding setting should have a GUI element added in the `frontend/src/SettingsView.svelte` file. https://github.com/redballoonsecurity/ofrak/blob/eb0c5d6d2a3942abc11054f2b46f3a99757b37bd/frontend/src/SettingsView.svelte#L243-L246

And the color should be used in `frontend/src/App.svelte`. https://github.com/redballoonsecurity/ofrak/blob/eb0c5d6d2a3942abc11054f2b46f3a99757b37bd/frontend/src/App.svelte#L160-L165

As part of this change, it's worth adding a comment to this CSS that points a developer to these three places.

frontend/src/BinaryModifyView.svelte
```diff
@@ -181,7 +181,9 @@
         await selectedResource.queue_patch(patchData, startOffset, endOffset);
       }
 
-      resourceNodeDataMap[$selected].modified = true;
+      resourceNodeDataMap[$selected] = {
+        lastModified: true,
+      };
```
rbs-jacob, line 186: Because of the reactive block linked below, if all of the values in the object are set to `undefined` (which they are here, except for `lastModified`), the values will be reverted to their defaults. This will, for example, collapse objects that may have been collapsed before. https://github.com/redballoonsecurity/ofrak/blob/eb0c5d6d2a3942abc11054f2b46f3a99757b37bd/frontend/src/ResourceTreeNode.svelte#L132-L158

My suggestion is to set the variable the way it was before with `.lastModified`, but wrap everything in a check to make sure the object is not `undefined` before you set its field.

```suggestion
      if (resourceNodeDataMap[$selected]) {
        resourceNodeDataMap[$selected].lastModified = true;
      } else {
        resourceNodeDataMap[$selected] = {
          lastModified: true,
        };
      }
```

This change will have to be made in a couple of places where this issue was allowed to slip through the cracks in the "run components in GUI" PR.

rbs-jacob (changes requested):
Good bug fix! Noticed two small things to improve.

frontend/src/BinaryModifyView.svelte
rbs-jacob: As an alternative, there could be a helper method that does the setting. Something like this (haven't tested, just a draft off the dome).

``` javascript
function updateDataMap(id, key, value) {
  if (resourceNodeDataMap[id]) {
    resourceNodeDataMap[id][key] = value;
  } else {
    resourceNodeDataMap[id] = {key: value};
  }
}
```

Then that function could be called with `updateDataMap($selected, "lastModified", true)` in this case.

dannyp303: I didn't wind up going with the helper function

frontend/src/FindReplaceView.svelte
```diff
@@ -118,7 +118,7 @@
           if (result === "modified") {
             for (const resource of results[result]) {
               resourceNodeDataMap[resource["id"]] = {
-                modified: true,
+                lastModified: true,
               };
```
rbs-jacob, line 122: Same suggestion as the other changes – probably better to not overwrite the entire object.

rbs-jacob (changes requested):
Two small changes that seemed to have slipped through the cracks. Otherwise, this looks ready to merge.

rbs-jacob (approved):
Thanks for making the changes, looks good to me!

PR #322, opened 2023-06-06
Add initial USB DDR pad support to the GUI

rbs-jacob:

Add DDR pad support to the OFRAK GUI. Only maps buttons on the pad to keyboard shortcut callbacks – some more work is required to handle combos. 

@EdwardLarson

### DIFF
--- a/frontend/src/Gamepad.svelte
+++ b/frontend/src/Gamepad.svelte
@@ -1,0 +1,108 @@
+<script>
+  import { onMount } from "svelte";
+
+  import { shortcuts } from "./keyboard.js";
+
+  let animationLoop,
+    gamepad,
+    buttons = {
+      0: {
+        name: "Up",
+        shortcut: "arrowup",
+      },
+      1: {
+        name: "Down",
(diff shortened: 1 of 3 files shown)


frontend/src/Gamepad.svelte
```diff
@@ -0,0 +1,108 @@
...
+    if (!gamepad || !gamepad.connected) {
+      disconnected();
+      return;
+    }
+
+    gamepad.buttons.forEach((b, i) => {
+      if (!buttons[i]) {
+        return;
+      }
+      if (buttons[i].pressed && !b.pressed) {
```
EdwardLarson, line 65: This block of code is a little dense, hard to parse. I'm not sure there's a super easy way to make iit easy to read, but maybe a small comment would help.
```suggestion
     // on button release, if the button corresponds to an OFRAK shortcut and that shortcut is found, trigger the shortcut
      if (buttons[i].pressed && !b.pressed) {
```

EdwardLarson (approved):

PR #323, opened 2023-06-06
bump black version

EdwardLarson:

Bump black version to avoid spurious import-time errors with INFO logging level.

**Link to Related Issue(s)**

With the logging level set to INFO, at import time, black generates some errors about missing cache etc. These are completely inconsequential but do look a bit nasty (error file not found etc.) so could concern/confuse users. The issue in black is documented here:
https://github.com/psf/black/pull/3193

It has been resolved (making these errors show up with DEBUG log level only, which is fine I guess) but we are on an older version which doesn't include these changes.

### DIFF
--- a/ofrak_core/requirements.txt
+++ b/ofrak_core/requirements.txt
@@ -1,7 +1,7 @@
 aiohttp~=3.8.1
 aiohttp-cors~=0.7.0
 beartype~=0.12.0
-black==22.6.0
+black==23.3.0
 fdt==0.3.3
 importlib-metadata>=4.13
 intervaltree==3.1.0


rbs-jacob (approved):
Pending passing tests, this looks good!

ANogin:
@Edward-Larson this created an inconsistency - `ofrak_core` requires `black==23.3.0`, but `ofrak_io` (`ofrak_io/setup.py `) still has `black==22.6.0`.

P.S. We should land https://github.com/redballoonsecurity/ofrak/pull/218 eventually to make it possible to detect these kinds of issues in CI tests

PR #324, opened 2023-06-07
Feature/stream resource from frontend

dannyp303:

Upload files in chunks to handle files larger than 2G

**Link to Related Issue(s)**


EdwardLarson (changes requested):
Please merge master into this to make sure the changelog is updated properly (i.e. this changelog entry goes into Unreleased section, not 3.1.0 section, as automatic merge will likely do)

frontend/src/StartView.svelte
```diff
@@ -134,15 +136,33 @@
...
       return;
     }
-
-    const rootModel = await fetch(
-      `${$settings.backendUrl}/create_root_resource?name=${f.name}`,
-      {
-        method: "POST",
-        body: await f.arrayBuffer(),
+    if (f.size > warnFileSize) {
+      for (var start = 0; start < f.size; start += fileChunkSize) {
```
rbs-jacob, line 140:
```suggestion
      for (let start = 0; start < f.size; start += fileChunkSize) {
```

Do not use `var`!

- <https://medium.com/@codingsam/awesome-javascript-no-more-var-working-title-999428999994>
- <https://codeburst.io/javascript-var-let-or-const-which-one-should-you-use-2fd521b050fa>

frontend/src/StartView.svelte
```diff
@@ -134,15 +136,33 @@
...
+        let end = Math.min(start + fileChunkSize, f.size);
+        const result = await fetch(
+          `${$settings.backendUrl}/root_resource_chunk?name=${f.name}&addr=${start}`,
+          {
+            method: "POST",
+            body: await f.slice(start, end),
+          }
+        );
```
rbs-jacob, line 148: If you're feeling frisky, you can make this loop concurrent by turning it into an array with a `.map` call. Then you can collect the promises by doing something like `await Promise.all(startOffsets.map(i => doFetch(i)))`. 

Not sure if there will be much gain from many concurrent connections. Perhaps worth testing if you want? Not worth blocking the PR over this, though.

ofrak_core/ofrak/gui/server.py
```diff
@@ -260,6 +263,44 @@ async def run_until_cancelled(self):  # pragma: no cover
...
+        name = request.query.get("name")
+        addr = request.query.get("addr")
+        if name is None:
+            raise HTTPBadRequest(reason="Missing resource name from request")
+        if addr is None:
+            raise HTTPBadRequest(reason="Missing chunk address from request")
+        if name not in self.resource_builder.keys():
+            self.resource_builder[name] = {}
+        chunk_data = await request.read()
+        self.resource_builder[name][int(addr)] = chunk_data
```
rbs-jacob, line 277: Doing this dictionary keying based on filename seems risky and prone to collisions... 

Can you create a resource first, get its ID, and then update its data after everything has been uploaded?

rbs-jacob (changes requested):

frontend/src/StartView.svelte
dannyp303: Need to use .map more, i'll do that.

frontend/src/StartView.svelte
dannyp303: woops! this should be outdated now anyway

ofrak_core/ofrak/gui/server.py
```diff
@@ -260,6 +264,68 @@ async def run_until_cancelled(self):  # pragma: no cover
...
 
+    @exceptions_to_http(SerializedError)
+    async def init_chunked_root_resource(self, request: Request) -> Response:
+        name = request.query.get("name")
+        size_param = request.query.get("size")
+        size = int(size_param) if size_param is not None else None
+        if name is None:
+            raise HTTPBadRequest(reason="Missing resource name from request")
+        if size is None:
+            raise HTTPBadRequest(reason="Missing chunk size from request")
```
rbs-jacob, line 275: Minor, but I think this is a little easier to read. May also make mypy happier.

```suggestion
        if name is None:
            raise HTTPBadRequest(reason="Missing resource name from request")
        if size_param is None:
            raise HTTPBadRequest(reason="Missing chunk size from request")
        size = int(size_param)
```

ofrak_core/ofrak/gui/server.py
```diff
@@ -260,6 +264,68 @@ async def run_until_cancelled(self):  # pragma: no cover
...
+        root_resource: Resource = await self._ofrak_context.create_root_resource(
+            name, bytearray(b"\x00" * size), (File,)
```
rbs-jacob, line 277: Two questions:

- What is the benefit of creating the resource with a `bytearray` instead of just `bytes`?
- What is the benefit of creating a resource with any data at all? It will all be overwritten after the chunks come in, no?

ofrak_core/ofrak/gui/server.py
```diff
@@ -260,6 +264,68 @@ async def run_until_cancelled(self):  # pragma: no cover
...
+        )
+        return json_response(root_resource.get_id().hex())
+
+    @exceptions_to_http(SerializedError)
+    async def root_resource_chunk(self, request: Request) -> Response:
+        id = request.query.get("id")
+        start_param = request.query.get("start")
+        start = int(start_param) if start_param is not None else None
+        end_param = request.query.get("end")
+        end = int(end_param) if end_param is not None else None
```
rbs-jacob, line 291: Same change as above – move below the `is None` checks?

ofrak_core/ofrak/gui/server.py
```diff
@@ -260,6 +264,68 @@ async def run_until_cancelled(self):  # pragma: no cover
...
+            return HTTPBadRequest(reason="Missing root resource `id` from request")
+        if name is None:
+            return HTTPBadRequest(reason="Missing root resource `name` from request")
+
+        try:
+            root_resource, data = self.resource_builder[id]
+            script_str = rf"""
+            if root_resource is None:
+                root_resource = await ofrak_context.create_root_resource_from_file("{name}")"""
+            root_resource.queue_patch(Range(0, len(bytearray(data))), bytearray(data))
```
rbs-jacob, line 317: Is it necessary to cast `data` to a `bytearray` to get the length? I would worry about it doing expensive constructor operations just to throw away the object after it gets the length. If it is necessary to cast for getting the lenght, maybe store the casted version in a variable instead of casting twice?

```suggestion
            root_resource.queue_patch(Range(0, len(data)), bytearray(data))
```

rbs-jacob (changes requested):

EdwardLarson (approved):

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -119,6 +119,23 @@ async def test_create_root_resource(
...
+        end = min(start + chunk_size, len(hello_world_elf))
+        res = await ofrak_client.post(
+            "/root_resource_chunk",
+            params={"name": "hello_world_elf", "addr": start},
+            data=hello_world_elf[start:end],
+        )
+    create_resp = await ofrak_client.post(
+        "/create_chunked_root_resource", params={"name": "hello_world_elf"}
+    )
+    assert create_resp.status == 200
```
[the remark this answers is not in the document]
rbs-jacob: @dannyp303 if you're going to leave this unresolved, do you think it makes sense to make an issue to track that someday we want to add a test like this?

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -26,6 +29,14 @@ def hello_world_elf() -> bytes:
     return hello_elf()
 
 
+@pytest.fixture()
+async def large_test_file(ofrak_context: OFRAKContext) -> Resource:
+    with tempfile.NamedTemporaryFile() as temp:
+        for i in range(0, 256, 1):
+            temp.write(int.to_bytes(i, 1, "big") * 4096)
+        yield await ofrak_context.create_root_resource_from_file(temp.name)
```
rbs-jacob, line 37: Two minor things:

- I think the suggestion below is equivalent, but a little more concise
- Is this file big enough? By my calculation, it's `4096 * 4 * 256 ≈ 4MB`, but not really on the order of the gigabytes we might be seeing with this changeset

```suggestion
async def large_test_file(ofrak_context: OFRAKContext) -> Resource:
    with tempfile.NamedTemporaryFile() as temp:
        for i in range(256):
            temp.write(int.to_bytes(i, 1, "big") * 4096)
        yield await ofrak_context.create_root_resource_from_file(temp.name)
```

PR #332, opened 2023-06-12
Bugfix: use ghidra config

EdwardLarson:

Actually use the fields of the ofrak_ghidra.config when setting up the ghidra repository

**Link to Related Issue(s)**
Basically we had infrastructure to configure ofrak for a local ghidra setup, but the ofrak_ghidra code was not looking at that config at all.

### DIFF
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra/run_ghidra_server.sh
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra/run_ghidra_server.sh
@@ -9,13 +9,13 @@ GHIDRA_REPO_HOST=$5
 GHIDRA_REPO_PORT=$6
 
 # cd /opt/rbs/ghidra_10.1.2_PUBLIC
-cd ${GHIDRA_PATH}
+cd "${GHIDRA_PATH}"
 ./server/svrInstall
 
 # The following command sometimes fails for unclear reasons, so we give it a few tries.
 max_retries=5
 retry_counter=0
-while ! ./server/svrAdmin -add $GHIDRA_USER; do
+while ! ./server/svrAdmin -add "$GHIDRA_USER"; do
   if [ $retry_counter -ge $max_retries ]; then
     echo "Failure running './server/svrAdmin -add $GHIDRA_USER'. Retried $retry_counter times." >&2
     exit 1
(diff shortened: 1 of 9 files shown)


EdwardLarson:
Definitely remaining TODO:

[ ] Adding a bit of logic to handle existing ofrak ghidra configs, as the expected hosts have changed a bit (`ghidra://` or `http://` prefix is now added automatically to the hostname) so that pulling/updating this wont break existing setups

Possible remaining TODO:

[ ] Logic to start up and shut down the ofrak_ghidra server when starting/stopping OFRAK

The main issue with that is that you probably want to leave it up when doing some development with ofrak_ghidra, so you don't have to be constantly starting/stopping the repository server. So maybe the repository should be left alive when OFRAK exits. This is already essentially the case so it should be benign to do so.

disassemblers/ofrak_ghidra/ofrak_ghidra/run_ghidra_server.sh
```diff
@@ -23,5 +42,11 @@ while ! ./server/svrAdmin -add root; do
   echo "Retrying running './server/svrAdmin -add root' ($retry_counter/$max_retries)..."
 done
 
+
 ./server/ghidraSvr restart
-./support/analyzeHeadless . dummy -postScript CreateRepository.java -scriptPath ${OFRAK_GHIDRA_SCRIPTS_PATH} -deleteProject -noanalysis
+./support/analyzeHeadless . dummy -postScript CreateRepository.java $GHIDRA_USER $GHIDRA_PASS $GHIDRA_REPO_HOST $GHIDRA_REPO_PORT -scriptPath ${OFRAK_GHIDRA_SCRIPTS_PATH} -deleteProject -noanalysis
```
rbs-jacob, line 47: Shell variables should be quoted when used in case they have spaces.

Probably worth running this script through [`shellcheck`](https://github.com/koalaman/shellcheck) to automatically find potential problems.

```suggestion
./support/analyzeHeadless . dummy -postScript CreateRepository.java "$GHIDRA_USER" "$GHIDRA_PASS" "$GHIDRA_REPO_HOST" "$GHIDRA_REPO_PORT" -scriptPath "${OFRAK_GHIDRA_SCRIPTS_PATH}" -deleteProject -noanalysis
```

disassemblers/ofrak_ghidra/ofrak_ghidra/server/__main__.py
```diff
@@ -7,18 +7,32 @@
...
+    GHIDRA_PASS,
+    GHIDRA_REPOSITORY_HOST,
+    GHIDRA_REPOSITORY_PORT,
 )
 
 
-def _run_ghidra_server(args):
+def _run_ghidra_server(*args):
     if sys.platform == "linux" or sys.platform == "darwin":
         subprocess.call(["chmod", "+x", GHIDRA_START_SERVER_SCRIPT])
```
rbs-jacob, line 19: I know this isn't directly part of the PR, but I wonder if it makes sense to use the more portable [`os.chmod`](https://docs.python.org/3/library/os.html#os.chmod) here? More portable, but not necessarily more readable. 

```suggestion
        os.chmod(GHIDRA_START_SERVER_SCRIPT, os.stat(GHIDRA_START_SERVER_SCRIPT).st_mode | stat.S_IEXEC)
```

rbs-jacob (approved):
Very minor suggested changes. Other than that, this is good to go!

rbs-jacob (approved):

PR #334, opened 2023-06-22
Resource Tree Bytes and Strings Search Bar 

dannyp303:

Add a search bar to the ResourceTreeView to filter resources by string and bytes. 

**Link to Related Issue(s)**

@rbs-jacob @Edward-Larson

### DIFF
--- a/frontend/src/ResourceSearchBar.svelte
+++ b/frontend/src/ResourceSearchBar.svelte
@@ -86,7 +86,7 @@
     {/each}
   </select>
   <form
-    on:keyup|preventDefault="{async (e) => {
+    on:submit|preventDefault="{async (e) => {
       try {
         if (searchType == 'String') {
           searchFilter = await rootResource.search_for_string(
--- a/frontend/src/ResourceSearchBar.svelte
+++ b/frontend/src/ResourceSearchBar.svelte
@@ -60,7 +60,11 @@
   </select>
   <form
     on:submit|preventDefault="{async (e) => {
-      searchFilter = await rootResource.search_for_string(searchQuery);
+      if (searchType == 'String') {
+        searchFilter = await rootResource.search_for_string(searchQuery);
+      } else if (searchType == 'Bytes') {
+        searchFilter = await rootResource.search_for_bytes(searchQuery);
+      }
     }}"
   >
     <label>
--- a/frontend/src/ResourceSearchBar.svelte
+++ b/frontend/src/ResourceSearchBar.svelte
@@ -44,7 +44,7 @@
 <script>
   import Checkbox from "./Checkbox.svelte";
   export let rootResource, searchFilter;
-  let searchType, searchQuery, bytesInput, regex, ph_string;
+  let searchType, searchQuery, bytesInput, regex, ph_string, case_ignore;
   let searchTypes = ["String", "Bytes"];
 
   $: if (searchQuery != null && searchType == "Bytes") {
@@ -80,7 +80,11 @@
   <form
     on:keyup|preventDefault="{async (e) => {
       if (searchType == 'String') {
-        searchFilter = await rootResource.search_for_string(searchQuery, regex);
+        searchFilter = await rootResource.search_for_string(
+          searchQuery,
+          regex,
+          case_ignore
+        );
       } else if (searchType == 'Bytes') {
         searchFilter = await rootResource.search_for_bytes(searchQuery, false);
       }
@@ -94,8 +98,11 @@
 
 <div class="optionbar">
   {#if searchType == "String"}
-    <Checkbox checked="{false}" bind:value="{regex}" leftbox="{true}">
+    <Checkbox checked="{case_ignore}" bind:value="{regex}" leftbox="{true}">
       Pattern
+    </Checkbox>
+    <Checkbox checked="{false}" bind:value="{case_ignore}" leftbox="{true}">
+      Ignore Case
     </Checkbox>
   {/if}
 </div>
--- a/frontend/src/ResourceSearchBar.svelte
+++ b/frontend/src/ResourceSearchBar.svelte
@@ -16,6 +16,7 @@
     border: 1px solid;
     border-bottom: 1px solid var(--main-fg-color);
     flex: 1;
+    text-indent: .5em;
   }
 
   form {
@@ -60,13 +61,13 @@
 
   $: if (searchType == "String") {
     if (regex) {
-      placeholderString = " Search for a Regex Pattern";
+      placeholderString = "Search for a Regex Pattern";
     } else {
-      placeholderString = " Search for a String";
+      placeholderString = "Search for a String";
     }
   } else if (searchType == "Bytes") {
     regex = false; // Regex for bytes not yet implemented
-    placeholderString = " Search for Bytes";
+    placeholderString = "Search for Bytes";
   }
 </script>
 
--- a/frontend/src/ResourceSearchBar.svelte
+++ b/frontend/src/ResourceSearchBar.svelte
@@ -48,7 +48,7 @@
   let searchType, searchQuery, bytesInput, regex, placeholderString, caseIgnore;
   let searchTypes = ["String", "Bytes"];
 
-  $: if (searchQuery != null && searchType == "Bytes") {
+  $: if (searchQuery && searchType === "Bytes") {
     try {
       searchQuery = searchQuery.match(/[0-9a-fA-F]{1,2}/g).join(" ");
       bytesInput?.setCustomValidity("");
--- a/frontend/src/ResourceSearchBar.svelte
+++ b/frontend/src/ResourceSearchBar.svelte
@@ -44,9 +44,12 @@
   let searchTypes = ["String", "Bytes"];
 
   $: if (searchQuery != null && searchType == "Bytes") {
-    searchQuery = searchQuery.match(/[0-9a-fA-F]{1,2}/g).join(" ");
-    console.log(searchQuery);
-    bytesInput?.setCustomValidity("");
+    try {
+      searchQuery = searchQuery.match(/[0-9a-fA-F]{1,2}/g).join(" ");
+      bytesInput?.setCustomValidity("");
+    } catch {
+      bytesInput?.setCustomValidity("Invalid bytes representation.");
+    }
   }
 </script>
 
@@ -71,11 +74,7 @@
       {#if searchType == "String"}
         <input placeholder=" Search for a String" bind:value="{searchQuery}" />
       {:else if searchType == "Bytes"}
-        <input
-          placeholder=" Search for Bytes"
-          pattern="([0-9a-fA-F][0-9a-fA-F]\s*)*"
-          bind:value="{searchQuery}"
-        />
+        <input placeholder=" Search for Bytes" bind:value="{searchQuery}" />
       {/if}
     </label>
   </form>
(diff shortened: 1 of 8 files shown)


ofrak_core/ofrak/gui/server.py
```diff
@@ -739,6 +742,61 @@ async def search_for_vaddr(self, request: Request) -> Response:
...
+            raise ValueError("Invalid search query.")
+        string_query: Union[bytes, re.Pattern[bytes]] = string_query_string.encode()
+        if case_ignore:
+            string_query = re.compile(string_query, re.IGNORECASE)
+        elif regex:
+            string_query = re.compile(string_query)
+
+        offsets = await resource.search_data(string_query)
+        if string_query == "":
+            return json_response(None)
```
rbs-jacob, line 762: Is there a reason to return `None` instead of an empty list here? 

Also, if `string_query == ""`, I think it makes sense to return earlier than this. Not only does it make sense to return before actually doing the search (to avoid the expensive operation of searching for an empty string that will be present between every single character), but it probably makes sense to return before compiling the regular expression as well.

ofrak_core/ofrak/gui/server.py
```diff
@@ -739,6 +742,61 @@ async def search_for_vaddr(self, request: Request) -> Response:
...
+        found_resources = []
+        if len(offsets) > 0:
+            # found_resources[resource.get_id().hex()] = offsets
```
rbs-jacob, line 765: Dead code? Also repeated below in 3 other spots.

```suggestion
```

ofrak_core/ofrak/gui/server.py
```diff
@@ -739,6 +742,61 @@ async def search_for_vaddr(self, request: Request) -> Response:
...
+
+        return json_response(found_resources)
+
+    @exceptions_to_http(SerializedError)
+    async def search_for_bytes(self, request: Request):
+        resource = await self._get_resource_for_request(request)
+        body = await request.json()
+        bytes_query_request = body["search_query"]
+        if bytes_query_request == "":
+            return json_response(None)
```
rbs-jacob, line 783: Same as above: why return `None` instead of the empty list? To me, the latter is more correct – there are no results, so we return an empty list of results. Whereas `None` feels more suited to actual error responses.

Plus, if an empty list is returned, the front end never needs to check for a `None` response. It can just use the normal logic for handling a list in a loop, but the loop body will simply be skipped.

ofrak_core/ofrak/gui/server.py
```diff
@@ -739,6 +742,61 @@ async def search_for_vaddr(self, request: Request) -> Response:
...
+        bytes_query = bytes.fromhex("".join(bytes_query_request.split(" ")))
```
rbs-jacob, line 784: Why split and join instead of just removing all non-hex characters?

```suggestion
        bytes_query = bytes.fromhex(re.sub(r"[^0-9a-fA-F]+", "", bytes_query_request))
```

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -888,6 +888,40 @@ async def test_get_config(ofrak_client: TestClient, hello_world_elf):
...
+    create_resp = await ofrak_client.post(
+        "/create_root_resource", params={"name": "hello_world_elf"}, data=hello_world_elf
+    )
+    create_body = await create_resp.json()
+    resource_id = create_body["id"]
+    resp = await ofrak_client.post(
+        f"/{resource_id}/search_for_string",
+        json={"search_query": "hello", "case_ignore": True, "regex": True},
+    )
+    resp_body = await resp.json()
```
rbs-jacob, line 901: We should probably add an assert that it actually finds the string. And we should make a similar check that it searches for one that isn't there and doesn't find it. 

Should be easy to get that from the body and assert.

That way we're actually confirming that this route works and does what it says it's supposed to.

rbs-jacob (changes requested):

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,109 @@
...
+
+<script>
+  import Checkbox from "./Checkbox.svelte";
+  export let rootResource, searchFilter;
+  let searchType, searchQuery, bytesInput, regex, placeholderString, caseIgnore;
+  let searchTypes = ["String", "Bytes"];
+
+  $: if (searchQuery != null && searchType == "Bytes") {
+    try {
+      searchQuery = searchQuery.match(/[0-9a-fA-F]{1,2}/g).join(" ");
```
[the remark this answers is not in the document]
dannyp303: because I'm updating the input reactively, i need to match both 1 and 2 or else the first digit will never make it into the input. This is so we force the bytes format in the input.

frontend/src/ResourceTreeNode.svelte
```diff
@@ -156,7 +157,7 @@
     lastModified = resourceNodeDataMap[self_id].lastModified;
     allModified = resourceNodeDataMap[self_id].allModified;
   }
-
+  $: filter = searchFilter;
```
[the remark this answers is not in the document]
dannyp303: I was getting cyclical reactivity with `$: searchFilter = searchFilter` and without a reactive block here the filter would not update when updated in the other component. If you see another way to do it I would be grateful.

ofrak_core/ofrak/gui/server.py
```diff
@@ -739,6 +742,61 @@ async def search_for_vaddr(self, request: Request) -> Response:
...
+        resource = await self._get_resource_for_request(request)
+        body = await request.json()
+        string_query_string = body["search_query"]
+        regex = body["regex"]
+        case_ignore = body["case_ignore"]
+        if not isinstance(string_query_string, str):
+            raise ValueError("Invalid search query.")
+        string_query: Union[bytes, re.Pattern[bytes]] = string_query_string.encode()
+        if case_ignore:
+            string_query = re.compile(string_query, re.IGNORECASE)
```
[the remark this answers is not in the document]
dannyp303: This annoyed me, but I couldn't find an alternate way. The only way I could see to efficiently do case insensitivity is with regex, so if someone uses a regex pattern it will still work even if not checked. So I made them mutually inclusive, and when you check `Ignore Case` in the GUI it will auto check the regex option as well.

ofrak_core/ofrak/gui/server.py
dannyp303: On the first part, yes. An empty list should be the result of a search with no results. Otherwise it isn't apparent that your search is empty. And yeah you're correct, I did this fast this morning.

ofrak_core/ofrak/gui/server.py
dannyp303: Ah yeah kind of, I'm planning on using those after ed gets his hex search in so it will automatically focus the location of the result in the hex view, but that should go in a seperate PR.

ofrak_core/ofrak/gui/server.py
dannyp303: It looks nicer, easier to read for long strings of bytes, much of our hex input follows this format.

ofrak_core/ofrak/gui/server.py
rbs-jacob: Second part sounds good, makes sense to return `None`.

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,109 @@
...
+      if (searchType == 'String') {
+        searchFilter = await rootResource.search_for_string(
+          searchQuery,
+          regex,
+          caseIgnore
+        );
+      } else if (searchType == 'Bytes') {
+        searchFilter = await rootResource.search_for_bytes(searchQuery, false);
+      }
+    }}"
```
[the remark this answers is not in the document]
dannyp303: pop up alert doesn't make sense now that i'm using `on:keyup`, it would probably spam the browser.

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,110 @@
...
+  <select bind:value="{searchType}">
+    {#each searchTypes as type}
+      <option value="{type}">
+        {type}
+      </option>
+    {/each}
+  </select>
+  <form
+    on:keyup|preventDefault="{async (e) => {
+      if (searchType == 'String') {
```
EdwardLarson: Should check for if the query is an empty string. My browser completely froze up when I cleared the search bar, I believe because it was finding a ridiculous number of matches.

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,129 @@
...
+  >
+    <label>
+      <input placeholder="{placeholderString}" bind:value="{searchQuery}" />
+    </label>
+  </form>
+</div>
+
+<div class="optionbar">
+  {#if searchType == "String"}
+    <Checkbox checked="{caseIgnore}" bind:value="{regex}" leftbox="{true}">
```
EdwardLarson, line 122: So when `caseIgnore` is checked, `regex` will be automatically checked as well? On my machine this seems to only work one-way, I can't check `regex` individually. I guess because the `caseIgnore` variable is still false it semi-reactively is setting `regex` back to false right away, if that makes sense.

dannyp303: Yeah, `regex` will automatically check itself when `caseIgnore` is checked. I can select `regex` individually on mine. When you uncheck `caseIgnore` it will uncheck `regex` with it, I think that makes sense though. Only changes to the `caseIgnore` variable should affect `regex`, it shouldn't check state otherewise, not sure if that is what you are seeing?

ofrak_core/ofrak/gui/server.py
```diff
@@ -739,6 +742,57 @@ async def search_for_vaddr(self, request: Request) -> Response:
...
+        elif regex:
+            string_query = re.compile(string_query)
+
+        offsets = await resource.search_data(string_query)
+        found_resources = []
+        if len(offsets) > 0:
+            found_resources.append(resource.get_id().hex())
+        for child in await resource.get_descendants():
+            offsets = await child.search_data(string_query)
```
EdwardLarson, line 767:
```suggestion
            if child.get_data_id() is None:
                continue
            offsets = await child.search_data(string_query)
```
Handles resources that don't have data, otherwise an AttributeError gets raised.

frontend/src/ResourceSearchBar.svelte
EdwardLarson: Not sure the exact source difference that makes it behave differently for me, but I did put the regex and caseIgnore variables inside an object. In any case, I found a solution:

`checked="{searchOptions.caseIgnore || searchOptions.regex}"`

Only downside is that the regex checkbox stays checked once caseIgnore is unchecked, but that's pretty minor.

ofrak_core/ofrak/gui/server.py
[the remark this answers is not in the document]
rbs-jacob: One solution is to use `.lower()` (or `.casefold()`) on the string before running the query. But that's definitely less efficient than a regular expression. 

To continue using the regular expression, you should probably [run `re.escape` on the `string_query` first(https://docs.python.org/3/library/re.html#re.escape).

```suggestion
            string_query = re.compile(re.escape(string_query), re.IGNORECASE)
```

EdwardLarson:
Folded into https://github.com/redballoonsecurity/ofrak/pull/345

PR #336, opened 2023-06-26
Feature: GUI hex search

EdwardLarson:

Add searchbar to HexView in order to search for a string or bytes within a resource.

**Link to Related Issue(s)**

@rbs-jacob 
@dannyp303


frontend/src/HexByte.svelte
```diff
@@ -0,0 +1,61 @@
+<style>
```
rbs-jacob, line 1: Does this component even actually get used anywhere?

ofrak_core/ofrak/core/elf/analyzer.py
```diff
@@ -322,15 +322,11 @@ async def analyze(self, resource: Resource, config=None) -> AttributesType[Named
         elf_r = await section.get_elf()
         string_section = await elf_r.get_section_name_string_section()
         try:
-            string_section_data = await string_section.resource.get_data(
-                Range(section_header.sh_name, Range.MAX)
+            ((_, raw_section_name),) = await string_section.resource.search_data(
+                re.compile(b".[^\x00]+\x00"), start=section_header.sh_name, max_matches=1
```
rbs-jacob, line 326: It probably makes sense to compile this outside of this function call. Also, why does there have to be a non-newline character, followed by one or more non-null bytes? I think the initial `.` is erroneous.

```suggestion
                re.compile(b"[^\x00]+\x00"), start=section_header.sh_name, max_matches=1
```

ofrak_core/ofrak/core/elf/model.py
```diff
@@ -512,10 +513,10 @@ def get_section_index(self) -> Optional[int]:
     async def get_name(self) -> str:
         elf = await self.resource.get_only_ancestor_as_view(Elf, ResourceFilter.with_tags(Elf))
         string_section = await elf.get_string_section()
-        string_section_data = await string_section.resource.get_data(Range(self.st_name, Range.MAX))
-        name_string_end = string_section_data.find(b"\x00")
-        raw_symbol_name = string_section_data[:name_string_end]
-        return raw_symbol_name.decode("ascii")
+        ((_, raw_symbol_name),) = await string_section.resource.search_data(
+            re.compile(b".[^\x00]+\x00"), start=self.st_name, max_matches=1
```
rbs-jacob, line 517: Same comment as above. This should probably only be compiled once. And the `.` seems unnecessary.

```suggestion
            re.compile(b"[^\x00]+\x00"), start=self.st_name, max_matches=1
```

ofrak_core/ofrak/core/strings.py
```diff
@@ -82,8 +82,7 @@ async def modify(self, resource: Resource, config: StringFindReplaceConfig) -> N
                 f"If you expect that the string to replace is null-terminated, then an overflow "
                 f"of one byte when config.null_terminate = True will not have any effect."
             )
-        original_data = await resource.get_data()
-        offsets = [m.start() for m in re.finditer(to_find, original_data)]
+        offsets = [offset for offset in await resource.search_data(to_find)]
```
rbs-jacob, line 85: Does it even need to be cast to a list? It [looks like it returns a `tuple`](), but the docs describe it as returning a "list." Also, there are no type annotations for the return value :(

```suggestion
        offsets = list(await resource.search_data(to_find))
```

https://github.com/redballoonsecurity/ofrak/blob/6e7289501bfa62989145459162b378094bb0ab59/ofrak_core/ofrak/service/data_service.py#L149-L174

https://github.com/redballoonsecurity/ofrak/blob/6e7289501bfa62989145459162b378094bb0ab59/ofrak_core/ofrak/resource.py#L267-L280

ofrak_core/ofrak/resource.py
```diff
@@ -242,6 +244,42 @@ async def get_data_range_within_root(self) -> Range:
...
+        Search for some data in this resource. The query may be a regex pattern (a return value
+        of `re.compile`). If the query is a regex pattern, returns a list of pairs with both the
+        offset of the match and the contents of the match itself. If the query is plain bytes, a
+        list of only the match offsets are returned.
+
+        :param query: Plain bytes to exactly match or a regex pattern to search for
+        :param start: Start offset in the data model to begin searching
+        :param end: End offset in the data model to stop searching
+
+        :return: A list of offsets matching a plain bytes query, or a list of (offset, match) pairs
```
rbs-jacob, line 278:
```suggestion
        :return: A tuple of offsets matching a plain bytes query, or a list of (offset, match) pairs
```

ofrak_core/ofrak/service/data_service.py
```diff
@@ -134,6 +146,33 @@ async def delete_models(self, data_ids: Iterable[DataId]) -> None:
             root.delete_mapped_model(model)
             del self._model_store[model.id]
 
+    async def search(self, data_id, query, start=None, end=None, max_matches=None):
```
rbs-jacob, line 149: Type annotation on return value?

ofrak_core/ofrak/service/data_service_i.py
```diff
@@ -180,3 +180,46 @@ async def delete_models(self, data_ids: Iterable[bytes]) -> None:
...
+        offset of the match and the contents of the match itself. If the query is plain bytes, a
+        list of only the match offsets are returned.
+
+        :param data_id: Data model to search
+        :param query: Plain bytes to exactly match or a regex pattern to search for
+        :param start: Start offset in the data model to begin searching
+        :param end: End offset in the data model to stop searching
+        :param max_matches: Maximum number of matches to return
+
+        :return: A list of offsets matching a plain bytes query, or a list of (offset, match) pairs
```
rbs-jacob, line 222:
```suggestion
        :return: A tuple of offsets matching a plain bytes query, or a list of (offset, match) pairs
```

ofrak_core/ofrak/service/data_service.py
```diff
@@ -134,6 +146,33 @@ async def delete_models(self, data_ids: Iterable[DataId]) -> None:
...
+        model = self._get_by_id(data_id)
+        root = self._get_root_by_id(model.root_id)
+        start = model.range.start if start is None else model.range.start + start
+        end = model.range.end if end is None else min(model.range.end, model.range.start + end)
+        if type(query) is bytes:
```
rbs-jacob, line 154: Support for subclassing is more robust to future changes.

```suggestion
        if isinstance(query, bytes):
```

ofrak_core/ofrak/core/strings.py
```diff
@@ -145,13 +144,12 @@ async def unpack(self, resource: Resource, config: None) -> None:
         else:
             pattern = self.SHORT_STRING_PATTERN
 
-        data = await resource.get_data()
-
         children = [
             resource.create_child_from_view(
-                AsciiString(m.group(1).decode("ascii")), data_range=Range(m.start(), m.end())
+                AsciiString(string[:-1].decode("ascii")),
```
rbs-jacob, line 149: Is the actual goal here to strip off the last byte? Or is it to `rstrip` at most one trailing null bytes? Might make more sense to implement as the latter, since that's closer to the actual intention.

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,130 @@
...
+  >
+    <label>
+      <input
+        placeholder=" Search for a {searchType}"
+        bind:value="{searchQuery}"
+      />
+    </label>
+  </form>
+  <div class="resultwidgets">
+    {#if searchResults.matches !== undefined && searchResults.matches !== null}
```
rbs-jacob, line 119: Empty lists are not falsy in JavaScript, so casing on `searchResults.matches` is true is not going to accidentally catch the "No match" case.

```suggestion
    {#if searchResults.matches}
```

rbs-jacob (changes requested):
Requested changes.

I just reviewed by looking over stuff. I figure I'll actually run it, and probably make some CSS adjustments, after this first round of changes goes in. That way we don't spend time changing CSS when functionality will change, possibly rendering the CSS work obsolete.

frontend/src/HexByte.svelte
```diff
@@ -0,0 +1,61 @@
...
+        return result;
+      } else if (range.start < T) {
+        L = m + 1;
+      } else if (range.start > T) {
+        R = m - 1;
+      }
+    }
+
+    return null;
+  }
```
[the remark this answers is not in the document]
EdwardLarson: This HexByte.svelte was erroneously committed, it's unused and shouldn't be in this PR. Do you still think the binary search implementation should be moved to helpers.js if it's only used in HexView.svelte?

ofrak_core/ofrak/core/strings.py
EdwardLarson: The regex only matches exactly one trailing null byte, but I agree that a `strip()` call is clearer.

ofrak_core/ofrak/service/data_service.py
EdwardLarson: The typing is all done in the `DataServiceInterface` because it requires 2 type stubs via overloading.

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,130 @@
...
+    searchResults = { matches: searchResults.matches, index: nextIndex };
+  }
+
+  function prevMatch() {
+    let nextIndex = searchResults.index - 1;
+    if (nextIndex < 0) {
+      nextIndex = Math.max(searchResults.matches.length - 1, 0);
+    }
+    searchResults = { matches: searchResults.matches, index: nextIndex };
+  }
```
rbs-jacob, line 86: Another strange characteristic is the way that multiple contiguous regions can be matched multiple times. Personally, I think this behavior should be considered a bug. Even if it's intended, it's unintuitive for the user what's going on. 

Maybe the particular instance of the found string that corresponds to the latest find should be highlighted in a different color? Kind of how <kbd>Ctrl</kbd>+<kbd>f</kbd> in the browser will highlight all the instances it's found, but highlight the current one in a different color. 

https://github.com/redballoonsecurity/ofrak/assets/99368685/52f9fe00-4218-4d5a-ad00-bec16db02d93

rbs-jacob (changes requested):
Actually ran it this time. Found a few small bugs/issues in terms of UI behavior, and made some CSS change recommendations. In the interest of learning, I did not commit them directly to the branch, but tried to break them out into PR comments that explain and justify.

frontend/src/ResourceSearchBar.svelte
```diff
@@ -0,0 +1,130 @@
...
+      } else {
+        searchResults.matches = await search(searchQuery, searchType);
+        searchResults.index = 0;
+        prevQuery = searchQuery;
+      }
+    }}"
+  >
+    <label>
+      <input
+        placeholder=" Search for a {searchType}"
```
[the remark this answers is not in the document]
dannyp303: FYI ed merged my fork into his for some of this, might make sense to just pull from there again

frontend/src/ResourceSearchBar.svelte
EdwardLarson: I did a simple implementation to try to address this, just underlining the bytes if the range they match (from the binary search) is the same index as the selected range. It sort of works, but doesn't work well for exactly the case being raised here: Multiple matching ranges overlapping, since not all of the bytes in the currently selected range are highlighted.  I would guess that the binary search finds one of the overlapping ranges instead, not prioritizing the selected one.

So I think it would actually considerably complicate the logic, unless you see another simple way to try it. I think we should leave this as future work.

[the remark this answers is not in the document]
EdwardLarson: It's not just if there is only result found, the result has to be exactly at the start of a row. Honestly this sort of off-by-one error seems like it will take more time to fix than it's worth. The search bar still jumps _basically_ to the match, so to the user experience is barely impacted. But this is definitely not intended behavior.

EdwardLarson:
Folded into https://github.com/redballoonsecurity/ofrak/pull/345

PR #337, opened 2023-06-26
EXT Filesystem Components

rbs-jacob:

Add packers and unpackers for EXT filesystems (versions 2 through 4).

@whyitfor @EdwardLarson 

**Outstanding tasks:**

- [ ] Packer
- [x] Tests

### DIFF
--- a/ofrak_core/ofrak/core/extfs.py
+++ b/ofrak_core/ofrak/core/extfs.py
@@ -1,4 +1,5 @@
 import asyncio
+import re
 import tempfile
 from dataclasses import dataclass
 from subprocess import CalledProcessError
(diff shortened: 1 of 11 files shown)


EdwardLarson (commented):
Is there any dependencies that need to be installed? That is, is a change to the Dockerstub missing from this PR?

IIRC this doesn't work for Windows. I feel it's alright that this only works on Linux/Mac systems, if not ideal. The `ComponentExternalTool` being there can protect against trying to run it on Windows. I do think there should be a TODO comment somewhere in here explicitly saying Windows support is limited.

rbs-jacob:
> Is there any dependencies that need to be installed? That is, is a change to the Dockerstub missing from this PR?
> 
> IIRC this doesn't work for Windows. I feel it's alright that this only works on Linux/Mac systems, if not ideal. The `ComponentExternalTool` being there can protect against trying to run it on Windows. I do think there should be a TODO comment somewhere in here explicitly saying Windows support is limited.

On macOS, there is a dependency that needs to be installed via `brew`, and is included as a `ComponentExternalTool`. There are no such external dependencies on Linux, since I believe Debian bundles the required tools by default.

```
jacob@blade:~$ docker run --rm -it debian:latest bash
root@25be021f7df1:/# which debugfs
/usr/sbin/debugfs
```

While it's true that I haven't been able to find a version that works for Windows, I suspect that the `7z` command line tool can at least extract EXT filesystems. It may not be able to rebuild them, though.

rbs-jacob:
In the interest of getting this merged in, we're omitting a packer for now. A packer is considerably more complicated to add because of cross-platform concerns with the normal workflow for making a populated EXT filesystem. Whereas for unpacking, the `debugfs` utility can easily be used cross-platform. 

The workflow for packing usually involves:

- Making an empty file with some `dd` incantation
- Using some variation of `mkfs.ext4` to make the file into a filesystem
- Mounting the filesystem with `mount`
- Populating the filesystem
- Unmounting it

This can all be done fairly easily from a Linux system, but for other OSes, the kernels lack support for EXT filesystems, so they cannot easily be mounted. The packer will likely have to involve some alternative, cross platform way to build and populate them. As such, we're waiting for a separate PR for the packer so that this work doesn't stall in the short-term.

EdwardLarson (commented):
The debugf dependency should be "private" so as to not be imported elsewhere. 

The test cases could maybe be combined into one class rather than a separate one for each EXT version, maybe with some clever pytest fixtures, but that's arguably harder to read and not important anyway.

Issue #238, opened 2023-02-27
PatchMaker VLE support


rbs-jacob:
**What is the use case for the feature?**

Patching VLE binaries with the OFRAK PatchMaker.

**Does the feature contain any proprietary information about another company's intellectual property?**

Negative.

**How would you implement this feature?**

95% implemented on https://github.com/redballoonsecurity/ofrak/compare/master...rbs-jacob:ofrak:feature/ppc-vle-toolchain The remaining 5% was discussed briefly on #204, and is summarized below.

Though free to download, the NXP toolchain required for VLE patching is gated behind a login wall. It therefore can't be pulled automatically by the [`ofrak_patch_maker/Dockerstub`](https://github.com/redballoonsecurity/ofrak/compare/master...rbs-jacob:ofrak:feature/ppc-vle-toolchain#diff-f9762e6fa57d8376ce4fe0518ab490f8c989e35307196af540cacd7df928a013R80) install. A couple of strategies to overcome this:

- Conditionally copy the ZIP into the Docker image and don't install it if the file doesn't exist
    - This would work for users, but would cause CI tests for the VLE PatchMaker to fail
- Bundle the files directly
    - The GNU toolchain is licensed under GPL, but I don't know if it acceptable for us to distribute NXP's builds outside of their auth wall
- Use a different toolchain with VLE support

**Are you interested in implementing it yourself?**

I (mostly) already have :)

PR #338, opened 2023-06-27
Feature/ppcvletoolchain

paulnoalhyt:

Support for PPC VLE in patchmaker and Ghidra.

**Link to Related Issue(s)**
https://github.com/redballoonsecurity/ofrak/issues/238

This pull request's goal is to add support for PPC VLE in different OFRAK components:
- definition of the VLE sub-isa
- Installation of the toolchain in the patchmaker Dockerfile. Support of this toolchain in the patchmaker and automated tests for it
- Support for VLE analysis in Ghidra and fixups in the instruction unpacker

@rbs-jacob

### DIFF
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/PreAnalyzePPCVLE.java
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/PreAnalyzePPCVLE.java
@@ -1,0 +1,51 @@
+import ghidra.app.script.GhidraScript;
+import ghidra.program.model.mem.*;
+import ghidra.program.model.lang.*;
+import ghidra.program.model.pcode.*;
+import ghidra.program.model.util.*;
+import ghidra.program.model.reloc.*;
+import ghidra.program.model.data.*;
+import ghidra.program.model.block.*;
+import ghidra.program.model.symbol.*;
+import ghidra.program.model.scalar.*;
+import ghidra.program.model.listing.*;
+import ghidra.program.model.address.*;
+
+import java.math.BigInteger;
+
+public class PreAnalyzePPCVLE extends GhidraScript {
+
+    @Override
+    public void run() throws Exception {
+        try {
+            // Set the language to PPC VLE
+            Language language = (Language) getLanguage(new LanguageID("PowerPC:BE:64:VLE-32addr"));
+            Program p = currentProgram;
+            p.setLanguage(language, language.getDefaultCompilerSpec().getCompilerSpecID(), false, monitor);
+            ProgramContext programContext = p.getProgramContext();
+            // Set the vle bit (Ghidra has a "vle" register for that, but on real devices, the VLE
+            // bit is defined per memory page, as a page attribute bit) so that instructions are
+            // decoded correctly.
+            for (Register register : programContext.getContextRegisters()) {
+                if (register.getName().equals("vle")){
+                    RegisterValue newValue = new RegisterValue(programContext.getBaseContextRegister());
+                    BigInteger value = BigInteger.ONE;
+                    newValue = setRegisterValue(newValue, register, value);
+                    programContext.setDefaultDisassemblyContext(newValue);
+                    println("Set the vle bit.");
+                }
+            }
+        } catch(Exception e) {
+            println(e.toString());
+            e.printStackTrace(System.out);
+            throw e;
+        }
+    }
+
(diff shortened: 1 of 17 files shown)


rbs-jacob:
Instead of #341 for allowing secrets (which may or may not work), maybe we can just mark the PPC VLE stuff with `# pragma: no cover` and not be blocked on that? 

Test coverage is all well and good, but I don't know how much we want to block on it for this. There will still be test coverage after merge on the `master` branch (where secrets will work), just not on PR branches from external contributor forks.

ofrak_core/ofrak/core/elf/model.py
```diff
@@ -863,5 +863,28 @@ async def get_program_header_by_index(self, index: int) -> ElfProgramHeader:
...
+                    ppc_vle = True
+                    break
+            if not ppc_vle:
+                section_headers = await self.get_section_headers()
+                for section_header in section_headers:
+                    if section_header.sh_flags & SHF_PPC_VLE != 0:
+                        ppc_vle = True
+                        break
+            if ppc_vle:
+                return SubInstructionSet.PPCVLE
```
EdwardLarson, line 886: I'm not 100% sold on this representation of VLE. We already have it as a "mode" similar to Thumb. Similar to Thumb, the whole binary is not necessarily VLE. More info here: https://www.nxp.com/docs/en/engineering-bulletin/EB687.pdf

> The VLE extension may be used globally within an application, or applied only to specific sections of the application. The e200z0 core is an exception, it supports only the VLE instruction set. 

We could consider it as both a "mode" and a sub-isa because it sort of acts like both. But maybe we should only consider the binary to have the sub-isa of "PPCVLE" if ALL of the (executable) sections/segments are VLE. Otherwise, it's normal PPC, with some areas that are PPC VLE.

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/PreAnalyzePPCVLE.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public void run() throws Exception {
+        try {
+            // Set the language to PPC VLE
+            Language language = (Language) getLanguage(new LanguageID("PowerPC:BE:64:VLE-32addr"));
+            Program p = currentProgram;
+            p.setLanguage(language, language.getDefaultCompilerSpec().getCompilerSpecID(), false, monitor);
+            ProgramContext programContext = p.getProgramContext();
+            // Set the vle bit (Ghidra has a "vle" register for that, but on real devices, the VLE
```
EdwardLarson, line 26: Do you know why Ghidra isn't able to set this bit properly? The infrastructure to handle VLE in this PR only deals with ELFs, which Ghidra should be able to read that bit flag from just fine.

ofrak_core/ofrak/core/elf/model.py
paulnoalhyt: I chose to use a `SubInstructionSet` for several reasons:
- I've never seen an actual target running both VLE and normal PPC instructions. Even though, indeed, the mode can be set per page of memory (https://www.nxp.com/docs/en/supporting-information/VLEPIM.pdf):
    > This alternate encoding set is selected on an instruction page basis. A single page attribute bit selects between standard PowerPC Book E instruction encodings and the VLE instructions for the particular page of memory. This page attribute is an extension to the existing PowerPC Book E page attributes. Pages can be freely intermixed, allowing for a mixture of code with both types of encodings
- tagging a whole executable as PPC VLE, we can force Ghidra to analyze everything as VLE instructions (it has a tendency to prefer decoding normal PPC instructions when it can)
- even on a target whare all instructions are VLE, not all executable sections and segments would have this VLE flag

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/PreAnalyzePPCVLE.java
paulnoalhyt: No, I'm not sure why. Ghidra and IDA both default to decoding instructions as normal PPC, which leads to a lot of wrong analysis. That's why I went with setting this bit before any analysis starts. That way Ghidra will decode all instructions as VLE.

ofrak_core/ofrak/core/elf/model.py
paulnoalhyt: I've only seen ELF files where all code sections are PPC VLE. But I think that in theory an ELF file could contain both PPC and PPC VLE code.

PR #345, opened 2023-07-03
Merged gui searches

EdwardLarson:

Add a search bar to the ResourceTreeView to filter resources by string and bytes.
Add searchbar to HexView in order to search for a string or bytes within a resource.

**Link to Related Issue(s)**
Combines #336 and #334

Adds a Svelte implementation a search bar with two usages in the GUI: One to search the resource tree for all resources containing some bytes/string, and one to search the currently selected resource for bytes/string.

For more details, see the individual PRs.

@dannyp303 
@rbs-jacob


rbs-jacob (changes requested):
Some changes requested. Also, as I type, I repeatedly get some errors in the console that should probably be handled silently.

image

ofrak_core/ofrak/gui/server.py
```diff
@@ -983,15 +982,14 @@ async def get_tags_and_num_components(self, request: Request):
...
     async def search_data(self, request: Request) -> Response:
         resource: Resource = await self._get_resource_for_request(request)
-        mode = request.query.get("type")
-        regex = request.query.get("regex") == "true"
-        case_ignore = request.query.get("caseignore") == "true"
-
+        body = await request.json()
+        mode = body.get("type")
+        regex = body.get("regex") == "true"
+        case_ignore = body.get("caseignore") == "true"
```
rbs-jacob, line 988: JSON has a boolean type. These should be serialized as booleans, no? I suspect that either the values are being serialized wrong on the front end, or these checks will never be true.

frontend/src/ofrak/remote_resource.js
```diff
@@ -646,24 +646,26 @@ export class RemoteResource extends Resource {
...
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({
+        search_query: query,
+        type: options.searchType,
+        regex: options.regex,
+        case_ignore: options.caseIgnore,
+      }),
```
rbs-jacob, line 659: Unless there are other options that you don't want to pass, you can do the following. This will change the snake_case values the server is expecting to receive to camelCase ones, though. 

```suggestion
      body: JSON.stringify({
        search_query: query,
        ...options,
      }),
```

It has the effect of making this easier to change in the future – options don't need to be changed in this function anymore, only at the spots where they are passed in.

frontend/src/ofrak/remote_resource.js
```diff
@@ -674,8 +676,8 @@ export class RemoteResource extends Resource {
       },
       body: JSON.stringify({
         search_query: searchQuery,
-        regex: regex,
-        case_ignore: caseIgnore,
+        regex: options.regex,
+        case_ignore: options.caseIgnore,
```
rbs-jacob, line 680: Same as above – `...options`. Applies to changes below in this file as well.

rbs-jacob (changes requested):
Two very minor fixes, and this is good to go! We're almost there!

frontend/src/ofrak/remote_resource.js
dannyp303: love it

ofrak_core/ofrak/gui/server.py
```diff
@@ -916,6 +979,43 @@ async def get_tags_and_num_components(self, request: Request):
                 all_resource_tags_l.remove(resource_tag)
         return json_response(all_resource_tags_l)
 
+    @exceptions_to_http(SerializedError)
+    async def search_data(self, request: Request) -> Response:
+        resource: Resource = await self._get_resource_for_request(request)
+        body = await request.json()
+        mode = body.get("type")
```
rbs-jacob, line 986: Anywhere in the back end that it gets `type` from `options` needs to be changed to `searchType`, or `searchType` needs to be renamed to `type` in the front end. Either way, the `...options` change made it so the JSON passed to the back end uses all of the exact names from the front end. 

```suggestion
        mode = body.get("searchType")
```

rbs-jacob (changes requested):

rbs-jacob (approved):

PR #349, opened 2023-07-24
Intel Hex format unpacker/packer

EdwardLarson:

Add Identifier, Unpacker, Packer for Intel Hex format

**Link to Related Issue(s)**

### DIFF
--- a/ofrak_core/ofrak/core/ihex.py
+++ b/ofrak_core/ofrak/core/ihex.py
@@ -4,7 +4,7 @@ import sys
 from dataclasses import dataclass
 from typing import List, Union, Tuple
 
-from bincopy import BinFile  # type:ignore
+from bincopy import BinFile
 
 from ofrak.component.analyzer import Analyzer
 from ofrak.component.identifier import Identifier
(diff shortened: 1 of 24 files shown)


ofrak_core/ofrak/core/ihex.py
```diff
@@ -0,0 +1,146 @@
+import logging
+import re
+import sys
+from dataclasses import dataclass
+from typing import List, Union, Tuple
+
+from bincopy import BinFile  # type:ignore
```
rbs-jacob, line 7: Out of curiosity, why does this need a type ignore? I haven't noticed imports causing mypy errors before.

ofrak_core/ofrak/core/ihex.py
```diff
@@ -0,0 +1,146 @@
...
+    async def unpack(self, resource: Resource, config=None):
+        if resource.has_tag(Ihex):
+            ihex_program, binfile = _binfile_analysis(await resource.get_data())
+
+            await resource.create_child_from_view(ihex_program, data=bytes(binfile.as_binary()))
+
+        elif resource.has_tag(IhexProgram):
+            ihex_program = await resource.view_as(IhexProgram)
+            for seg in ihex_program.segments:
+                segment_data_range = seg.translate(-ihex_program.address_limits.start)
```
rbs-jacob, line 74: Why this translation? An explanatory comment might be helpful.

rbs-jacob (changes requested):
It probably makes sense to change the packer and unpacker to become two packers and two unpackers. 

Right now both have two distinct code paths that case on tags. This is exactly what separate components are for, and I don't see the benefit to having those if statements each be inside one packer/unpacker instead of having the cases broken up across different packers and unpackers.

ofrak_core/ofrak/core/ihex.py
EdwardLarson: It was to deal with the mypy error about not being able to type check it:
> ofrak/core/ihex.py:7: error: Skipping analyzing "bincopy": module is installed, but missing library stubs or py.typed marker

The way we usually deal with them is adding an ignore for that module in the `mypy.ini` file. Updated.

rbs-jacob (approved):

PR #354, opened 2023-07-28
Add styling for disabled buttons

EdwardLarson:

Add styling for buttons in the GUI when they are disabled.

**Link to Related Issue(s)**

In order to do this, I at first added new global style vars calculated by blending the background and foreground colors, and were part of the settings structure. All that ended up being unnecessary, but it did lead to a small improvement to the settings loading that handles future, new settings which might not be in the user's cached settings.

@rbs-jacob

### DIFF
--- a/frontend/src/StartView.svelte
+++ b/frontend/src/StartView.svelte
@@ -66,8 +66,7 @@
   }
 
   button:disabled {
-    background-color: var(--faded-bg-color);
-    color: var(--faded-fg-color);
+    opacity: 0.45;
   }
 
   option {
--- a/frontend/src/StartView.svelte
+++ b/frontend/src/StartView.svelte
@@ -55,14 +55,19 @@
     box-shadow: none;
   }
 
-  button:hover,
-  button:focus {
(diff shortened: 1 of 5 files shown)


frontend/src/StartView.svelte
```diff
@@ -55,16 +55,20 @@
...
     box-shadow: inset 1px 1px 0, inset -1px -1px 0;
   }
 
-  button:active {
+  button:active:not([disabled]) {
     box-shadow: inset 2px 2px 0, inset -2px -2px 0;
   }
 
+  button:disabled {
+    opacity: 0.45;
```
rbs-jacob, line 69: <https://developer.mozilla.org/en-US/docs/Web/CSS/cursor>

```suggestion
    opacity: 0.45;
    cursor: not-allowed;
```

rbs-jacob (changes requested):
It looks good! 

One small suggestion is to change the cursor for disabled buttons. Otherwise, this is good to go.

rbs-jacob (approved):

PR #355, opened 2023-08-01
Save resources affected by data patches and dependency updates on a re…

EdwardLarson:
…source being saved

Save resources affected by data patches and dependency updates on a resource being saved

**Link to Related Issue(s)**
Calling `resource.save()` will apply data patches, which for mapped resources will also modify one or more ancestors. Without these changes, those resources would be marked as "dirty" after the original `save` call. This has the confusing result of making local state overall "dirtier" after a `save` call, rather than less.

With these changes, if `save` call on one resource modifies other resources, those resources' models will also be pushed and not dirty.

@dannyp303


rbs-jacob (approved):
The changes here make sense to me individually, but I don't totally understand how they come together to solve the original bug. Still, as far as I can tell, this seems fine.

PR #356, opened 2023-08-07
add checkbox for experimental feature to landing page

EdwardLarson:

**Link to Related Issue(s)**

### DIFF
--- a/frontend/src/StartView.svelte
+++ b/frontend/src/StartView.svelte
@@ -98,6 +98,17 @@
   .clickable {
     cursor: pointer;
   }
+  .experiment-features-check {
+    position: fixed;
+    top: 3em;
+    left: 3em;
+  }
+
+  .experiment-features-check {
+    position: fixed;
+    top: 3em;
+    left: 3em;
+  }
 </style>
 
 <script>
@@ -111,7 +122,8 @@
   import { remote_model_to_resource } from "./ofrak/remote_resource";
 
   import { onMount } from "svelte";
-  import { numBytesToQuantity } from "./helpers";
+  import { numBytesToQuantity, saveSettings } from "./helpers";
+  import Checkbox from "./Checkbox.svelte";
 
   export let rootResourceLoadPromise,
     showRootResource,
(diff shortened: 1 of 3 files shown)


frontend/src/StartView.svelte
```diff
@@ -98,6 +98,12 @@
   .clickable {
     cursor: pointer;
   }
+  .experiment-features-check {
+    position: fixed;
+    top: 3em;
+    left: 3em;
+  }
```
rbs-jacob, line 105: I don't really understand why position in the top left like this? Feels kind of misaligned/out of place with the rest of the interface.

frontend/src/StartView.svelte
```diff
@@ -279,6 +289,13 @@
 </script>
 
 {#if !tryHash}
+  <div class="experiment-features-check">
+    <Checkbox
+      bind:checked="{$settings.experimentalFeatures}"
+      leftbox="{true}"
+      nomargin="{true}">Enable Experimental OFRAK Features</Checkbox
+    >
```
rbs-jacob, line 297: Perhaps set `style:color="currentColor"` or something like that? Right now the other colors change when the mouse is moved, but the color of the checkbox and its label do not.

rbs-jacob (changes requested):
Two minor questions/comments we should resolve before merging.

frontend/src/StartView.svelte
EdwardLarson: I felt it shouldn't be literally front & center, so I wanted it out of the way in the corner. The bottom corners were tricky because the animal popups would overlap it and I was trying to get this done quick so I didn't bother with working through that, and just chose a top corner. 

If you think there's a better spot on the page we can quickly move this to, then let's do it. But if it's not perfect, I plan to iterate on this anyway with a button to access the full settings panel from the StartView, so this checkbox is "temporary" (though not with a definite lifetime).

rbs-jacob: Given that this is temporary, I think it is okay. I would otherwise work to find a better spot if it was going to be in the UI long-term. 

Probably worth adding a comment noting that this is a short-term thing, though. Don't want it to slip through the cracks on the project management side and not have that noted somewhere in the code.

PR #364, opened 2023-08-22
Bugfix: ghidra binja reliability

EdwardLarson:

Fix shell expansion bug with headless Ghidra invocation, and make sure Binja installation script uses Python3 

**Link to Related Issue(s)**

### DIFF
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/GetCodeRegions.java
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/GetCodeRegions.java
@@ -82,7 +82,7 @@ public class GetCodeRegions extends HeadlessScript {
         }
 
         String toJson() {
-            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name);
+            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name.replace("\"", "\\\""));
         }
     }
 }
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
@@ -82,7 +82,7 @@ public class GetCodeRegionsDuplicate extends HeadlessScript {
         }
 
         String toJson() {
-            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name);
+            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name.replace("\"", "\\\""));
         }
     }
 }
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
@@ -1,0 +1,88 @@
+import com.google.common.base.Strings;
+import ghidra.app.util.headless.HeadlessScript;
+import ghidra.app.script.GhidraState;
+import ghidra.program.model.address.Address;
+import ghidra.program.model.address.AddressRange;
+import ghidra.program.model.address.AddressRangeIterator;
+import ghidra.program.model.address.AddressSet;
+import ghidra.program.model.block.BasicBlockModel;
+import ghidra.program.model.block.CodeBlock;
+import ghidra.program.model.block.CodeBlockIterator;
+import ghidra.program.model.block.CodeBlockModel;
+import ghidra.program.model.listing.*;
+import ghidra.program.model.mem.Memory;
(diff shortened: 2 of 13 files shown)


disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
```diff
@@ -0,0 +1,88 @@
...
+import java.math.BigInteger;
+
+public class GetCodeRegionsDuplicate extends HeadlessScript {
+    @Override
+    public void run() throws Exception {
+        try{
+            String response = new GetCodeRegionsDuplicate.Result().toJson();
+            storeHeadlessValue("OfrakResult_GetCodeRegions", response);
+        } catch (Exception e){
+            println(e.toString());
```
rbs-jacob, line 39: Does this `println` actually work? I remember that in Java you usually need `System.out.println`, but I don't see where that is imported?

disassemblers/ofrak_ghidra/ofrak_ghidra_test/test_additional_script_dirs.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Test that the
+    [GhidraProject][ofrak_ghidra.components.ghidra_analyzer.GhidraProject] object can
+    be successfully generated
+    """
+    await hello_world_elf_resource.identify()
+    ghidra_project = await hello_world_elf_resource.view_as(GhidraProject)
+    assert isinstance(ghidra_project, GhidraProject)
+
+    await hello_world_elf_resource.run(example_ghidra_package.GhidraExampleComponent, None)
```
rbs-jacob, line 23: Is it worth asserting anything about the resulting analysis? 

That way this test will confirm not only that it runs successfully, but also that the information is propagated back to OFRAK.

disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
```diff
@@ -0,0 +1,88 @@
...
+        final String name;
+
+        ResultCodeRegion(MemoryBlock block) throws CancelledException {
+            this.start = block.getStart().getOffset();
+            this.size = block.getSize();
+            this.name = block.getName();
+        }
+
+        String toJson() {
+            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name);
```
rbs-jacob, line 85: Hopefully the name doesn't contain any double quotes!

For this example, probably no action necessary. But elsewhere in our Ghidra scripts, I know that there are instances of very jankily-made JSON (with no string escaping) that may lead to bugs down the line.

rbs-jacob (changes requested):

disassemblers/ofrak_ghidra/ofrak_ghidra_test/test_additional_script_dirs.py
EdwardLarson: I don't think so, that main thing the test is addressing is that the the headless Ghidra invocation succeeds, the server starts, and the process stays alive as the server waits for connections/analysis requests. Without escaping the semicolon joining script directories, this doesn't happen, and the test fails long before it gets to run a script. The script itself and any data returned doesn't matter, as long as OFRAK can get to the point of successfully sending a request to the server within the headless instance.

disassemblers/ofrak_ghidra/ofrak_ghidra_test/assets/example_ghidra_package/ghidra_scripts/GetCodeRegionsDuplicate.java
EdwardLarson: `println` is needed to actually write the message to the Ghidra logs, and is imported along with the rest of the standard Ghidra API implicitly when the Ghidra script is built. The normal java way to stdout seems to get swallowed by the void.

rbs-jacob (approved):

rbs-jacob (approved):

disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/GetCodeRegions.java
```diff
@@ -82,7 +82,7 @@ class ResultCodeRegion {
         }
 
         String toJson() {
-            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name);
+            return String.format("{\"start\":%d,\"size\":%d,\"name\":\"%s\"}", start, size, name.replace("\"", "\\\""));
```
rbs-jacob, line 85: For arbitrary strings, you'd also need to at least escape `\`. Because otherwise I could set `name = 'start\\", ::::'` to make this generate malformed JSON. 

That doesn't warrant changing this PR, though, since all of these names are coming directly from Ghidra, and can't contain slashes or quotes anyway (as far as I can tell)!

EdwardLarson: Yeah, I knew that I wasn't really implementing full input sanitization and with backslashes in particular. It does seem unlikely to run into problems with these names though (section names and function names) as it would be highly unusual for these to include anything but alphanumeric characters plus underscores/dashes.

PR #183, opened 2023-01-20
Release Candidate: ofrak 2.1.0

EdwardLarson:
**One sentence summary of this PR (This should go in the CHANGELOG!)**
Release ofrak 2.1.0, ofrak_patch_maker 3.0.0, and ofrak_type 2.1.0

**Link to Related Issue(s)**

**Please describe the changes in your request.**

**Anyone you think should look at this, specifically?**

PR #376, opened 2023-09-13
Maintenance: Install python-lzo from pip

EdwardLarson:

Move/"promote" `python-lzo` module from an external dependency of the Ubi and Ubifs components to a normal pip dependency

**Link to Related Issue(s)**
Not sure why `python-lzo` was marked as an external dependency originally, but it seems erroneous. It of course adds an extra dependency not installed as part of a pip install.

Remove that dependency object and installation line in Dockerstub, add it to requirements.txt.
Also make a small change to one of the ofrak server tests, that sometimes causes a problem ([this error](https://stackoverflow.com/questions/48840378/python-attempt-to-decode-json-with-unexpected-mimetype)).

### DIFF
--- a/ofrak_core/Dockerstub
+++ b/ofrak_core/Dockerstub
@@ -19,9 +19,6 @@ RUN apt-get -y update && \
       unar \
       zstd
 
-# python-lzo needed by ubireader
-RUN python3 -m pip install python-lzo
-
 # Install apktool and uber-apk-signer
 RUN apt-get -y update && apt-get -y install openjdk-11-jdk
 RUN wget https://raw.githubusercontent.com/iBotPeaches/Apktool/v2.3.3/scripts/linux/apktool -O /usr/local/bin/apktool && \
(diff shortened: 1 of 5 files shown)


rbs-jacob:
I don't know if this is why, but [`vmlinux-to-elf` pins a specific Git commit of `python-lzo`](https://github.com/marin-m/vmlinux-to-elf/blob/fa5c9305ae1c4bbcd2debabb810e7613def690a7/setup.py#L5-L15). Just FYI.

whyitfor:
@EdwardLarson , IIRC, the issue is that:

- `python-lzo` is a dependency of `ubireader`, which is not pip-installed
- `python-lzo` requires certain headers to be installed 

If this is true, then the changes in the PR could make the pip install of ofrak possibly fail on systems without these headers.

Since `python-lzo` is used directly by `ubireader`, and not used by OFRAK, I believe you decided in #183 to install it this way.

Issue #389, opened 2023-10-18
java is_tool_installed() not working correctly.


sprout42:
**What is the problem? (Here is where you provide a complete Traceback.)**
The dependency check for `java` using the `--help` argument does not result in a successful return code which prevents a properly installed `java` from being detected.
```
$ ofrak deps --missing-only
[ ] apktool
        https://ibotpeaches.github.io/Apktool/
        [ApkPacker, ApkUnpacker]
[ ] binwalk
        https://github.com/ReFirmLabs/binwalk
        [BinwalkAnalyzer]
[ ] java
        https://openjdk.org/projects/jdk/11/
        [ApkPacker]
[ ] python-lzo
        https://github.com/jd-boyd/python-lzo
        [UbiPacker, UbiIdentifier, UbifsUnpacker, UbiUnpacker, UbiAnalyzer, UbifsIdentifier, UbifsAnalyzer]

$ java --help
Unrecognized option: --help
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

**Please provide some information about your environment.**
- Ubuntu 22.04.03 LTS
```
$ python3 -m platform
Linux-5.15.0-83-generic-x86_64-with-glibc2.35

$ python3 -c "import sys;print(sys.executable)"
/usr/bin/python3
```
- `openjdk-11-jdk` installed through `apt`
- `ofrak` installed through `pip` in a clean virtual environment
```
python -m pip freeze
aiohttp==3.8.6
aiohttp-cors==0.7.0
aiosignal==1.3.1
async-timeout==4.0.3
attrs==23.1.0
beartype==0.12.0
black==23.3.0
cffi==1.16.0
charset-normalizer==3.3.0
click==8.1.7
cstruct==5.2
fdt==0.3.3
frozenlist==1.4.0
gitdb==4.0.10
GitPython==3.1.32
idna==3.4
immutabledict==2.2.0
importlib-metadata==6.8.0
intervaltree==3.1.0
jefferson==0.4.5
keystone-engine==0.9.2
lief==0.12.3
lzallright==0.2.3
multidict==6.0.4
mypy-extensions==1.0.0
ofrak==3.2.0.post0
ofrak-io==1.1.0
ofrak-patch-maker==4.0.2
ofrak-type==2.2.0
orjson==3.8.14
packaging==23.2
pathspec==0.11.2
pefile==2023.2.7
platformdirs==3.11.0
pycdlib==1.12.0
[...]

PR #390, opened 2023-10-18
Fix apktool and java CLI component check errors

### LINKED ISSUE #388: apktool is_tool_installed() not working correctly.
The dependency check for `apktool` using the `--help` argument does not result in a successful return code which prevents a properly installed `apktool` from being detected.
```
$ ofrak deps --missing-only
[ ] apktool
        https://ibotpeaches.github.io/Apktool/
        [ApkPacker, ApkUnpacker]
[ ] binwalk
        https://github.com/ReFirmLabs/binwalk
        [BinwalkAnalyzer]
[ ] java
        https://openjdk.org/projects/jdk/11/
        [ApkPacker]
[ ] python-lzo
        https://github.com/jd-boyd/python-lzo
        [UbiPacker, UbiIdentifier, UbifsUnpacker, UbiUnpacker, UbiAnalyzer, UbifsIdentifier, UbifsAnalyzer]

$ apktool --help
Unrecognized option: --help
Apktool v2.5.0-dirty - a tool for reengineering Android apk files
with smali v2.4.0-debian and baksmali v2.4.0-debian
Copyright 2010 Ryszard Wiśniewski <brut.alll@gmail.com>
Copyright 2010 Connor Tumbleson <connor.tumbleson@gmail.com>

usage: apktool
 -advance,--advanced   prints advance information.
 -version,--version    prints the version then exits
usage: apktool if|install-framework [options] <framework.apk>
 -p,--frame-path <dir>   Stores framework files into <dir>.
 -t,--tag <tag>          Tag frameworks using <tag>.
usage: apktool d[ecode] [options] <file_apk>
 -f,--force              Force delete destination directory.
 -o,--output <dir>       The name of folder that gets written. Default is apk.out
 -p,--frame-path <dir>   Uses framework files located in <dir>.
 -r,--no-res             Do not decode resources.
 -s,--no-src             Do not decode sources.
 -t,--frame-tag <tag>    Uses framework files tagged by <tag>.
usage: apktool b[uild] [options] <app_path>
 -f,--force-all          Skip changes detection and build all files.
 -o,--output <dir>       The name of apk that gets written. Default is dist/name.apk
 -p,--frame-path <dir>   Uses framework files located in <dir>.

For additional info, see: https://ibotpeaches.github.io/Apktool/ 
For smali/baksmali info, see: https://github.com/JesusFreke/smali
```

**Please provide some information about your environment.**
- Ubuntu 22.04.03 LTS
```
$ python3 -m platform
Linux-5.15.0-83-generic-x86_64-with-glibc2.35

$ python3 -c "import sys;print(sys.executable)"
/usr/bin/python3
```
- `apktool` installed through `apt`
- `ofrak` installed through `pip` in a clean virtual environment
```
python -m pip freeze
aiohttp==3.8.6
aiohttp-cors==0.7.0
aiosignal==1.3.1
async-timeout==4.0.3
attrs==23.1.0
beartype==0.12.0
black==23.3.0
cffi==1.16.0
charset-normalizer==3.3.0
click==8.1.7
cstruct==5.2
fdt==0.3.3
frozenlist==1.4.0
gitdb==4.0.10
GitPython==3.1.32
idna==3.4
immutabledict==2.2.0
importlib-metadata==6.8.0
intervaltree==3.1.0
jefferson==0.4.5
keystone-engine==0.9.2
lief==0.12.3
lzallright==0.2.3
multidict==6.0.4
mypy-extensions==1.0.0
ofrak==3.2.0.post0
ofrak-io==1.1.0
ofrak-patch-maker==4.0.2
ofrak-type==2.2.0
orjson==3.8.14
packaging==23.2
pathspec==0.11.2
pefile==2023.2.7
platformdirs==3.11.0
pycdlib==1.12.0
pycparser==2.21
python-magic==0.4.27
reedsolo==1.7.0
smmap==5.0.1
sortedcontainers==2.2.2
synthol==0.1.1
tomli==2.0.1
typeguard==2.13.3
typing-inspect==0.7.1
typing_extensions==4.8.0
ubi-reader==0.8.5
xattr==0.10.1
yarl==1.9.2
zipp==3.17.0
```

The `--help` option provided as the `install_check_arg` isn't supported by `apktool`resulting in a return code of 1 from the `apktool`
https://github.com/redballoonsecurity/ofrak/blob/6aa48f99a63a0a5f8fd37101f16d1b38172a1305/ofrak_core/ofrak/core/apk.py#L25C86-L25C94
```
$ apktool --help > /dev/null
Unrecognized option: --help
$ echo $?
1
```
If the tool is run with no arguments it will return a success value of 0
```
$ apktool > /dev/null
$ echo $?
0
```

every time

1. install `apktool` through apt
2. install `ofrak`
3. run `ofrak deps --missing-only` 

change the `"--help"` `install_check_arg` to an empty string:
```
APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/", "")
```

This seems pretty simple and straight forward.

Not at the moment.

rbs-jacob:

Fix `java` and `apktool` CLI arguments for checking components.

Note that the argument `-help` passed to the `java` command seems maximally portable based on the following test ([copied from the issue discussion here](https://github.com/redballoonsecurity/ofrak/issues/389#issuecomment-1769142342)).

``` bash
for CMD in -h --help -help; do 
  echo "${CMD}"
  for JAVA in /usr/lib/jvm/*/bin/java; do 
    "${JAVA}" "${CMD}" >/dev/null 2>&1 \
      && printf '\t\033[1;32mWorked\033[1;m  ' \
      || printf '\t\033[1;31mFailed\033[1;m  '
    echo "${JAVA}"
  done
done
```

image

**Link to Related Issue(s)**

Fix #388 and #389.

### DIFF
--- a/ofrak_core/ofrak/core/apk.py
+++ b/ofrak_core/ofrak/core/apk.py
@@ -22,7 +22,7 @@ from ofrak.core.magic import Magic, MagicMimeIdentifier
 from ofrak_type.range import Range
 
 
-APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/")
+APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/", "-version")
 JAVA = ComponentExternalTool(
     "java",
     "https://openjdk.org/projects/jdk/11/",
--- a/ofrak_core/ofrak/core/apk.py
+++ b/ofrak_core/ofrak/core/apk.py
@@ -22,11 +22,11 @@ from ofrak.core.magic import Magic, MagicMimeIdentifier
 from ofrak_type.range import Range
 
 
-APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/", "--help")
+APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/")
 JAVA = ComponentExternalTool(
     "java",
     "https://openjdk.org/projects/jdk/11/",
-    "--help",
+    "-help",
     apt_package="openjdk-11-jdk",
     brew_package="openjdk@11",
 )
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
 ### Fixed
 - Improved flushing of filesystem entries (including symbolic links and other types) to disk. ([#373](https://github.com/redballoonsecurity/ofrak/pull/373))
+- Fix `java` and `apktool` CLI arguments for checking components. ([#390](https://github.com/redballoonsecurity/ofrak/pull/390))
 
 ### Changed
 - `Resource.flush_to_disk` method renamed to `Resource.flush_data_to_disk`. ([#373](https://github.com/redballoonsecurity/ofrak/pull/373))
--- a/ofrak_core/ofrak/core/strings_analyzer.py
+++ b/ofrak_core/ofrak/core/strings_analyzer.py
@@ -31,9 +31,10 @@ class _StringsToolDependency(ComponentExternalTool):
 
     async def is_tool_installed(self) -> bool:
         try:
-            cmd = [self.tool]
-            if self.install_check_arg is not None:
-                cmd.append(self.install_check_arg)
+            cmd = [
+                self.tool,
+                self.install_check_arg,
+            ]
             proc = await asyncio.create_subprocess_exec(
                 *cmd,
                 stdout=asyncio.subprocess.DEVNULL,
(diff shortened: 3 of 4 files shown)


ofrak_core/ofrak/core/apk.py
```diff
@@ -22,11 +22,11 @@
 from ofrak_type.range import Range
 
 
-APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/", "--help")
+APKTOOL = ComponentExternalTool("apktool", "https://ibotpeaches.github.io/Apktool/")
```
whyitfor, line 25: #388 suggests using "" as the argument.

Is there any reason you decided against this? It seems preferable, as it doesn't change the signature of ComponentExternalTool.

ANogin: Could probably also use `"-version"`

rbs-jacob: Even though the empty string works here (I did try it), not all CLI applications will ignore or properly handle an empty string argument. It feels more correct to me that if we don't want to pass an additional argument, we use `None` rather than the empty string. 

In this case, I wasn't worried about changing the signature since the same parameters are still accepted, albeit one is now a keyword argument. As a compromise, I could make that parameter typed as `Optional`, but not be a keyword argument with a default value. That way the signature changes even less, but still enables what I consider to be more robust, correct behavior. 

Thoughts @whyitfor?

whyitfor (approved):
LGTM

PR #392, opened 2023-11-10
build_image.py: Auto-enable Docker buildkit, and make git optional

EdwardLarson:

Allow Docker build to continue even if there is a problem getting local git history, and default to forcing Docker buildkit to be enabled.

**Link to Related Issue(s)**
Currently if there is a problem with local git repo, `build_image.py` will fail and be unable to build. As this isn't strictly necessary to build the image, just to give it a nice tag, the build should be able to continue without it.

Some features in the build require Docker buildkit to be enabled, and we say as much in the docs to build the image. However, this is a point of failure (for some reason I could not get the buildkit to be enabled in my machine) and a "gotcha" that can be removed entirely by just setting the Docker buildkit enabled in the subprocess we spin up to do the Docker build. In case there is a desire to specifically NOT use the buildkit, an option is added to not alter the subprocess environment at all (buildkit may or may not be enabled depending on user's environment).

Sufficiently summarized above.

@whyitfor

### DIFF
--- a/build_image.py
+++ b/build_image.py
@@ -39,6 +39,7 @@ class OfrakImageConfig:
     install_target: InstallTarget
     cache_from: List[str]
     entrypoint: Optional[str]
+    buildkit: bool
 
     def validate_serial_txt_existence(self):
         """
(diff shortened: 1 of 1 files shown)


ANogin:
I am not able to change the source branch for this, so created a new PR #515. This one should be closed.

whyitfor:
Closing as #515 fixes the issue.

PR #405, opened 2024-01-12
Add x86 32-bit GNU toolchain

rbs-jacob:

Add x86 32-bit GNU toolchain

**To Do**

- [x] Add tests

@whyitfor


whyitfor:
I've made some improvements to the X86_32 Toolchain (fixed typo in name, added 4-byte alignment method).

Additionally, the PR has been enhanced to do the following:
- Use default GCC linker, bin-parser, assembler for X86_64 toolchain, instead of compiling from source from binutils. Doing so seems to have no downside, makes the build faster, and makes it easier to support running OFRAK on AARCH64 machines 
- Fix bug which resulted in X86_64 Toolchain not being built for AARCH64 hosted Docker
- Update PatchMaker tests to use `relocatable=False`. This is needed b/c the X86_32 toolchain, when `relocatable=True`, creates `..text.__x86.get_pc_thunk.ax` and `.text.__x86.get_pc_thunk.dx:` which the linker tries to put in our resulting text section; this causes problems since the tests do not account for this space.

whyitfor (approved):
LGTM, after getting tests to pass.

PR #409, opened 2024-02-01
Modified the FreeSpaceModifier to mark patch bytes written to the binary as not free

rbs-alexr:

Change `FreeSpaceModifier` & `PartialFreeSpaceModifier` behavior: an optional stub that isn't free space can be provided and fill-bytes for free space can be specified. 

**Link to Related Issue(s)**
N/A.

This changes the signature and intended behavior of the configs for `FreeSpaceModifier` and `PartialFreeSpaceModifier`. Each modifier now allows for injecting a "stub" at the beginning of the targeted code to be freed. This stub is not included in the `FreeSpace`. The config's "fill" parameter is now the data pattern used to overwrite the bytes that fall in the `FreeSpace`.

@rbs-afflitto, @rbs-alexr.
Optionally: @rbs-jacob, @EdwardLarson .

### DIFF
--- a/ofrak_core/ofrak/core/free_space.py
+++ b/ofrak_core/ofrak/core/free_space.py
@@ -450,14 +450,12 @@ async def _find_and_delete_overlapping_children(resource: Resource, freed_range:
         await overlapping_child.resource.save()
 
 
-def _get_fill(freed_range: Range, fill: Optional[bytes]):
-    if not fill:
-        return b"\x00" * freed_range.length()
-    else:
-        diff_len = freed_range.length() - len(fill)
-        if diff_len < 0:
-            raise ValueError("config.fill value cannot be longer than the range to be freed.")
-        return fill + b"\x00" * diff_len
+def _get_patch(freed_range: Range, stub: bytes, fill: bytes) -> bytes:
+    total_fill_length = freed_range.length() - len(stub)
+    remainder = total_fill_length % len(fill)
+    final = b"".join([stub, fill * (total_fill_length // len(fill)), fill[:remainder]])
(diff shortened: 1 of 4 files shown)


whyitfor:
@rbs-afflitto, @rbs-alexr, can you please review these update?

ofrak_core/ofrak/core/free_space.py
```diff
@@ -450,34 +450,37 @@ async def _find_and_delete_overlapping_children(resource: Resource, freed_range:
...
-    if not fill:
-        return b"\x00" * freed_range.length()
-    else:
-        diff_len = freed_range.length() - len(fill)
-        if diff_len < 0:
-            raise ValueError("config.fill value cannot be longer than the range to be freed.")
-        return fill + b"\x00" * diff_len
+def _get_patch(freed_range: Range, stub: bytes, fill: bytes) -> bytes:
+    total_fill_length = freed_range.length() - len(stub)
+    remainder = total_fill_length % len(fill)
```
rbs-afflitto, line 455: This will raise a `ZeroDivisionError` if the user passes in a fill=`b""`. I think we should check for `len(fill) == 0` and either raise a `ValueError` stating that `fill must be at least 1 byte`, or to set fill to `b"\x00"` in that case.

rbs-alexr:
I took a look at Andrew's feedback and your fix, it all looks good!

ofrak_core/ofrak/core/free_space.py
whyitfor: Good point -- I've addressed this in the config and added tests.

rbs-afflitto (approved):
Looks good to me

Issue #410, opened 2024-02-06
ESP Unpacker


Jepson2k:
**What is the use case for the feature?**
To unpack ESP binaries (ESP8266, ESP32, etc.)

**Does the feature contain any proprietary information about another company's intellectual property?**
No its in their documentation, available online.

**How would you implement this feature?**
By referencing the existing examples and following suite.

**Are there any (reasonable) alternative approaches?**
Yes, many.

**Are you interested in implementing it yourself?**
Yes its mostly done but I have a few questions I've outlined in the code comments.

PR #412, opened 2024-02-06
Added ESP Unpacker

Jepson2k:

Added an ESP unpacker to support unpacking Espressif binaries.

**Link to Related Issue(s)**
#410 

Added esp.py to ofrak_core/ofrak/core which contains multiple resource view for ESP binaries as well as identification and unpacking. Uses ESPTool.py for most of the work.

@rbs-jacob

There are multiple "TODO:"s located in the code, a few of which are notes for things I'm unsure about, and the rest are questions that I couldn't find answers to in the contribution guidelines.


ofrak_core/ofrak/core/esp.py
```diff
@@ -0,0 +1,779 @@
...
+## Analyzing a Binary Image
+To analyze a binary image and get a complete summary of its headers and
+segments, use the :ref:`image_info <image-info>` command with the ``--version
+2`` option.
+"""
+
+LOGGER = logging.getLogger(__name__)
+
+ESP_BINARY_MAGIC = 0xE9
+ESP8266V2_BINARY_MAGIC = 0xEA  # In the esptool.py code, but haven't seen it in the documentation
```
dannyp303, line 179: I'm not sure about this, 0xE9 and 0xEA will match a lot of arm instructions, we might just have to tag manually if there isn't a larger magic we can find.

Jepson2k: I was concerned about that as well. I could implement a more heuristic based approach. Byte 2 (zero index) is the flash mode which is always 0,1, 2, or 3 and the lower half of byte three is always 0, 1, 2, or 0xf. If you think that would narrow it down enough I can implement that. If we go the manual tagging route, do I just remove the identify function?

EdwardLarson: The more constraints, the better.

If manual tagging turns out to be the only way to avoid a very high false positive ID rate, then the Identifier component as a whole can be removed.

ofrak_core/ofrak/core/esp.py
```diff
@@ -0,0 +1,779 @@
...
+                attribute_filters=[
+                    ResourceAttributeValueFilter(
+                        ESPSectionStructure.SectionIndex, self.section_index
+                    )
+                ],
+            ),
+        )
+
+
+# TODO: Program or Generic Binary? Not sure...
```
EdwardLarson, line 427: I think `Program` is indeed appropriate here. If it's something can can be broken up into memory sections, it usually makes sense as a `Program`.
Another related consideration is whether it has an entry point - this isn't part of the `Program` at the moment, but it's something I think would make sense to add to the general model of what a Program is. So I'm curious about this case, does ESP32 define and entry point?

ofrak_core/ofrak/core/esp.py
```diff
@@ -0,0 +1,779 @@
...
+        flash_fr_bits = image.flash_size_freq & 0x0F
+        return flash_s_bits, flash_fr_bits
+
+    # TODO: Display boolean or value?
+    def __parse_wp_pin(self, image, chip):
+        if chip != "esp8266":
+            return False if image.wp_pin == image.WP_PIN_DISABLED else True
+        return None
+
+    # TODO: Probably useful to display chip name but not sure where to put it atm
```
EdwardLarson, line 689: For things like this, it makes sense to add attributes to the `ESP` class and split the logic to parse these out into an analyzer. Sometimes there is overlap with the unpacking logic, and that's okay: the unpacker can call the analyzer and the code doesn't need to be duplicated.

ofrak_core/ofrak/core/elf/modifier.py
rbs-jacob, line 1: This change seems... random?

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+    """
+    extended_header = f.read(16)
+    if extended_header[-1] not in [0, 1]:
+        return "esp8266"
+
+    chip_id = int.from_bytes(extended_header[4:5], "little")
+    for rom in [n for n in ROM_LIST if n.CHIP_NAME != "ESP8266"]:
+        if chip_id == rom.IMAGE_CHIP_ID:
+            return rom.CHIP_NAME.lower()
+    return "esp8266"
```
rbs-jacob: For type-checking purposes, this should probably return an enum instead of a string.

rbs-jacob: Also, why use a temporary file wrapper instead of indexing into the data itself? All of the data is already stored in memory.

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+    except IndexError:
+        raise UnpackerError("File is empty")
+    if magic not in [
+        ESPLoader.ESP_IMAGE_MAGIC,
+        ESP8266V2FirmwareImage.IMAGE_V2_MAGIC,
+    ]:
+        raise UnpackerError(
+            "This is not a valid image " "(invalid magic number: {:#x})".format(magic)
+        )
+    return magic
```
rbs-jacob: This function and the one above statefully read bytes from the file handle. Instead, they should statelessly index into a bytes object or bytearray. Moreover, they read several bytes to only look at the first or last one.

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+#    IDENTIFIER    #
+####################
+class ESPAppIdentifier(Identifier):
+    """
+    Identify ESP apps.
+
+    :param targets: A tuple containing the target resource types for identification
+    """
+
+    targets = (File, GenericBinary, Program, ESPFlashSection)
```
rbs-jacob:
```suggestion
    targets = (File, GenericBinary, Program, ESPFlashSection)
```

I think some of these (`File`, at least) already inherit from `GenericBinary`, and don't need to be listed here.

ofrak_core/ofrak/core/esp/app_model.py
rbs-jacob, line 1: Philosophically, there are a lot of child types here that we don't need to define. 

To make an analogy: the OFRAK tar unpacker unpacks files, not tar blocks. We want to do the equivalent here and unpack only what will actually be semantically useful for further layers of unpacking or analysis. We don't necessarily need tagged children for each part of the file type, especially if those parts are metadata.

Unfortunately, this guiding philosophy isn't documented anywhere, so there isn't a way you could have known this before. Also if any of this is based on the Elf unpacker, that one deviates from this philosophy a bit.

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+                version=unpacked_data[3],
+                project_name=unpacked_data[4],
+                time=unpacked_data[5],
+                date=unpacked_data[6],
+                idf_ver=unpacked_data[7],
+                app_eld_sha256=unpacked_data[8],
+                reserv2=unpacked_data[9],
+            ),
+            unpacked_data[0],
+        )
```
rbs-jacob:
```suggestion
        return (
            ESPAppDescription(*unpacked_data),
            unpacked_data[0],
        )
```

Would this work?

rbs-jacob: Same comment applies to other places where struct.unpack output is manually passed to dataclass constructors, including immediately below

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+    """
+
+    targets = (ESPApp,)
+    outputs = (ESPAppAttributes,)
+
+    async def analyze(self, resource: Resource, config: Optional[ESPAppConfig]) -> ESPAppAttributes:
+        config = await get_esp_app(resource, config)
+        try:
+            calculated_checksum = int(config.image.calculate_checksum())
+            checksum_valid = True if config.image.checksum == calculated_checksum else False
```
rbs-jacob:
```suggestion
            checksum_valid = (config.image.checksum == calculated_checksum)
```

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+            hash_valid = False
+            calculated_hash = 0
+            # append_digest is only available on ESP32 images, not ESP8266
+            if hasattr(config.image, 'append_digest') and config.image.append_digest:
+                calculated_hash = int.from_bytes(config.image.calc_digest[:4], 'little')  # Use first 4 bytes as int
+                if config.image.stored_digest == config.image.calc_digest:
+                    hash_valid = True
+                else:
+                    hash_valid = False
```
rbs-jacob:
```suggestion
                hash_valid = config.image.stored_digest == config.image.calc_digest
```

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+            return ESPAppAttributes(
+                chip_name=self.__get_chip_name(config.image),
+                checksum_valid=checksum_valid,
+                calculated_checksum=calculated_checksum,
+                hash_valid=hash_valid,
+                calculated_hash=calculated_hash,
+            )
+        finally:
+            if hasattr(config, "f"):
+                config.f.close()
```
rbs-jacob: I suspect that changing the code to not seek statefully will eliminate the need for this try/finally.

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+
+    def __get_chip_name(self, image):
```
rbs-jacob:
```suggestion
    def _get_chip_name(self, image):
```

This and other internal methods should use single underscores

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+            packed_data = await resource.get_data()
+            
+            with NamedTemporaryFile(suffix=".bin", delete=False) as temp_file:
+                temp_file.write(packed_data)
+                temp_file.flush()
+                
+                # Use esptool to verify the image
+                from esptool import main as esptool_main
+                import sys
+                from io import StringIO
```
rbs-jacob: Can we move these imports to the top?

ofrak_core/ofrak/core/esp/app.py
```diff
@@ -0,0 +1,959 @@
...
+                
+                # Capture esptool output
+                old_stdout = sys.stdout
+                sys.stdout = captured_output = StringIO()
+                
+                try:
+                    # Run esptool image_info command
+                    sys.argv = ['esptool.py', 'image_info', '--version', '2', temp_file.name]
```
rbs-jacob: Is there a reason we have to call it in this process like this? Why can't we do a subprocess call?

ofrak_core/ofrak/core/esp/flash.py
rbs-jacob, line 1: I suspect the linter has not been run on this file (possibly others as well). Please do that

ofrak_core/requirements.txt
```diff
@@ -23,3 +23,4 @@ tempfile312~=1.0.1
 typeguard~=2.13.3
 ubi-reader==0.8.5
 xattr==0.10.1;platform_system!="Windows"
+esptool
```
rbs-jacob: Please pin the version

rbs-jacob (changes requested):
I did not review all of this in great detail, but there are some substantial changes needed before I'd be comfortable merging this in

Jepson2k:
@rbs-jacob I believe I've resolved all prior comments and issues with the latest commits. The macOS reds appear to be related to a no_autobump! infra issue thats affecting the nightly as well.

PR #415, opened 2024-02-08
Feature/angr decompilation

dannyp303:

Added options for different available data views in GUI as well as new decompilation using new angr enabled decompilation analyze.
**Link to Related Issue(s)**


frontend/src/App.svelte
```diff
@@ -211,18 +203,12 @@ Answer by running riddle.answer('your answer here') from the console.`);
...
-          <TextView />
-        {:else}
-          <HexView
-            dataLenPromise="{dataLenPromise}"
-            resources="{resources}"
-            scrollY="{hexScrollY}"
-            bind:resourceNodeDataMap="{resourceNodeDataMap}"
-          />
-        {/if}
+        <DataView
```
EdwardLarson, line 206: Nitpicky, but I feel like maybe `DataView` is a misleading name since "data" in OFRAK usually refers to raw bytes. I suppose the text and assembly views are just another way of decoding that same data, but the decompilation view is further disconnected and I would prefer a more unique name less likely to mislead. Something like `ResourceBodyView` or `ContentsView`, maybe.

frontend/src/views/DataView.svelte
```diff
@@ -0,0 +1,118 @@
...
+  let hasDecompView = false;
+  let dataSearchResults = {};
+  let searchFunction;
+
+  function checkTags() {
+    display_type = "hex";
+    hasTextView = ["ofrak.core.binary.GenericText"].some((tag) =>
+      $selectedResource.has_tag(tag)
+    );
+    hasAsmView = [
```
EdwardLarson, line 47: I think it can be made simpler to add more tab-able views, if all you need to do is add an entry to a table for the new view and what tags it supports (and also implement the view itself, but that's always the case). This is only relevant if you think it's likely that there are more views of this nature we might realistically add soon-ish. The only one I can think of at the moment is a CFG view for `ComplexBlock`s, and I'm not sure how realistic that is. However if, after implementing this tabbing feature, you are bursting with ideas for how we can add to it, then I encourage implementing it with scalability in mind. If not though, then the current implementation is perfectly fine and even if it were totally inconvenient for adding new tabs, it can be changed later, so you can ignore the rest of this comment.

The idea for how to implement this is to replace this logic (and the svelte if/else chain later) with a mapping from each of the possible content views to the group of tags they are applicable for. Then the list of applicable content views is built from checking each of them against the actual resource's tags, the group of "tab" buttons is built from that list with a Svelte `{#each ...}`, and each button's on-click assigns the corresponding content view Svelte component to a common variable, which can be used by Svelte to show the selected view without an if/else chain with:

```
<svelte:component
              this="{selectedContentView}"
              bind:searchFunction="{searchFunction}"
            />
```

(similar to how we use a variable to hold the type of the view in the bottom left pane, and change that variable, rather than switch on the different cases of what it might be)

disassemblers/ofrak_angr/ofrak_angr_test/test_decompilation.py
```diff
@@ -0,0 +1,35 @@
...
+
+    ipdb.set_trace()
+    decomps = []
+    for complex_block in complex_blocks:
+        await complex_block.resource.identify()
+        angr_resource: AngrDecompilationAnalysis = await complex_block.resource.view_as(
+            AngrDecompilationAnalysis
+        )
+        decomps.append(await angr_resource.resource.analyze(AngrDecompilationAnalysis))
+    assert len(decomps) == 11
```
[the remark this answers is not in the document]
dannyp303: Hmm yeah, I had something better here but the white space was causing issues. I think just checking non-empty strings and that names of functions/printed strings are somewhere in there works for me.

frontend/src/App.svelte
dannyp303: ContentsView is good, I didn't think about this name much

frontend/src/views/DataView.svelte
dannyp303: good idea, implemented.

dannyp303:
Split into 2 seperate PRs to address hex view changes, then add decompilatin view.

PR #416, opened 2024-02-09
Use python3.8 in docker images

ANogin:

Update docker configuration to use Python 3.8, rather than 3.7

**Link to Related Issue(s)**
N/A

Switches the docker configuration from the previous pinned python:3.7-bullseye docker version to the latest-as-of-today python:3.8-bullseye one.

@whyitfor

### DIFF
--- a/ofrak_core/Dockerstub
+++ b/ofrak_core/Dockerstub
@@ -61,7 +61,7 @@ RUN cd /tmp && \
 
 # Install Jefferson
 WORKDIR /tmp
-RUN wget https://bootstrap.pypa.io/pip/get-pip.py && python3.9 get-pip.py && python3.7 get-pip.py && rm get-pip.py
+RUN wget https://bootstrap.pypa.io/pip/get-pip.py && python3.9 get-pip.py && python3.8 get-pip.py && rm get-pip.py
 RUN python3.9 -m pip install jefferson
 WORKDIR /
 
--- a/ofrak_type/setup.py
+++ b/ofrak_type/setup.py
@@ -32,7 +32,7 @@ setuptools.setup(
             "fun-coverage==0.2.0",
             "hypothesis~=6.39.3",
             "mypy==0.942",
-            "pytest",
+            "pytest<8.0",
             "pytest-cov",
         ]
     },
--- a/build_image.py
+++ b/build_image.py
@@ -195,7 +195,7 @@ def create_dockerfile_base(config: OfrakImageConfig) -> str:
         dockerfile_base_parts += [f"### {dockerstage_path}", dockerstub]
 
     dockerfile_base_parts += [
-        "FROM python:3.7-bullseye@sha256:338ead05c1a0aa8bd8fcba8e4dbbe2afd0283b4732fd30cf9b3bfcfcbc4affab",
+        "FROM python:3.8-bullseye@sha256:e1cd369204123e89646f8c001db830eddfe3e381bd5c837df00141be3bd754cb",
         "",
     ]
 


ofrak_core/Dockerstub
```diff
@@ -61,7 +61,7 @@ RUN cd /tmp && \
 
 # Install Jefferson
 WORKDIR /tmp
-RUN wget https://bootstrap.pypa.io/pip/get-pip.py && python3.9 get-pip.py && python3.7 get-pip.py && rm get-pip.py
+RUN wget https://bootstrap.pypa.io/pip/get-pip.py && python3.9 get-pip.py && python3.8 get-pip.py && rm get-pip.py
 RUN python3.9 -m pip install jefferson
```
EdwardLarson, line 65: The upgrade to this Python version juxtaposed with the next line makes me think maybe we should just bump to 3.9 while we are at it, and remove the weird alternative python version install. It also gives more of a buffer for the next Python version EoL. What do you think?

ANogin: Hm, might be a reasonable thing to do - let me test and see if anything else breaks from it.

ANogin: @EdwardLarson unfortunately, with Python 3.9, some of the `test_ofrak/service/serialization_service/test_pjson.py` tests fail in `ofrak_core`. Let's maybe do this one for now, and worry about 3.9 later?

ofrak_type/setup.py
```diff
@@ -32,7 +32,7 @@ def run(self):
             "fun-coverage==0.2.0",
             "hypothesis~=6.39.3",
             "mypy==0.942",
-            "pytest",
+            "pytest<8.0",
```
whyitfor, line 35: Do we have an understanding of why this is needed? It's not intuitive why limiting a version is required when bumping the Python version.

whyitfor (commented):
With this change, all of our testing will be done on Python 3.8.

Since python3.7 is [eol](https://devguide.python.org/versions/), we will eventually want to update the `python_requires` for the ofrak setup.py files to alert dropping of 3.7. This isn't strictly needed now, though, and could arguably cause annoyance to users of OFRAK currently on 3.7.

@ANogin, perhaps it makese sense to add to the README or to the documentation a note that 3.7 is generally supported by that all testing is done on 3.8, which is the recommended version?

ofrak_type/setup.py
ANogin: @whyitfor I put it in the log message for that commit - this is needed because otherwise we get 8.0.0 by default and hit https://github.com/pytest-dev/pytest/issues/11890
https://github.com/TvoroG/pytest-lazy-fixture/issues/65, and I did not take time to investigate whether there are any better workarounds. This is not needed with python3.7 because starting with version 8.0.0 (just released Jan 27!), pytest requires python 3.8.

ANogin:
> @ANogin, perhaps it makese sense to add to the README or to the documentation a note that 3.7 is generally supported by that all testing is done on 3.8, which is the recommended version?

There were two places in the documentation that specified the Python version, I've added a note to both.

ofrak_type/setup.py
whyitfor: Great, and now we have that record in the PR as well.

whyitfor (approved):
LGTM.

whyitfor:
@EdwardLarson, do you think this necessitates changelog entires? My read is no: the changelogs seem to describe the Python projects, and not necessarily the Docker build environment.

EdwardLarson (approved):

EdwardLarson:
> @EdwardLarson, do you think this necessitates changelog entires? My read is no: the changelogs seem to describe the Python projects, and not necessarily the Docker build environment.

Nah, I agree with your take on the changelog. This is not a new feature/bugfix.

Issue #419, opened 2024-02-12
Circular dependency cases a `make develop` weirdness.


ANogin:
**What is the problem? (Here is where you provide a complete Traceback.)**

#226 introduced an unfortunate issue (which I discovered using PR #218 - IMHO we really out to land that one, and use in CI) - when the docker build runs `make develop`, it would first go to `ofrak_core`, where the new `"test": "ofrak_angr~=1.0",  "ofrak_capstone~=1.0"]` dependency would cause it to download `ofrak_angr` from PyPI, then it would go into `ofrak_angr` and then the `make develop` there would overwrite the downloaded one with the local one (and similar with `ofrak_capstone`). With #417, this means that not only the `ofrak_angr` itself is downloaded from PyPI, but also the `angr` itself and all its dependencies installed in `base.Dockerfile` are updated to the versions required by the `ofrak_angr`  from PyPI. Then `make develop` would put them all back. 

**Please provide some information about your environment.**
N/A

**If you've discovered it, what is the root cause of the problem?**
See above.

**How often does the issue happen?**
Always.

**What are the steps to reproduce the issue?**
See above.

**How would you implement this fix?**
Ideally, we ought to get rid of the `"test": "ofrak_angr~=1.0",  "ofrak_capstone~=1.0"]` dependency. But no idea what that would take.

**Are there any (reasonable) alternative approaches?**

Options:

- Leave as is, and give up on being able to install all dependencies in `base.Dockerfile` and not rely on PyPI in `make develop` afterwards. Seems pretty suboptimal in the long run…
- Implement a way to have the top-level `make develop` first `pip install -e .` without `[test]`, in `ofrak_core`, then `make develop` in `ofrak_angr`, then go back and do a full `make develop` in `ofrak_core`. I see three ways, none ideal:
   * Change the `develop` target in `ofrak_core/Makefile` to do this when `../ofrak_angr` exists. The downside is that this has a very unexpected side-effect for people calling this make develop for other reasons
   * Have the code in top-level `build_image.py` for making the `"$INSTALL_TARGET:` rule in top-level `Makefile` now do this special thing for `ofrak_core` (that is,  when calling `make develop` for each component in order, call `make develop_no_test`  in `ofrak_core` instead, then go back and call `make -C ofrak_core develop` at the end). OK, but special-case'ing `ofrak_core` is not very clean.
[...]

PR #417, opened 2024-02-09
Update to angr==9.2.77

ANogin:
- [X] I have reviewed the [OFRAK contributor guide](https://ofrak.com/docs/contributor-guide/getting-started.html) and attest that this pull request is in accordance with it.

**One sentence summary of this PR (This should go in the CHANGELOG!)**
Update to recent angr==9.2.77, which also necessitates Python >= 3.8 and capstone==5.0.0.post1

**Link to Related Issue(s)**
N/A

**Please describe the changes in your request.**

**Anyone you think should look at this, specifically?**
@EdwardLarson 

Note:
- This requires python 3.8, so includes and depends on #416 
- Two tests in `ofrak_angr` fail with this one

PR #420, opened 2024-02-12
Fix circular dependencies by making `ofrak_angr` and `ofrak_capstone` optional for `ofrak_core` tests

ANogin:

Optmizes the build of the --target=development docker containers for the `ofrak_core` "forward" test dependencies on `ofrak_angr` and `ofrak_capstone`

**Link to Related Issue(s)**
#419

Split the `make develop` for `ofrak_core` into two stages, to have the install process agree with the dependency order.

@whyitfor


ANogin:
While this works on master, when merged with #417, then during `make develop`, there is still a "fight" about the version of angr to install, so I am no longer sure this is an OK way to go for #419, marked as draft for now...

ANogin:
@whyitfor @rbs-jacob does this approach make sense?

ANogin:
Specifically, this results in something like
```make
develop:
        $(MAKE) -C ofrak_type STAGE=1 develop
        $(MAKE) -C ofrak_io STAGE=1 develop
        $(MAKE) -C ofrak_patch_maker STAGE=1 develop
        $(MAKE) -C ofrak_core STAGE=1 develop
        $(MAKE) -C ofrak_angr STAGE=1 develop
        $(MAKE) -C ofrak_capstone STAGE=1 develop
        $(MAKE) -C frontend STAGE=1 develop
        $(MAKE) -C ofrak_core STAGE=2 develop
```
where `$(MAKE) -C ofrak_core STAGE=1 develop` would `$(PIP) install -e .` (without circular dependencies), and then `$(MAKE) -C ofrak_core STAGE=2 develop` would `$(PIP) install -e .[docs,test]` (with circular dependencies, but they are already installed, so no issue).

whyitfor (commented):
This approach could work, but it introduces some complexity that I'm not sure is worth it.

Could the same end result be achieved by modifying the ofrak `develop` target to ensure proper install order?

Something like:
```
develop:
    $(PIP) install -e .
    $(PIP) install -e .[docs,test]
```

If a 2-3 line change addresses this issue I'd be in favor of that.

Alternatively, should the `develop` target by default install all ofrak-related dependencies from the source repository (and not from pypi)? I believe @rbs-jacob has suggested such a solution as being advantageous for development as well

ANogin:
> Could the same end result be achieved by modifying the ofrak `develop` target to ensure proper install order?
> 
> Something like:
> 
> ```
> develop:
>     $(PIP) install -e .
>     $(PIP) install -e .[docs,test]
> ```

No, if would have to be something like

```
develop:
     $(PIP) install -e .
     $(MAKE) -C disassemblers/ofrak_capstone develop
     $(MAKE) -C disassemblers/angr develop
     $(PIP) install -e .[docs,test]
```

Would that be OK?

> Alternatively, should the `develop` target by default install all ofrak-related dependencies from the source repository (and not from pypi)? I believe @rbs-jacob has suggested such a solution as being advantageous for development as well

What would be the approach for that? One that would probably work is:
- In the top-level Makefile, have `make develop` simply do `pip install -e path1 path2 path3...." for all OFRAK paths before the existing `$(MAKE) -C path1 develop; ${MAKE) -C path2 develop; ...`, except that it would require that all OFRAK modules are PIP modules (I think frontend is currently the main exception).

ANogin:
@whyitfor , I implemented an alternative approach where only the `make develop` target in `ofrak_core/Makefile` changes - does this work?

P.S. The reson for the other two changes is that:
- root `make image` rule is changed because `ofrak-core-dev.yml` no longer makes sense, as a minimal dev install needs to include capstone and angr (that is, `ofrak-angr.yml`)
- `importlib-resources` moved from setup to requirements so that it is installed in the `base.Dockerfile` stage (there because my test for this fix working right is that the full `make develop` suceeeds without needing network access)

Makefile
```diff
@@ -11,7 +11,7 @@ inspect: autoflake check-black
 
 .PHONY: image
 image:
-	python3 build_image.py --config ofrak-core-dev.yml --base --finish
+	python3 build_image.py --config ofrak-angr.yml --base --finish
```
whyitfor, line 14:
```suggestion
	python3 build_image.py --config ofrak-core-dev.yml --base --finish
```
Try reverting this change

whyitfor:
@ANogin, can you pull master into this for the CI tests?

ANogin:
@whyitfor , note my earlier caveat:

> root `make image` rule is changed because `ofrak-core-dev.yml` no longer makes sense, as a minimal dev install needs to include capstone and angr (that is, `ofrak-angr.yml`)

So reverting this change is not the right thing to do.

ANogin:
@whyitfor in order for things to still [kind of] work with `ofrak-core-dev.yml`, I tried an alternative approach, which I think is cleaner - the `ofrak_angr` and `ofrak_capstone` are not included as dependencies, and the tests that do rely on them are just skipped if they are not installed. The result is that `make -C /ofrak_core test` succeeds in e.g. `ofrak-angr.yml` but in `ofrak-core-dev.yml` while all tests succeed, the coverage check then fails with `FAIL Required function coverage of 100.00% not reached. Total coverage: 99.91%` (because of the skipped tests).

whyitfor (approved):
This seems like a good incremental compromise.

PR #424, opened 2024-02-18
Fix occasional spurious test failures.

ANogin:

Fix two tests that fail on occasion for [IMHO] no good reason.

**Link to Related Issue(s)**
N/A

Two tests would sometime fail for me:
- `ofrak_core/test_ofrak/components/test_comments.py::test_comment_content` had a default 200ms time limit, and once in a while, the limit would be violated (particularly, if other things happen to be running at the same time the tests run) - I have seem numbers as high as 1.6 s on occasion. I bumped the limit to 5s.
- `ofrak_core/test_ofrak/unit/test_ofrak_context.py::test_get_ofrak_context_over_time` would occasionally fail with the `RuntimeError: There is no current event loop in thread 'MainThread'` error message. I am guessing that would happen when another test executed earlier in the same worker would happen to mess with the asyncio event loop, so I made sure an event loop was available (similar to how the other tests in the same file do it). Not 100% sure this is the right thing, but seems to work.

Not sure. @whyitfor maybe?

### DIFF
--- a/ofrak_core/test_ofrak/components/test_comments.py
+++ b/ofrak_core/test_ofrak/components/test_comments.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 import pytest
 from hypothesis import given, HealthCheck, settings
 from hypothesis.strategies import text
@@ -38,7 +39,10 @@ async def test_adding_comments(executable_resource: Resource):
 # We suppress the function_scoped_fixture health check because the executable_resource fixture
 # doesn't need to be reset between individual runs of hypothesis (since the comment overrides
 # the previous one every time).
-@settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+@settings(
+    suppress_health_check=[HealthCheck.function_scoped_fixture],
+    deadline=timedelta(seconds=5),
+)
 @given(comment_str=text())
 async def test_comment_content(executable_resource: Resource, comment_str: str):
     """Test comments with all kinds of string contents."""
--- a/ofrak_core/test_ofrak/unit/test_ofrak_context.py
+++ b/ofrak_core/test_ofrak/unit/test_ofrak_context.py
@@ -64,6 +64,9 @@ def test_ofrak_context_exclude_components_missing_dependencies():
 
 
 def test_get_ofrak_context_over_time():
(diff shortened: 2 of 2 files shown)


ofrak_core/test_ofrak/components/test_comments.py
```diff
@@ -38,7 +39,10 @@ async def test_adding_comments(executable_resource: Resource):
 # We suppress the function_scoped_fixture health check because the executable_resource fixture
 # doesn't need to be reset between individual runs of hypothesis (since the comment overrides
 # the previous one every time).
-@settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+@settings(
+    suppress_health_check=[HealthCheck.function_scoped_fixture],
+    deadline=timedelta(seconds=5),
```
whyitfor, line 44: I'm not sure this is necessary. What this [deadline](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.deadline) is, as I understand, is the time it takes to generate the hypothesis test case. With default settings, this test gets run 200 times per test run, so the increased time delta could lead to long test runs.

I believe if you run tests on your machine hypothesis will generate new tests cases, so the longer you run it the more likely it might generate large `text` strategies. This, anyway, is my hypothesis on what you are observing. What if you try setting `text(max_size=100)` on line 46? That should increase the reliability of this test within the default 200ms time.

ANogin: @whyitfor , no per the link you cited, 

>  a duration [...] that each individual example (i.e. **each time your test function is called**, not the whole decorated test) within a test is not allowed to exceed. Tests which take longer than that may be **converted into errors** 

(emphasis mine). This is effectively a timeout - if for some reason, the test lasts [significantly] longer, it is reported as an error.

whyitfor: I see, you're right. I validated this with the following test:

```python
from hypothesis import strategies as st
from hypothesis import given
import time

@given(text=st.text())
def test_deadline(text: str):
    print(text)
    time.sleep(0.25)
    assert text == text
```

whyitfor (approved):
LGTM

PR #425, opened 2024-02-18
Move away from $OFRAK_DIR argument to $PACKAGE_PATH for Docker builds

whyitfor:

- $PACKAGE_PATH generalizes build_image.py such that it works with any yaml file containing a valid list of Python packages

Before this change, it was possible to use build_image.py a directory above the OFRAK repository by passing in an OFRAK_DIR arugment in the yml file. This change generalizes the approach so that build_image.py can be run in any ancestor directory; it also removes the need to pass in an OFRAK_DIR argument.

**Link to Related Issue(s)**
N/A.

See above.

@rbs-jacob

### DIFF
--- a/frontend/Dockerstub
+++ b/frontend/Dockerstub
@@ -2,9 +2,8 @@ RUN apt-get update && apt-get install --yes nginx
 
 COPY --from=svelte --chown=root:root /home/node/frontend/dist /ofrak_gui
 
-ARG OFRAK_DIR=.
-COPY $OFRAK_DIR/frontend/nginx.conf /etc/nginx/sites-enabled/default
+COPY $PACKAGE_PATH/nginx.conf /etc/nginx/sites-enabled/default
 
-COPY $OFRAK_DIR/mkdocs.yml /mkdocs.yml
-COPY $OFRAK_DIR/docs /docs
-COPY $OFRAK_DIR/examples /examples
+COPY $PACKAGE_PATH/../mkdocs.yml /mkdocs.yml
+COPY $PACKAGE_PATH/../docs /docs
+COPY $PACKAGE_PATH/../examples /examples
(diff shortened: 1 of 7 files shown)


frontend/Dockerstub
```diff
@@ -2,9 +2,8 @@ RUN apt-get update && apt-get install --yes nginx
...
-ARG OFRAK_DIR=.
-COPY $OFRAK_DIR/frontend/nginx.conf /etc/nginx/sites-enabled/default
+COPY $PACKAGE_PATH/nginx.conf /etc/nginx/sites-enabled/default
 
-COPY $OFRAK_DIR/mkdocs.yml /mkdocs.yml
-COPY $OFRAK_DIR/docs /docs
-COPY $OFRAK_DIR/examples /examples
+COPY $PACKAGE_PATH/../mkdocs.yml /mkdocs.yml
+COPY $PACKAGE_PATH/../docs /docs
+COPY $PACKAGE_PATH/../examples /examples
```
ANogin, line 9: This seems like a bit of a hack to me, and the way we are doing it is causing these `COPY` directives to be done twice when both `frontend` and `tutorial` packages are included. Any reason not create an actual "`ofrak_docs`" (or some such) package that would include docs, examples, and mkdocs.yml and its own `Dockerstub` with these `COPY` instructions? Just a thought...

whyitfor: I had a similar thought when refactoring this. I've made an issue for this, https://github.com/redballoonsecurity/ofrak/issues/426#issue-2142324456, as it is related to other documentation changes that make sense. Either way, I don't think this should block this PR.

rbs-jacob (approved):
Haven't run it myself to test, but after a quick pass, this seems like a positive improvement that makes sense.

PR #427, opened 2024-02-21
Refactor/hexview

dannyp303:

- Refactor HexView and related components to use mousewheel instead of scroll and compartmentalize all comonents to src/hex. 
**Link to Related Issue(s)**


frontend/src/hex/ByteclassView.svelte
```diff
@@ -124,17 +125,13 @@
...
-    ) {
+    if (data !== undefined && data.length > alignment * 3) {
       // Offset Y by 0.5 because of: https://stackoverflow.com/a/48970774
       context.strokeRect(
         0,
-        Math.ceil($scrollY.top * canvas.height) - 0.5,
+        Math.ceil((currentPosition / dataLength) * canvas.height) - 0.5,
         alignment,
-        Math.ceil(($scrollY.viewHeight * canvas.height) / 2)
+        Math.ceil(($screenHeight / dataLength) * canvas.height) - 0.5
```
rbs-jacob, line 134: Why does this no longer need to be divided by two?

frontend/src/hex/HexView.svelte
```diff
@@ -58,41 +80,53 @@
...
     startWindow = 0,
     endWindow = 0;
+  let hexDisplay = undefined;
 
   const alignment = 16,
     chunkSize = 4096,
     windowSize = chunkSize * 10,
     windowPadding = 1024;
 
+  $: dataLenPromise = dataLenPromise;
```
rbs-jacob, line 90:
```suggestion
```

Does this need to be here?

frontend/src/hex/HexView.svelte
```diff
@@ -251,7 +253,7 @@
       };
     }
 
-    if (dataSearchResults.matches) {
+    if (dataSearchResults.matches != undefined) {
```
rbs-jacob, line 256:
```suggestion
    if (dataSearchResults.matches != null) {
```

frontend/src/hex/HexView.svelte
```diff
@@ -265,123 +267,137 @@
...
+    hexDisplay = document.getElementById("scrollable");
+    $screenHeight =
+      Math.floor(hexDisplay.offsetHeight / lineHeight) * alignment;
+    chunkDataPromise = dataLenPromise.then(getNewData);
   }
+
+  onMount(() => {
+    hexDisplay = document.getElementById("scrollable");
+    refreshHeight();
+  });
```
rbs-jacob, line 285: The more "correct" way to do this in Svelte is to `bind:this` on the `scrollable` element. The variable should be reactively updated on mount.

frontend/src/hex/MinimapView.svelte
```diff
@@ -0,0 +1,47 @@
+<style>
+  .minimap {
+    width: 64px;
+    max-width: 64px;
+    height: 75vh;
```
rbs-jacob, line 5:
```suggestion
    height: 75vh;
```

Why should it only be 75% of the view height?

frontend/src/utils/Pane.svelte
```diff
@@ -10,117 +10,31 @@
     flex-flow: row;
     justify-content: space-between;
     align-items: stretch;
+    overflow: hidden;
   }
 
   .inner {
-    overflow: auto;
+    overflow: hidden;
```
rbs-jacob, line 17: This may have fixed some short-term problems, but I am really skeptical that this is not the cause of a bunch of other scroll things breaking.

I don't think this should have overflow hidden.

rbs-jacob (changes requested):
Some outstanding issue that should be fixed before this can be merged.

- [x] Likely due to `Pane` refactor, this PR removes padding all over the GUI that should be fixed
- [x] The search bar in the tree view is no longer pinned to the top, and scrolls away when you scroll the tree
- [x] The minimap now floats half way between the right of the hex view and the right of the screen when it should be pinned to the right of the screen
- [x] The minimap no longer takes up the full height of the screen, which makes it less practical to use, less easy to see details, etc.
- [x] The minimap and "go to offset" input no longer have a black background, and the hex can be seen around/behind them if the hex area is thin enough to cause overlap
- [ ] The "breadcrumb" that used to be at the top of the hex view is no longer visible (it does seem to come back in the disassembly view for some reason)
- [ ] The hex view does not allow scrolling past the bottom when using the scroll wheel – I know there is a comment in the code that this is a bug, but I actually ended up kind of preferring it this way
- [ ] Scrolling with the mini map does not round to the nearest line correctly (see video at the bottom)
- [ ] The bottom of the red "view window" in the minimap is not opaque
- [ ] Scrolling in the disassembly view no longer works
- [ ] Scrolling in the text view no longer works

Before: 
image

After:
image

Scrolling not working right:

https://github.com/redballoonsecurity/ofrak/assets/99368685/94f57fb6-a866-40fe-8d5d-e7b89db6fc5c

Left button bar lack of background:

image

Minimap needs background, and should be full height:

image

frontend/src/utils/SearchBar.svelte
```diff
@@ -113,7 +117,8 @@
       searchQuery == prevQuery &&
       searchOptions.searchType == prevOptions.searchType &&
       searchOptions.regex == prevOptions.regex &&
-      searchOptions.caseIgnore == prevOptions.caseIgnore
+      searchOptions.caseIgnore == prevOptions.caseIgnore &&
+      searchResults.matches != undefined //Search results are the only attribute of the search we have control of from other components, so we use it to clear the search when changeing selected resrouces.
```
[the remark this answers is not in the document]
dannyp303: Search bar takes a callback that tells it how to perform the search, and an object to populate with the results. I needed to clear the search bar from another component, so I did so by clearing the search results.

frontend/src/hex/ByteclassView.svelte
dannyp303: Why did it need to in the first place? 
The screen height should be the number of bytes being displayed as it is calculated with:
`  $screenHeight =
      Math.floor(hexDisplay.offsetHeight / lineHeight) * alignment;`
being the number of lines currently being displayed * 16 bytes per line. i.e bytes per line. 

The height of the rectangle is then the percentage of the current data being displayed * the size of the canvas.

rbs-jacob (changes requested):
Edge case that may not be worth worrying about, but the minimap gets cut off if the hex view gets too small.

https://github.com/redballoonsecurity/ofrak/assets/99368685/97052f59-8010-4887-b698-f36fbb3b3aa4

frontend/src/views/ComponentsView.svelte
```diff
@@ -95,7 +105,7 @@
   hr {
     border: 1px dashed var(--main-fg-color);
     width: 100%;
-    margin: 2em 0;
+    margin: 0.5em 0;
```
rbs-jacob, line 108:
```suggestion
    margin: 1em 0;
```

Let's split the difference and bump it up a little.

frontend/src/utils/Pane.svelte
```diff
@@ -10,117 +10,31 @@
...
 
 <script>
-  import { onMount } from "svelte";
-
-  export let paddingVertical = "3em",
-    paddingHorizontal = "3em",
-    scrollY = undefined,
-    displayMinimap = false;
-
+  export let paddingVertical = "0",
```
rbs-jacob, line 23: Reflecting on our discussion, I think we should have the default be at least `1em`, maybe `0.5em` at the very very least. Then, we can override this value to be `0` for the tree view since that's where it will look the best to have the pane move vertically.

My thinking is that this will make the default for non-tree views in the top left area of the page look much better. For example, there should (in my opinion) be more padding below the buttons at the bottom here:

image

frontend/src/ofrak/remote_resource.js
```diff
@@ -170,9 +170,10 @@ export class RemoteResource extends Resource {
     if (this.data_id === null) {
       return null;
     }
-    let result = await fetch(`${this.uri}/get_data_length`).then((r) =>
-      r.json()
-    );
+    let result = await fetch(`${this.uri}/get_data_length`).then(async (r) => {
+      let _result = await r.json();
+      return _result;
```
rbs-jacob, line 175: Is this code necessary?

frontend/src/hex/JumpToOffset.svelte
```diff
@@ -56,12 +40,12 @@
     if (e.key === 'Enter') {
       input.blur();
       try {
-        let result = calculator.calculate(input.value) + 1;
-        $scrollY.top = result / dataLength;
+        currentPosition = calculator.calculate(input.value);
```
rbs-jacob, line 43: This needs to be rounded. We can now go to an unaligned position like `7` and then the rest of the results will be misaligned when scrolling the hex view.

Also the value no longer clears properly after it has been set. So once it has been set the value no longer updates on scroll.

Last thing is that we should be using the actual value, not the placeholder value. Not only does having full-opacity text look better, but it means that a user can edit the current location instead of having to manually retype it from scratch every time. It would also solve the above problem.


https://github.com/redballoonsecurity/ofrak/assets/99368685/961d85d4-37cc-4d9b-b826-944bdbfcda24

frontend/src/hex/HexView.svelte
```diff
@@ -132,19 +173,15 @@
...
     ).map((chunk) => chunkList(buf2hex(chunk), 2));
   }
-  $: if (scrollY !== undefined && $scrollY !== undefined) {
-    chunkDataPromise = dataLenPromise.then(getNewData);
-  }
 
   async function calculateRanges(resource, dataLenPromise, colors) {
     const children = await resource.get_children();
-    if (children === []) {
+    if (children == []) {
```
rbs-jacob, line 184:
```suggestion
    if (children.length === 0) {
```

frontend/src/hex/HexView.svelte
```diff
@@ -265,123 +273,140 @@
...
-    }
+  function resetResource() {
+    dataSearchResults.matches = undefined;
+    currentPosition = 0;
   }
 
-  async function searchHex(query, options) {
-    return await $selectedResource.search_data(query, options);
+  function refreshHeight() {
+    hexDisplay = document.getElementById("scrollable");
```
rbs-jacob, line 282: I think we talked about having this be set by `bind:this`?

rbs-jacob (changes requested):

frontend/src/hex/HexView.svelte
dannyp303: Woops, i added this, but missed this re-assignment.

frontend/src/hex/HexView.svelte
```diff
@@ -67,32 +98,41 @@
...
+  // React to local data searches
+  $: if (dataSearchResults) {
+    const localDataSearchResults = dataSearchResults;
+    if (
+      localDataSearchResults.matches?.length > 0 &&
+      (localDataSearchResults.index || localDataSearchResults.index === 0)
+    ) {
+      dataLenPromise.then((dataLength) => {
+        currentPosition =
+          localDataSearchResults.matches[localDataSearchResults.index][0];
```
rbs-jacob, line 125: This should also be rounded, otherwise search can jump to an unaligned offset.

rbs-jacob (changes requested):
One additional observation about the pane changes is that the vertical and horizontal scrolls are handled differently in a way that I think is potentially confusing for a user. Specifically: there are horizontal scroll bars, but they're way at the bottom and can go out of view. Also if you middle click in the tree with the mouse wheel, it only lets you scroll horizontally. 

Similarly, horizontal scrolling the hex view doesn't work if it's too small. 

I've included a video, but I don't consider fixing the horizontal scroll critical for this PR. If it's an easy fix, we should do it. But it shouldn't block the PR getting merged.

https://github.com/redballoonsecurity/ofrak/assets/99368685/0e673b5f-d744-43a6-9541-701dad530cff

dannyp303:
> One additional observation about the pane changes is that the vertical and horizontal scrolls are handled differently in a way that I think is potentially confusing for a user. Specifically: there are horizontal scroll bars, but they're way at the bottom and can go out of view. Also if you middle click in the tree with the mouse wheel, it only lets you scroll horizontally.
> 
> Similarly, horizontal scrolling the hex view doesn't work if it's too small.
> 
> I've included a video, but I don't consider fixing the horizontal scroll critical for this PR. If it's an easy fix, we should do it. But it shouldn't block the PR getting merged.
> 
>  Screen.Recording.2024-02-27.at.5.37.07.PM.mov

Horizontal scrolling here would be a bit weird, I think the best solution would be to adjust the alignment size based on the hexview element width. In the interest of time, i think this should be a separate PR.

rbs-jacob:
> Horizontal scrolling here would be a bit weird, I think the best solution would be to adjust the alignment size based on the hexview element width. In the interest of time, i think this should be a separate PR.

I'm more concerned about horizontal scrolling in the tree view being weird than about not being able to horizontally scroll the hex view. The only reason I bring them up together is that I think they're likely related. 

Like I said, no need to block this PR on that.

frontend/src/hex/JumpToOffset.svelte
rbs-jacob: The last part of this one is still unresolved – the go to offset text input should set the `value` and not the `placeholder` so that it looks better and stays updated.

dannyp303:
> > Horizontal scrolling here would be a bit weird, I think the best solution would be to adjust the alignment size based on the hexview element width. In the interest of time, i think this should be a separate PR.
> 
> I'm more concerned about horizontal scrolling in the tree view being weird than about not being able to horizontally scroll the hex view. The only reason I bring them up together is that I think they're likely related.
> 
> Like I said, no need to block this PR on that.


Not related, the hex view doesnt "scroll" so i have scrolling totally turned off for that view, if I enable it, it does the "double scroll" that initiated this entire PR. Regardin the resource view, easy fix, theres an `overflow-x: auto` in the treebox div that isnt required any more.

rbs-jacob (approved):
Thanks for all of the revisions. Looks good to me! 🤑

PR #433, opened 2024-02-26
Catch PermissionError when checking for dependencies

hikir1:

Catch PermissionError when checking for dependencies, and issue a warning.

Fixes [this issue](https://github.com/redballoonsecurity/ofrak/issues/430)

### DIFF
--- a/ofrak_core/test_ofrak/unit/test_ofrak_context.py
+++ b/ofrak_core/test_ofrak/unit/test_ofrak_context.py
@@ -64,6 +64,9 @@ def test_ofrak_context_exclude_components_missing_dependencies():
 
 
 def test_get_ofrak_context_over_time():
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+
     # No active context before running OFRAK
     with pytest.raises(InvalidStateError):
         get_current_ofrak_context()
--- a/build_image.py
+++ b/build_image.py
@@ -192,6 +192,8 @@ def create_dockerfile_base(config: OfrakImageConfig) -> str:
             continue
         with open(dockerstage_path) as file_handle:
             dockerstub = file_handle.read()
+        # Cannot use ENV here because of multi-stage build FROM, so replace direclty in Docerkstage contents
+        dockerstub = dockerstub.replace("$PACKAGE_DIR", package_path)
         dockerfile_base_parts += [f"### {dockerstage_path}", dockerstub]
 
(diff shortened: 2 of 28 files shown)


rbs-jacob (approved):
This looks ready to merge to me

rbs-jacob (changes requested):
On second thought, needs tests.

ofrak_core/test_ofrak/unit/test_ofrak_context.py
```diff
@@ -14,6 +14,32 @@
...
+    for tool in (
+        ComponentExternalTool("apktool", "", "-version"),
+        _UnsquashfsV45Tool(),
+        _StringsToolDependency(),
+    ):
+        task = loop.create_task(tool.is_tool_installed())
+        loop.run_until_complete(task)
+        expected = f"Encountered PermissionError while searching PATH for {tool.tool}."
+        found = [r for r in caplog.records if r.message == expected]
+        assert len(found) == 1, "Missing error message from component_model"
```
rbs-jacob, line 41:
```suggestion
        assert any(r.message == expected for r in caplog.records), "Missing error message from component_model"
```

This way of doing it is a little more concise. Will probably need to re-run the linter after applying the suggestion, though.

ofrak_core/test_ofrak/unit/test_ofrak_context.py
```diff
@@ -14,6 +14,32 @@
...
+
+    monkeypatch.setenv("PATH", "/root")
+
+    for tool in (
+        ComponentExternalTool("apktool", "", "-version"),
+        _UnsquashfsV45Tool(),
+        _StringsToolDependency(),
+    ):
+        task = loop.create_task(tool.is_tool_installed())
+        loop.run_until_complete(task)
```
rbs-jacob, line 38: Why do it this way instead of just making the test `async` and using [`pytest.mark.asyncio`](https://pytest-asyncio.readthedocs.io/en/latest/)?

rbs-jacob (changes requested):

ofrak_core/test_ofrak/unit/test_ofrak_context.py
rbs-jacob: On a second read, my suggested change is not exactly the same as what the original test does, so feel free to resolve this. 

But this particular test is failing in the CI check, so this will have to be revised in some way before we can merge.

Issue #458, opened 2024-04-20
Add support for multiple comments on the same Range


Ivanov1ch:
**What is the use case for the feature?**
Making multiple comments without specifying a range - eg custom unpackers automatically adding comments with file metadata to be displayed in the GUI.
**Does the feature contain any proprietary information about another company's intellectual property?**
No
**How would you implement this feature?**
Refactor CommentsAttributes.comments to from the [current](https://github.com/redballoonsecurity/ofrak/blob/bdb378b0a153aef6de8e13b6b06a0c255fc80e68/ofrak_core/ofrak/core/comments.py#L18C5-L18C41) `comments: Dict[Optional[Range], str]` to `comments: Dict[Optional[Range], List[str]]`. Keeping the `Dict` would mean less changes to the existing codebase and quicker lookup times for specific comments at scale when compared to another approach like a List of comments. 
**Are there any (reasonable) alternative approaches?**
The [current approach handles the creation of a new comment on the same range as an existing comment by appending a newline](https://github.com/redballoonsecurity/ofrak/blob/bdb378b0a153aef6de8e13b6b06a0c255fc80e68/ofrak_core/ofrak/core/comments.py#L55). This combines the comments into one larger comment, which both messes up the `# Comment` formatting on the GUI and means that it is impossible to delete one comment without deleting both, making it a bad alternative.
**Are you interested in implementing it yourself?**
Yes. PR incoming

PR #459, opened 2024-04-20
Add support for multiple comments on the same Range

Ivanov1ch:

Added support for multiple comments on the same Range
**Link to Related Issue(s)**
#458 
Refactored `CommentsAttributes.comments` to from the [current](https://github.com/redballoonsecurity/ofrak/blob/bdb378b0a153aef6de8e13b6b06a0c255fc80e68/ofrak_core/ofrak/core/comments.py#L18C5-L18C41) comments: Dict[Optional[Range], str] to comments: Dict[Optional[Range], List[str]]. Adjusted `comments.py`, the web API, and the frontend to account for this. Also adjusted `test_comments.py` for more coverage on the new format. The interface for adding a comment in `comments.py` (`AddCommentModifierConfig`) has not been changed, but the interface for deleting a comment (`DeleteCommentModifierConfig`) has been. `DeleteCommentModifierConfig` has been modified in a way that should remain backwards compatibility - creating a `DeleteCommentModifierConfig` with the `comment_range` of the old type, `Optional[Range]` will still work with the new type: `Tuple[Optional[Range], Optional[str]]`, as the post-init hook will automatically convert it. `test_comments.py` has also been adjusted and expanded.
@rbs-jacob


Ivanov1ch:
A question that remains is about backwards compatibility and, of course, the hardest question of all - naming. In the current state, `DeleteCommentModifierConfig` is defined as follows:

```python
@dataclass
class DeleteCommentModifierConfig(ComponentConfig):
    """
    comment_range: Tuple[Optional[Range], comment_text=Optional[str]]
    If comment_text is provided, deletes the matching comment with the same Optional[Range]
    If comment_text is None, deletes ALL comments with the same Optional[Range]
    """

    comment_range: Tuple[Optional[Range], Optional[str]]

    def __post_init__(self):
        # Ensure there's always a two-element Tuple
        if type(self.comment_range) == tuple:
            # New format
            self.comment_range = (*self.comment_range, None)[:2]
        elif type(self.comment_range) == Range:
            # Old format: Range
            self.comment_range = (self.comment_range, None)
        else:
            # Old format: None (no Range provided)
            self.comment_range = (None, None)
```

So the field `comment_range` retained the same name as the current definition of the class:
```python
@dataclass
class DeleteCommentModifierConfig(ComponentConfig):
    comment_range: Optional[Range]
```

Which ensures that any old code that would run something like this still works:

```python
await executable_resource.run(
        DeleteCommentModifier,
        DeleteCommentModifierConfig(comment_range=Range(0, 2)),
    )
```

However, the name `comment_range` makes less sense now as this is more than just an `Optional[Range]`. Another option could be to leave `comment_range: Optional[Range]` and add a second field named something like `comments_to_delete: Tuple[Optional[Range], Optional[str]]` and use the post-init hook to populate it from `comment_range` if it is still used.

Any input would be appreciated.

ofrak_core/ofrak/core/comments.py
```diff
@@ -50,18 +49,34 @@ async def modify(self, resource: Resource, config: AddCommentModifierConfig) ->
 
         # Here I'm appending appending overlapping comments with a new line.
```
alchzh, line 50: These changes obsolete this comment

ofrak_core/ofrak/core/comments.py
```diff
@@ -50,18 +49,34 @@ async def modify(self, resource: Resource, config: AddCommentModifierConfig) ->
...
 @dataclass
 class DeleteCommentModifierConfig(ComponentConfig):
-    comment_range: Optional[Range]
+    """
+    comment_range: Tuple[Optional[Range], comment_text=Optional[str]]
+    If comment_text is provided, deletes the matching comment with the same Optional[Range]
+    If comment_text is None, deletes ALL comments with the same Optional[Range]
+    """
+
+    comment_range: Tuple[Optional[Range], Optional[str]]
```
alchzh, line 67: I would separate out the `comment_text` parameter as a separate parameter instead of doing funky stuff with tuples, it doesn't make intuitive sense for the "text" to be part of the "range". Makes the code a lot more readable later on too.

The typing of this param with the processing in `__post_init__` is quite complicated, it would accurately be represented as 

`Union[Optional[Range], Tuple[Optional[Range]], Tuple[Optional[Range], Optional[str]]]`

ofrak_core/ofrak/core/comments.py
```diff
@@ -82,10 +97,24 @@ async def modify(self, resource: Resource, config: DeleteCommentModifierConfig)
         except NotFoundError:
             comments = {}
         try:
-            del comments[config.comment_range]
+            if len(config.comment_range) == 1:
+                config.comment_range = (config.comment_range[0], None)
+
+            if config.comment_range[1] is None:
```
alchzh, line 103: See comment above about having a separate `config.comment_text` and `config.comment_range` to clean up this section

alchzh (changes requested):

Ivanov1ch:
@alchzh updated according to your feedback

whyitfor:
@Ivanov1ch, please pull master into this for the CI tests.

Ivanov1ch:
@whyitfor done

frontend/src/views/Comment.svelte
```diff
@@ -20,17 +20,18 @@
   import Hoverable from "../utils/Hoverable.svelte";
   import Icon from "../utils/Icon.svelte";
   export let comment, rootResource, selfId;
-  let text = comment[1];
-  let addresses = text.matchAll("#[a-fA-F0-9]+[@0x[0-9a-fA-F]+]*", text);
+  let addresses = comment.comment_text.matchAll(
+    "#[a-fA-F0-9]+[@0x[0-9a-fA-F]+]*"
```
rbs-jacob, line 24: I know you didn't write this regex originally, but I'm almost certain it doesn't do what it's supposed to do. May as well fix here.

`[xyz]` selects one of the characters `x`, `y`, or `z`, whereas `(xyz)` matches the group of characters `xyz`. This regular expression is almost certainly supposed to match the latter.

```suggestion
    "#[a-fA-F0-9]+(@0x[0-9a-fA-F]+)*"
```

I also wonder if that last `*` is actually supposed to be a `?`...

@dannyp303 any comment on this?

ofrak_core/test_ofrak/unit/test_ofrak_server.py
```diff
@@ -454,29 +454,110 @@ async def test_find_and_replace(ofrak_client: TestClient, hello_world_elf):
...
+    resp = await client.post(
+        f"/{create_body['id']}/add_comment",
+        json=[[0, len(hello_world_elf) + 1], "test comment out of bounds"],
+    )
+    assert resp.status == 500
+    resp = await client.post(
+        f"/{create_body['id']}/add_comment",
+        json=[[-1, len(hello_world_elf)], "test comment out of bounds"],
+    )
+    assert resp.status == 500
```
rbs-jacob, line 491: It's probably a bit brittle to assert that this returns a 500 (internal server) error. It would likely be better to assert that this does not return a 200.

rbs-jacob (changes requested):
Some minor comments, then this is probably good to go.

frontend/src/views/Comment.svelte
Ivanov1ch: `#[a-fA-F0-9]+(@0x[0-9a-fA-F]+)?` seems very reasonable

dannyp303: > I know you didn't write this regex originally, but I'm almost certain it doesn't do what it's supposed to do. May as well fix here.
> 
> `[xyz]` selects one of the characters `x`, `y`, or `z`, whereas `(xyz)` matches the group of characters `xyz`. This regular expression is almost certainly supposed to match the latter.
> 
> I also wonder if that last `*` is actually supposed to be a `?`...
> 
> @dannyp303 any comment on this?

I think the original is the proper one, it's named `address` but really what it is is `#<resource_id>@offset into file`, and the result of the match should be that entire regex pattern, not just the address into it. If you see the function `createAddressButton` it better shows what the result should be. Now if this is the best way to do this, probably not.

rbs-jacob (approved):

PR #79, opened 2022-10-21
Switch pigz for gzip

alexreuter:
**Link to Related Issue(s)**
[Related Issue](https://github.com/redballoonsecurity/ofrak/issues/54)

**Please describe the changes in your request.**
- Command line usage of gzip in ofrak_components is replaced with pigz
- Dockerstub updated to install pigz in the generated image
- Documentation for Mac updated

**Anyone you think should look at this, specifically?**
@whyitfor

PR #472, opened 2024-06-04
Use GzipFile python unpacker for speed, fall back on pigz if needed

ANogin:
- [X ] I have reviewed the [OFRAK contributor guide](https://ofrak.com/docs/contributor-guide/getting-started.html) and attest that this pull request is in accordance with it.

Use GzipFile python unpacker for speed, fall back on pigz if needed

**Link to Related Issue(s)**
N/A

Back before OFRAK became public we switched to using external gzip, and then later in #79 we switched to pigz. The reason the initial switch to gzip was made is because unlike python gzip, gzip and pigz binaries are willing to accept gzipped data with extra junk at the end as a "warning" (exit code 2, as opposed to exit code 1 for errors). Unfortunately, spawning external processes has nontrivial overhead, and asyncio is not enough to mitigate it. This commit tries a compromise - try `gzip.GzipFile` in python first, then fall back to pigz if BadGzipFile is raise.

@whyitfor

### DIFF
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 - View resource attribute string values containing only digits primarily as strings, alternatively as hex numbers. ([#423](https://github.com/redballoonsecurity/ofrak/pull/423))
 
 ### Changed
+- In `GripUnpacker`, use `gzip.GzipFile` python unpacker for speed, fall back on `pigz` if needed ([#472](https://github.com/redballoonsecurity/ofrak/pull/472))
 - Change `FreeSpaceModifier` & `PartialFreeSpaceModifier` behavior: an optional stub that isn't free space can be provided and fill-bytes for free space can be specified. ([#409](https://github.com/redballoonsecurity/ofrak/pull/409))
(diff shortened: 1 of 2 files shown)


whyitfor:
Would it make sense to add a benchmark test with this to record the speedup? Perhaps a recursive unpack of a filesystem pulled from http://tinycorelinux.net/15.x/x86/release/ or somewhere else?

We would probably want to find a pytest fixture that can record several runs of one test case, and/or perhaps fail if the execution time exceeds a certain threshold. Alternatively, we just add a fixture that records test duration

whyitfor (changes requested):
Please update the changelog

ANogin:
> Please update the changelog

Done.

rbs-jacob:
Since we're adding fallbacks like this, it probably also makes sense to fall back on `gzip` if `pigz` is not installed. We can be sure that it will be present in Docker images, but it likely will not for people using `pip install` OFRAK.

whyitfor:
@rbs-jacob, interesting idea re:`gzip`/`pigz`. Implementing that cleanly with our `ComponentExternalTool` setup might take some additional thought. I'm in favor of merging this in for now to get the performance improvement -- if you see value in that as a further change please add an issue.

whyitfor (approved):

Issue #311, opened 2023-05-19
Minor: `/tmp/ofrak.log` probably shouldn't be hardcoded


ANogin:
**What is the problem? (Here is where you provide a complete Traceback.)**

We probably should not assume we can have exclusive use of `/tmp/ofrak.log`. It may be somewhat safe in a controlled docker environment where there is likely at most one instance of OFRAK running at a time, but not with `pip install`, or whatever weird things people might choose to do, resulting in:

```
    ofrak = OFRAK()
  File "/ofrak_core/ofrak/ofrak_context.py", line 125, in __init__
    logging.getLogger().addHandler(logging.FileHandler(DEFAULT_OFRAK_LOG_FILE))
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1087, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1116, in _open
    return open(self.baseFilename, self.mode, encoding=self.encoding)
PermissionError: [Errno 13] Permission denied: '/tmp/ofrak.log'
```

**Please provide some information about your environment.**
At minimum we would like the following information on your platform and Python environment:
  - Platform: `redballoonsecurity/ofrak/ghidra`
  - Latest `master` branch.

**If you've discovered it, what is the root cause of the problem?**
`DEFAULT_OFRAK_LOG_FILE = os.path.join(tempfile.gettempdir(), "ofrak.log")` in `ofrak_core/ofrak/ofrak_context.py` seems to be hardcoded (as far as I can tell, the `DEFAULT_` part is misleading).

**How often does the issue happen?**
- Whenever '/tmp/ofrak.log' is not writable

**What are the steps to reproduce the issue?**
- Create '/tmp/ofrak.log' as root (e.g run ofrak as root), then run ofrak as non-root.
 
**How would you implement this fix?**
- Maybe at least include a process id in the log file name, or some such? Open to suggestions.

**Are there any (reasonable) alternative approaches?**
- Plenty

**Are you interested in implementing it yourself?**
- Sure - just let me know what behavior might make sense

whyitfor:
Interesting. I've never run into this issue. What steps did you take to create this situation where you did not have permission for `/tmp/ofrak.log`?

ANogin:
> Interesting. I've never run into this issue. What steps did you take to create this situation where you did not have permission for `/tmp/ofrak.log`?

[...]

PR #480, opened 2024-07-04
Make the log file name configurable and include timestamp by default

ANogin:

By default, the ofrak log is now `ofrak-YYYYMMDDhhmmss.log` rather than just `ofrak.log` and the name can be specified on the command line.

**Link to Related Issue(s)**
#311 

See summary

@whyitfor maybe?

### DIFF
--- a/ofrak_core/ofrak/cli/ofrak_cli.py
+++ b/ofrak_core/ofrak/cli/ofrak_cli.py
@@ -14,7 +14,7 @@ from importlib_metadata import entry_points
 
 from ofrak.component.interface import ComponentInterface
 from ofrak.model.component_model import ComponentExternalTool
-from ofrak.ofrak_context import OFRAKContext, OFRAK
+from ofrak.ofrak_context import DEFAULT_LOG_FILE, OFRAKContext, OFRAK
 from synthol.injector import DependencyInjector
 
 BACKEND_PACKAGES: Dict[Optional[str], Set[str]] = {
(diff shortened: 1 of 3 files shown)


ofrak_core/ofrak/cli/ofrak_cli.py
```diff
@@ -127,6 +127,11 @@ def add_ofrak_arguments(command_subparser):
             choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
             default=OFRAK.DEFAULT_LOG_LEVEL,
         )
+        command_subparser.add_argument(
+            "--log-file",
+            help="Log file to use; defaults to ofrak<YYYYMMDDhhmmss.log in a temp directory",
```
rbs-jacob, line 132:
```suggestion
            help="Log file to use; defaults to ofrak<YYYYMMDDhhmmss>.log in a temp directory",
```

rbs-jacob (changes requested):
Change seems fairly straightforward and makes sense to me.

One thing I want to point out (for @whyitfor in particular) is that this changes a top-level API, and may necessitate a major version bump in OFRAK. The argument added to the `OFRAK` initializer is optional, so it shouldn't really break anything, nor should tha block this from being merged. Just wanted to highlight that we should bump semver on the next tagged release.

ofrak_core/ofrak/cli/ofrak_cli.py
```diff
@@ -127,6 +127,11 @@ def add_ofrak_arguments(command_subparser):
             choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
             default=OFRAK.DEFAULT_LOG_LEVEL,
```
rbs-jacob, line 128: If the `DEFAULT_LOG_LEVEL` is a field of the `OFRAK` class, does it make sense for the `DEFAULT_LOG_FILE` path to be as well?

rbs-jacob (changes requested):

ofrak_core/ofrak/cli/ofrak_cli.py
ANogin: No opinion here - I kept it at top level, where it was, but could be moved into `OFRAK` class, I guess.

rbs-jacob: I think it's fine to leave as-is. I'll resolve this conversation and run the tests. Once they pass, this should be good to merge

rbs-jacob (approved):

Issue #476, opened 2024-06-14
Fall back on `gzip` if `pigz` fails


rbs-jacob:
Recent PR #472 uses Python's built-in `gzip` module for decompressing GZIP data if possible, and falls back on the system version of `pigz` if the Python version fails. 

Inside of OFRAK Docker images, we can be confident that `pigz` is present. But when a user installs OFRAK via `pip install`, they may not have `pigz` on their system. They are, however, likely to have `gzip` on their system. As such, we should fall back from Python `gzip` to `pigz` to `gzip` to increase the likelihood that *something* will work for GZIP decompression. 

Implementing this may involve modifying `ComponentExternalTool` ([see this comment on #472](https://github.com/redballoonsecurity/ofrak/pull/472#issuecomment-2151308732)). This may have implications for other OFRAK functionality using `ComponentExternalTool`.

alchzh:
If we're only using pigz as a fallback for the trailing data case and prefer python standard library gzip otherwise, I think it makes more sense to implement a fix in python ourselves, it's a pretty simple modification to the standard library package. (see this stackoverflow answer: https://stackoverflow.com/questions/4928560/how-can-i-work-with-gzip-files-which-contain-extra-data)

there isn't any sort of official Windows packaging for `pigz`

whyitfor:
Like the idea of removing dependencies if we can. If your solution passes all the tests (including https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/test_ofrak/components/test_gzip_component.py#L38, then it should work).

Would also make sense to update our testing code to use python std. library to set up/validate zip files for testing.

rbs-jacob:
One issue not taken into account in #472, when we switched the default GZIP unpacker to the Python standard library instead of `pigz`, is that `pigz` is almost certainly faster for large files where the constant cost of starting a process is negligible compared to the processing time. 

If, in fact, we can reliably use Python standard library `gzip` to correctly handle trailing garbage data as that StackOverflow post implies, then probably the logic should be something like:

- If file over size threshold and `pigz` is installed:
    - Inflate with `pigz`
- Else
    - Inflate with Python standard library `gzip`

alchzh:
Can be closed after #485 merged. Should have written "Fixes" in that PR 🤦

PR #485, opened 2024-07-17
Rewrite gzip component to support trailing bytes without external tool and compress with PIGZ. Addresses #476

alchzh:

Rewrite gzip component to support trailing bytes without external tool. 

**Link to Related Issue(s)**

#476
python/cpython#68489

- Replaces the use of the `gzip` library with calls to the underlying `zlib` library directly, in a way that doesn't fail with trailing garbage at the end of files.
- Substantially rewrites of gzip test cases to use a common template
  - adds a test for multi member gzip files
  - the trailing bytes test actually tests the unpack


ofrak_core/test_ofrak/components/test_gzip_component.py
```diff
@@ -1,87 +1,55 @@
...
 from pytest_ofrak.patterns.compressed_filesystem_unpack_modify_pack import (
     CompressedFileUnpackModifyPackPattern,
 )
-from pytest_ofrak.patterns.unpack_modify_pack import UnpackModifyPackPattern
 
 
-class TestGzipUnpackModifyPack(CompressedFileUnpackModifyPackPattern):
+class GzipUnpackModifyPackPattern(CompressedFileUnpackModifyPackPattern):
+    """Template for tests that test different inputs the gzip component should support
+    unpacking."""
```
rbs-jacob, line 16:
```suggestion
    """
    Template for tests that test different inputs the gzip component should support
    unpacking.
    """
```

Minor, but FYI for future that we format docstrings in a specific way. [More info in the OFRAK contributor style guide docs.](https://ofrak.com/docs/contributor-guide/getting-started.html#functions-and-methods)

ofrak_core/ofrak/core/gzip.py
```diff
@@ -41,44 +39,43 @@ class GzipUnpacker(Unpacker[None]):
...
+                # Forward any gzip warning message and continue
+                if proc.returncode == -2 or proc.returncode == 2:
+                    LOGGER.warning(stderr)
+                else:
+                    raise CalledProcessError(returncode=proc.returncode, cmd=cmd, stderr=stderr)
+
+            await resource.create_child(
+                tags=(GenericBinary,),
+                data=stdout,
+            )
```
rbs-jacob, line 78: Probably makes sense to break each case of this `if` into its own helper function. 

The reason is that we mandate 100% function coverage in testing before code can be merged, but we have no requirements about line coverage or branch coverage. We want to make sure that we test both of these cases, with no possibility of regressions in the future. As such, if each case is its own function, our function coverage testing will confirm that both branches are hit. 

I suspect that all of our test files are small, and that the `pigz` case is never tested.

rbs-jacob (changes requested):

ofrak_core/ofrak/core/gzip.py
alchzh: Refactored into separate functions and added test case with a big binary for coverage

rbs-jacob (changes requested):
This needs a changelog entry, but after that, should probably be good to go.

ofrak_core/test_ofrak/components/test_gzip_component.py
```diff
@@ -1,87 +1,103 @@
...
-    @pytest.fixture(autouse=True)
-    def create_test_file(self, tmpdir):
-        d = tmpdir.mkdir("gzip")
-        fh = d.join("hello.gz")
-        result = BytesIO()
-        with GzipFile(fileobj=result, mode="w") as gzip_file:
-            gzip_file.write(self.INITIAL_DATA)
-        fh.write_binary(result.getvalue())
+    def write_gzip(self, gzip_path: Path):
+        raise NotImplementedError
```
whyitfor, line 50: This should be `raise NotImplementedError()`: https://ofrak.com/docs/contributor-guide/getting-started.html#raising-exceptions.

ofrak_core/test_ofrak/components/test_gzip_component.py
```diff
@@ -1,87 +1,103 @@
...
+        with gzip.GzipFile(gzip_path, mode="w") as gzip_file:
+            middle = len(self.INITIAL_DATA) // 2
+            gzip_file.write(self.INITIAL_DATA[:middle])
+            gzip_file.write(self.INITIAL_DATA[middle:])
+
+
+class TestGzipWithTrailingBytesUnpackModifyPack(GzipUnpackModifyPackPattern):
+    def write_gzip(self, gzip_path: Path):
+        with gzip.GzipFile(gzip_path, mode="w") as gzip_file:
+            gzip_file.write(self.INITIAL_DATA)
```
whyitfor, line 89: Could you just concatenate DEADBEEF on this line?

ofrak_core/test_ofrak/components/test_gzip_component.py
whyitfor: Should it also be marked as an `@abstractmethod`?

ofrak_core/ofrak/core/gzip.py
```diff
@@ -41,44 +39,51 @@ class GzipUnpacker(Unpacker[None]):
...
+        pigz_installed = await PIGZ.is_tool_installed()
+        if len(data) >= 1024 * 1024 * 4 and pigz_installed:
+            uncompressed_data = await self.unpack_with_pigz(data)
+        else:
+            try:
+                uncompressed_data = await self.unpack_with_zlib_module(data)
+            except Exception:  # pragma: no cover
+                if not pigz_installed:
+                    raise
+                uncompressed_data = await self.unpack_with_pigz(data)
```
whyitfor, line 51: Since `unpack_with_zlib_module` now handles the trailing bytes, I'm not following why we are catching exceptions here?

It seems like this logic could be simplified to:

```suggestion
        pigz_installed = await PIGZ.is_tool_installed()
        if pigz_installed and len(data) >= 1024 * 1024 * 4:
            uncompressed_data = await self.unpack_with_pigz(data)
        else:
            uncompressed_data = await self.unpack_with_zlib_module(data)
```

ofrak_core/ofrak/core/gzip.py
```diff
@@ -87,19 +92,16 @@ class GzipPacker(Packer[None]):
...
     async def pack(self, resource: Resource, config=None):
         gzip_view = await resource.view_as(GzipData)
-
-        result = BytesIO()
-        with GzipFile(fileobj=result, mode="w") as gzip_file:
-            gzip_child_r = await gzip_view.get_file()
-            gzip_data = await gzip_child_r.get_data()
-            gzip_file.write(gzip_data)
-
+        gzip_child_r = await gzip_view.get_file()
```
whyitfor, line 98: Why is this change necessary? What was wrong with the Packer as it was implemented?

whyitfor (changes requested):
See comments.

ofrak_core/ofrak/model/component_model.py
```diff
@@ -18,7 +18,7 @@ class ComponentConfig:
     """
 
 
-@dataclass(frozen=True)
+@dataclass(unsafe_hash=True)
```
whyitfor, line 21: I'd rather not change this.

The `ComponentExternalTool` is a nice, simple, stateless class. Can you cache the tool installation state in the `GzipUnpacker`? Until we have more use cases for caching this result I'd rather keep this class simple and immutable.

ofrak_core/test_ofrak/components/test_gzip_component.py
alchzh: no, that would add it to the compressed data section instead of writing extra trailing bytes.

ofrak_core/ofrak/core/gzip.py
alchzh: I had that originally, but I figured in case of some other format incompatibility or big in python we're not aware of might as well have the fallback still.

ofrak_core/ofrak/core/gzip.py
alchzh: using `zlib` is much faster, avoids a bunch of unnecessary processing in GzipFile

test on Silesia corpus:
https://gist.github.com/alchzh/a46140e683ff5e05be8d6b027a4f3fba

almost 4x faster on "mozilla" which is a 50MB tarfile of a Linux application so probably most applicable for us (4.4s for GzipFile vs 1.2s for zlib)

alchzh: on that note, we should probably be using pigz for packing large files as well

ofrak_core/ofrak/core/gzip.py
alchzh: no longer using PIGZ for decompression

alchzh:
Based on benchmarks of zlib module vs pigz, no benefit to using pigz for decompression, but significant benefit to using it for compression.

 https://gist.github.com/alchzh/24ca62a6f87548ea2b1911c174e0ed54

ofrak_core/ofrak/core/gzip.py
```diff
@@ -87,19 +82,48 @@ class GzipPacker(Packer[None]):
     """
 
     targets = (GzipData,)
-    external_dependencies = (PIGZ,)
```
whyitfor, line 90: So basically this will run even if PIGZ isn't installed; OFRAK does not "track" this third party dependency.

This seems like an OK tradeoff; we can put it in our Docker image, and people can use this even if it is not installed.

(Writing this mainly to document this for posterity).

whyitfor (approved):
Looks good to me.

Especially like the profiling and data-driven approach.

ofrak_core/CHANGELOG.md
```diff
@@ -31,7 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 - By default, the ofrak log is now `ofrak-YYYYMMDDhhmmss.log` rather than just `ofrak.log` and the name can be specified on the command line ([#480](https://github.com/redballoonsecurity/ofrak/pull/480))
-- In `GripUnpacker`, use `gzip.GzipFile` python unpacker for speed, fall back on `pigz` if needed ([#472](https://github.com/redballoonsecurity/ofrak/pull/472))
+- In `GripUnpacker`, use the standard python `zlib` library to compress small files and decompress all files. Use `pigz` if it is installed to compress files 1MB and larger. ([#472](https://github.com/redballoonsecurity/ofrak/pull/472) and [#485](https://github.com/redballoonsecurity/ofrak/pull/485))
```
whyitfor, line 34:
```suggestion
- In `GzipUnpacker`, use the standard python `zlib` library to compress small files and decompress all files. Use `pigz` if it is installed to compress files 1MB and larger. ([#472](https://github.com/redballoonsecurity/ofrak/pull/472) and [#485](https://github.com/redballoonsecurity/ofrak/pull/485))
```

rbs-jacob (approved):
I did not review the tests in great detail, but I like the way that this implementation ended up. Seems like a big improvement!

ofrak_core/ofrak/core/gzip.py
```diff
@@ -41,44 +53,27 @@ class GzipUnpacker(Unpacker[None]):
...
+
+        # wbits > 16 handles the gzip header and footer
+        # We need to create a zlib.Decompress object in order to use this
+        # parameter in Python < 3.11
+        decompressor = zlib.decompressobj(wbits=16 + zlib.MAX_WBITS)
+        while data.startswith(b"\037\213"):
+            chunks.append(decompressor.decompress(data))
+            if decompressor.eof:
+                break
+            data = decompressor.unused_data.lstrip(b"\0")
```
rbs-jacob, line 71: I think this comment is very unclear as it stands. My understanding is that it means "this is why we're using `zlib.decompressobj`," but it reads like "this might not work in Python 3.11 because we aren't using a `zlib.Decompress` object."

I would also amend the comment to cite a source (or include justification) for why we're doing decompression this way here. To a future reader, I don't think it will be obvious why we're decompressing data that begins with these magic bytes in chunks like this.

rbs-jacob (changes requested):
I relented and included one minor change that I would like to see before merging this.

rbs-jacob (approved):
Thanks for this fix. Looks ready to merge once the tests pass!

PR #486, opened 2024-07-22
Fix usages of NamedTemporaryFiles on Windows

alchzh:

Fixed usages of NamedTemporaryFiles on Windows to avoid permissions errors.

A common pattern in the OFRAK codebase is opening a `tempfile.NamedTemporaryFile`, writing some data to it, calling `flush()`, then calling an external tool on the temporary file. Unfortunately, this always causes a permissions error on Windows because two processes can't have the same file open at the same time. The solution is to close the temporary file, without deleting it, before handing it off to the external tool. This behavior wasn't supported by `tempfile.NamedTemporaryFile` before Python 3.12 introduced the `delete_on_close` parameter (python/cpython#58451).

This PR
- copies [`tempfile.py` from Python 3.12](https://github.com/python/cpython/blob/3.12/Lib/tempfile.py) to `ofrak/tempfile.py` and sets `delete_on_close` to default to `False` in `NamedTemporaryFile`.
- changes imports of standard library `tempfile` to `ofrak.tempfile`
- changes `.flush()` to `.close()` inside `with NamedTemporaryFile()` blocks
- safely reopens files for reading when required
- Changes `ZstdPacker` and `ZstdUnpacker` to use stdin/stdout directly instead of writing to temporary files

@whyitfor


rbs-jacob:
Once the [`tempfile312` package](https://github.com/rbs-forks/tempfile312) being pushed to PyPI, this PR should be updated to use it. After that, it will be ready for review?

alchzh:
Test is failing because of the old mypy running in the tests not understanding `Self`, I think I'll change the type annotations in `tempfile312` to use strings for the current class instead of `Self`.

alchzh:
Ready for review

ofrak_core/ofrak/resource.py
```diff
@@ -1527,6 +1530,21 @@ async def summarize_tree(
...
+    async def temp_to_disk(
+        self,
+        prefix: Optional[str] = None,
+        suffix: Optional[str] = None,
+        dir: Optional[str] = None,
+        delete: bool = True,
+    ) -> AsyncIterator[str]:
+        with tempfile.NamedTemporaryFile(
+            mode="wb", prefix=prefix, suffix=suffix, dir=dir, delete_on_close=False, delete=delete
+        ) as temp:
```
rbs-jacob, line 1543: Do you think it makes sense to let additional keyword arguments fall through here? I lean towards passing along all `kwargs`, but don't feel strongly either way.

ofrak_core/test_ofrak/components/test_cpio_component.py
```diff
@@ -46,12 +46,9 @@ async def repack(self, cpio_resource: Resource) -> None:
...
     async def verify(self, repacked_cpio_resource: Resource) -> None:
-        resource_data = await repacked_cpio_resource.get_data()
-        with tempfile.NamedTemporaryFile() as temp_file:
-            temp_file.write(resource_data)
-            temp_file.flush()
+        async with repacked_cpio_resource.temp_to_disk() as temp_path:
             with tempfile.TemporaryDirectory() as temp_flush_dir:
-                command = f"(cd {temp_flush_dir} && cpio -id < {temp_file.name})"
+                command = f"(cd {temp_flush_dir} && cpio -id < {temp_path})"
                 subprocess.run(command, check=True, capture_output=True, shell=True)
```
rbs-jacob, line 52: This should be converted to use a list of arguments instead of a single string. I know you're keeping it consistent with the way it was done before, but there is much lower risk of command injection if we change `cwd` via the `subprocess.run` call instead of with `cd` in a subshell. 

Moreover, this should use the `asyncio` subprocess call instead of a blocking one.

rbs-jacob (changes requested):
Some small changes requested, but otherwise this looks like a good step in the right direction for portability.

I wonder if, in the future, we should move towards having one main temp directory created per OFRAK run. Then everything that now makes its own temp file or directory can live in there, and we can get rid of everything in the main temp directory all at once. It would be kind of like an arena allocation strategy for temp files. I suspect it might make debugging certain things easier if we can look at all of the extracted files in one place, and it may also reduce the risk of tempfile leaks in the code (if people don't use context managers, for example). No action needed on this, but I'm curious to hear your thoughts.

ofrak_core/ofrak/resource.py
alchzh: the only other argument we _might_ to pass is `dir`. We don't want to pass stuff like `encoding` through, since it specifically is creating a copy of the resource bytes on disk.

alchzh: wait I'm already passing `dir`. yeah in that case we shouldn't pass through anything else

rbs-jacob (approved):

PR #487, opened 2024-07-22
Improve Windows compatibility of `ofrak_core` tests

alchzh:

Several changes to make the `ofrak_core` test suite more Windows- :
- Add a `requires_deps_of` mark that skips tests automatically when external tool dependencies are not installed
- Skips various tests that are basically guaranteed to fail on Windows using a `skipif_windows` fixture and `skipif_windows` mark
- Leave the precompiled executable `string_test.out` used in `test_string.py` in the test assets directory instead of compiling it every time with GCC.
- Test using the python executable in `sys.exectuable` instead of the POSIX specific `cat` and `echo` in `test_component_external_tool.py`


whyitfor:
@alchzh, looks like there is a test coverage regression here:
```
=========== 1214 passed, 3 skipped, 18 warnings in 355.92s (0:05:55) ===========
fun-coverage --cov-fail-under=100
Name                                                                               Funcs   Miss   Cover   Missing
-----------------------------------------------------------------------------------------------------------------
...
ofrak/component/abstract.py                                                           11      1     91%   65
...
-----------------------------------------------------------------------------------------------------------------
TOTAL                                                                               1079      1    100%   

FAIL Required function coverage of 100.00% not reached. Total coverage: 99.91%
make[1]: *** [Makefile:19: test] Error 1
make: *** [Makefile:18: test_ofrak_core] Error 2
make[1]: Leaving directory '/ofrak_core'
```

ofrak_core/test_ofrak/components/test_filesystem_component.py
```diff
@@ -46,12 +47,15 @@ def __enter__(self):
...
-            os.makedev(1, 2)
+        if hasattr(os, "mkfifo"):
+            child_fifo = os.path.join(temp_dir, FIFO_PIPE_NAME)
+            if not os.path.exists(child_fifo):
+                os.mkfifo(child_fifo)
+
+        if hasattr(os, "mkdev"):
+            block_device = os.path.join(temp_dir, DEVICE_NAME)
+            if not os.path.exists(block_device):
+                os.makedev(1, 2)
```
whyitfor, line 58: Can you comment these to make it clear this is for Windows testing?

ofrak_core/pytest_ofrak/mark.py
```diff
@@ -0,0 +1,41 @@
...
+    pytest mark for skipping test function, module, or class if not all dependencies of a component
+    are satisfied.
+    """
+
+    uninstalled_data = asyncio.run(all_uninstalled_deps(args))
+    reason = "Missing deps:\n"
+    if uninstalled_data:
+        for component, deps in uninstalled_data.items():
+            reason += f"{', '.join(dep.tool for dep in deps)} of {component.__name__ }"
+    return pytest.mark.skipif(uninstalled_data, reason=reason)
```
whyitfor, line 41: I wonder if `requires_deps_of` should be configurable via the commandline.

In order to run testing on Windows, we most likely will need another test target other than `make test`, since any of the components that are untested will not have functional coverage. In doing so, however, we are using test coverage as a stopgap to test components -- one could image a situation where tests don't run because the dependency is incorrectly installed.

Having optional pytest cli argments to run/not run tests based on installed dependencies might make this decision clearer. (Either way, it seems we will need a separate test target for Windows.

ofrak_core/pytest_ofrak/mark.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+async def all_uninstalled_deps(
+    components: Sequence[Type[AbstractComponent]],
+) -> Mapping[Type[AbstractComponent], Sequence[ComponentExternalTool]]:
+    uninstalled = await asyncio.gather(*(uninstalled_deps(component) for component in components))
+    return {component: deps for component, deps in zip(components, uninstalled) if deps}
+
+
+def requires_deps_of(*args: Type[AbstractComponent]):
```
whyitfor, line 30: Right now the docstring does a much better job of describing what this marker does: it skips tests if dependencies are not installed. Can you pick a name that more accurately describes this?

ofrak_core/test_ofrak/components/assets/string_test.c
```diff
@@ -0,0 +1,38 @@
+
```
whyitfor, line 1: Can you add a README or a breadcrumb describing how you built string_test.out? Might be trivial, but it will be helpful down the line for reproducibility.

whyitfor (changes requested):
See comments.

ofrak_core/test_ofrak/components/test_elf_load_aligment_modifier.py
```diff
@@ -96,6 +99,7 @@ async def verify_modifier_result(test_case: ElfModifierTestCase, output_path: st
         assert process.stdout == test_case.expect_stdout
 
 
+@skipif_windows()
```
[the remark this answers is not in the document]
alchzh: not sure what you mean here? Although I see that I duplicated `skipif_windows` in this file, I can get rid of one of them if that's what you mean

ofrak_core/pytest_ofrak/mark.py
alchzh: Rewrote this in pytest's recommended format for custom markers and only applies when `--skip-tests-missing-deps` is specified on the command line

ofrak_core/pytest_ofrak/mark.py
alchzh: renamed to `skipif_missing_deps` (and using pytest's custom marker mechanism)

ofrak_core/ofrak/component/abstract.py
```diff
@@ -96,8 +106,14 @@ async def run(
             # Check if the problem was that one of the dependencies is missing
             missing_file = e.filename
             for dep in self.external_dependencies:
-                if dep.tool == missing_file:
+                if missing_file:
+                    if dep.tool == missing_file:
+                        raise ComponentMissingDependencyError(self, dep)
+                # on Windows a filename is not provided from subprocess FileNotFoundError, so just
+                # assume the any missing tool we find is the problem
+                elif not await dep.is_tool_installed():
```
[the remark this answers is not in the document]
alchzh: I assumed there was some motivation for the original behavior (minor performance benefit?), but since okay with the change I'll make it

whyitfor:
@rbs-jacob, can you review this?

ofrak_core/test_ofrak/components/test_elf_load_aligment_modifier.py
```diff
@@ -96,6 +96,7 @@ async def verify_modifier_result(test_case: ElfModifierTestCase, output_path: st
         assert process.stdout == test_case.expect_stdout
 
 
+@pytest.mark.skipif_windows
```
rbs-jacob, line 99: This test will also fail on macOS, but it will only be skipped on Windows, correct? 

Since this PR is Windows-specific, no need to do anything here. But probably worth making another issue documenting that these tests should only run on Linux.

ofrak_core/test_ofrak/components/test_extfs_component.py
```diff
@@ -16,6 +16,9 @@
 import test_ofrak.components
 
 
+pytestmark = pytest.mark.skipif_missing_deps([ExtUnpacker])
```
rbs-jacob, line 19: Could be me not understanding something pytest is implicitly doing, but why declare this variable here? I assumed the mark was a decorator, but I don't see this `pytestmark` variable being used to decorate any functions

rbs-jacob: Same comment applies to the creation of a similar variable in the tar tests

rbs-jacob (changes requested):

rbs-jacob:
This should also have a changelog entry! Thanks

ofrak_core/test_ofrak/components/test_extfs_component.py
alchzh: This is the pytest way to skip an entire test file/module

https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules

ofrak_core/pytest_ofrak/mark.py
```diff
@@ -0,0 +1,95 @@
...
+        "--skip-tests-missing-deps",
+        action="store_true",
+        default=False,
+        help="skip tests whose required external dependencies are missing",
+    )
+
+
+def pytest_configure(config):
+    # the skipif_missing_deps mark has to take an iterable rather than star args otherwise pytest's decorator magic will
+    # think the mark was applied to the component class not the actual test function
```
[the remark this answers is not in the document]
alchzh: the next line declares the `skipif_missing_deps` marker. I guess it could go in the help string for the marker instead

ofrak_core/pytest_ofrak/mark.py
```diff
@@ -0,0 +1,95 @@
...
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--skip-tests-missing-deps"):
+        deps_installed_data = asyncio.run(
+            _check_deps_installed(
+                [
+                    dep
+                    for item in items
+                    for component in _all_required_components(item)
+                    for dep in component.external_dependencies
+                ]
```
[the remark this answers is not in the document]
alchzh: there's not many duplicates, but there's also no reason not to use a `set`

alchzh:
> This should also have a changelog entry! Thanks

I don't see other test suite updates in the changelog, is it required here?

ofrak_core/test_ofrak/components/test_filesystem_component.py
rbs-jacob, line 1: I can't seem to add a comment to these specific lines since they're no longer part of the PR change set, but there is an error here:

https://github.com/redballoonsecurity/ofrak/blob/30b70b841cf0b87568411047b10517c24fa7787b/ofrak_core/test_ofrak/components/test_filesystem_component.py#L55-L58

It's checking for `hasattr(os, "mkdev")` but calls `os.makedev` later in the script.

rbs-jacob (changes requested):
Responding to the above question: even though there may not be other test suite changes in the changelog, we should set a precedent now for adding those things to the changelog. There is no real reason to exclude those things, and it may be helpful for safety certification stuff.

ofrak_core/pytest_ofrak/elf/fixtures.py
```diff
@@ -149,13 +150,13 @@ def large_elf_source_file(elf_test_directory):
...
 @pytest.fixture
-def large_elf_object_file(elf_test_directory):
+def large_elf_object_file(elf_test_directory, skipif_windows):
     subprocess.run(["make", "-C", elf_test_directory, "large_elf.o"])
     return os.path.join(elf_test_directory, "large_elf.o")
 
 
 @pytest.fixture
-def large_elf_file(elf_test_directory):
+def large_elf_file(elf_test_directory, skipif_windows):
```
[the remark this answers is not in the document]
alchzh: I've put them in an assets/ folder with LFS

alchzh:
changelog entry added

ofrak_core/pytest_ofrak/elf/fixtures.py
rbs-jacob, line 1: Is all of the commented out stuff still supposed to be in this file?

rbs-jacob (approved):

Issue #491, opened 2024-08-20
Have `File` inherit from `GenericBinary`


whyitfor:
While reviewing https://github.com/redballoonsecurity/ofrak/pull/469, I noticed that most Identifiers target `File` and `GenericBinary`. The impact of this is that OFRAK tries to run identifiers twice against any `File`: first it runs identifiers since it is a `File`, then re-runs the identifier when the `GenericBinary` tag is added.

Since all files are binaries (by definition), it seems reasonable to refactor `File` as such:

```
class File(GenericBinary, FilesystemEntry)
```

and have identifiers only target `GenericBinary`

**Which files would be affected?**
Identifiers, and `File`.

**Does the proposed maintenance include non-doc string _functional_ changes to the Python code?**
Perhaps

**Are you interested in implementing it yourself?**
Yes

whyitfor:
File used to test these changes: http://tinycorelinux.net/15.x/x86/release/TinyCore-current.iso

Profiling script used to help inform this change:

```
import asyncio
import logging
import selectors
import sys
import time

from ofrak import OFRAK, OFRAKContext  # type:ignore
from ofrak.core import Elf
import os
#os.environ["PYTHONASYNCIODEBUG"] = "1"

# Borrowed from https://gist.github.com/vxgmichel/620eb3a02d97d3da9dacdc508a5d5321
class TimedSelector(selectors.DefaultSelector):
    select_time = 0

    def reset_select_time(self):
        self.select_time = 0

    def select(self, timeout=None):
        if timeout is not None and timeout <= 0:
            return super().select(timeout)
        start = time.perf_counter_ns()
        try:
            return super().select(timeout)
        finally:
            self.select_time += time.perf_counter_ns() - start


class TimedEventLoopPolicy(asyncio.DefaultEventLoopPolicy):
    def new_event_loop(self):
        selector = TimedSelector()
        return asyncio.DefaultEventLoopPolicy._loop_factory(selector)


def mystrtime(ns: int) -> str:
    if ns > 1_000_000_000:
        return f"{ns/1_000_000_000:.2f}s"
    else:
        return f"{ns/1_000_000:.1f}ms"


async def unpack_one(ofrak_context: OFRAKContext, path: str) -> None:
    root = await ofrak_context.create_root_resource_from_file(path)
    await root.unpack_recursively(do_not_unpack=[Elf])
    await root.auto_run_recursively(all_identifiers=True)


async def main() -> None:
[...]

PR #492, opened 2024-08-21
Refactor `File` to inherit from `GenericBinary`

whyitfor:

This PR contains 4 important changes:

1. `File` now inherits from `GenericBinary` to avoid duplicative Identifier runs
2. Update auto-run component logic to run all Analyzers, not just the most specific ones
3. Refactor Magic identification for significant speed improvements
  - Refactor magic identification to use one identifier, named `MagicIdentifier`
  - Rename `MagicMimeIdentifier` to `MagicMimePattern`, as it is run by `MagicIdentifier`
  - Rename `MagicDescriptionIdentifier` to `MagicDescriptionPattern`, as it is run by `MagicIdentifier`
  - Add `RawMagicPattern` to efficiently run custom magic byte search logic within `MagicIdenfifier`
4. Update registered identifiers to make use of new `MagicIdentifier` for following resource tags: `Apk`, `Bzip2Data`, `CpioFilesystem`, `DeviceTreeBlob`, `Elf`, `Ext2Filesystem`, `Ext3Filesystem`, `Ext4Filesystem`, `GzipData`, `ISO9660Image`, `Jffs2Filesystem`, `LzmaData`, `XzData`, `LzoData`, `OpenWrtTrx`, `Pe`, `RarArchive`, `SevenZFilesystem`, `SquashfsFilesystem`, `TarArchive`, `Ubi`, `Ubifs`, `Uf2File`, `UImage`, `ZipArchive`, `ZlibData`, `ZstdData`  

**Link to Related Issue(s)**
#491 

The four changes in this PR result in significant speed improvements to OFRAK workflows that involve a large amount of automatic unpacking.

The following TinyCore file and profiling script were used to inform these changes: https://github.com/redballoonsecurity/ofrak/issues/491#issuecomment-2568472735.

On master (8af61457d14401c8a199c57b8d538d44f911e919), this script takes ~24 seconds to run:
```
Time in /TinyCore-current.iso unpack x2: 24.50s wall, 24.31s CPU, 122.0ms async select, 69.4ms blocked
```

With this PR, this time decreases to ~10.7 seconds:
```
Time in /TinyCore-current.iso unpack x2: 10.78s wall, 10.60s CPU, 125.6ms async select, 50.9ms blocked
```

This speedup is achieved by primarily by **minimizing** the number of identifiers that are run. On master, identifiers are run 50,312 times, compared to 5,852 after this PR:

1. Before this change, magic identifiers were triggered by both the `GenericBinary` and `File` tag. This resulted in situations where all identifiers were run twice -- first against a `GenericBinary`, and then against a `File` after that tag was added. Having `File` inherit from `GenericBinary` allows us to have magic identifiers target `GenericBinary` only. This semantically makes sense as well: all files are by definition generic binaries.  
2. Having `MagicMimeIdentifier` and `MagicDescriptionIdentifier` as separate identifiers incurred unnecessary component overhead, even though this was implemented with both identifiers using the same Magic analysis. Combining these into `MagicIdentifier` resulted in a roughly 20-30% speedup.
3. Creating the `RawMagicPattern` allowed for consolidating all raw magic pattern matching to be run inside of the `MagicIdentifier`. This reduces the amount of times data is copied around, and helps greatly reduce the number of times identifiers are run. This accounted for the bulk of the remaining speedup.

@rbs-jacob, @ANogin


ofrak_core/ofrak/core/magic.py
```diff
@@ -73,63 +73,102 @@ async def analyze(self, resource: Resource, config=None) -> Magic:
             return Magic(magic_mime, magic_description)
 
 
-class MagicMimeIdentifier(Identifier[None]):
-    """
-    Identify and add the appropriate tag for a given resource based on its mimetype.
+class MagicIdentifier(Identifier[None]):
     """
+    Identify resourced using three identifier patterns:
```
rbs-jacob, line 78:
```suggestion
    Identify resources using three identifier patterns:
```

ofrak_core/ofrak/core/magic.py
```diff
@@ -73,63 +73,102 @@ async def analyze(self, resource: Resource, config=None) -> Magic:
...
 
+    @staticmethod
+    def _run_magic_description_pattern(resource: Resource, magic_description: str):
+        for matcher, resource_type in MagicDescriptionPattern.matchers.items():
+            if matcher(magic_description):
+                resource.add_tag(resource_type)
+
+    @staticmethod
+    async def _run_raw_magic_patterns(resource: Resource):
+        data_length = min(await resource.get_data_length(), 8)
```
rbs-jacob, line 116: Why run the raw magic pattern matchers only against the first 8 bytes? I personally think 32 or 64 would be better.

Should it be documented somewhere that matcher functions will only get the first few bytes, and not all of the data? I would find this surprising as a user, and would expect it to be clearly documented somewhere.

docs/contributor-guide/component/identifier.md
```diff
@@ -1,14 +1,20 @@
...
+MagicDescriptionPattern.register(
+    UImage, lambda s: s.startswith("u-boot legacy uImage")
+)
 ```
 
-This line ensures that the `MagicDescriptionIdentifier` adds a `UImage` tag to resources matching that description pattern. As a result, any unpackers targeting a `UImage` will automatically run when `Resource.unpack` is run.
+This pattern (along with all other identifier patterns) will get run when the [MagicIdentifier][ofrak.core.magic.MagicIdentifier] runs, resulting in a `UImage` tag to resources matching the description.
 
+Registering magic patterns is a good first choice, since it is the most performance-efficient.
+When required, it is also possible to write custom identifiers. For an example of this, see [ApkIdentifier][ofrak.core.apk.ApkIdentifier].
```
rbs-jacob, line 17: We should add more docs here with examples of the other two magic patterns. For example, there are some aspects of the raw bytes pattern matcher that are pretty unintuitive (see comment below).

This is also a good place to add information about when to use patterns and when not to (perhaps pointing to the APK as an example of when the magic pattern thing isn't quite sufficient).

rbs-jacob (changes requested):
- [ ] Improve documentation
- [ ] Address suggestions below
- [ ] Changelog entry

ofrak_core/ofrak/core/__init__.py
```diff
@@ -36,6 +36,7 @@
 from ofrak.core.injector import *
 from ofrak.core.instruction import *
 from ofrak.core.iso9660 import *
+from ofrak.core.java import JavaArchive
```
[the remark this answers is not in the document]
rbs-jacob: That's really odd. Worth adding a comment in this file that mentions this so someone in the future has this context.

docs/contributor-guide/component/identifier.md
whyitfor: Updated documentation.

whyitfor (commented):
@rbs-jacob, thanks for the feedback! I believe I've addressed everything, ready for another review.

ofrak_core/ofrak/core/dtb.py
```diff
@@ -402,5 +390,14 @@ def _prop_from_fdt(p: fdt.items.Property) -> Tuple[DtbPropertyType, bytes]:
...
-MagicMimeIdentifier.register(DeviceTreeBlob, "Device Tree Blob")
-MagicDescriptionIdentifier.register(DeviceTreeBlob, lambda s: "device tree blob" in s.lower())
+MagicMimePattern.register(DeviceTreeBlob, "Device Tree Blob")
+MagicDescriptionPattern.register(DeviceTreeBlob, lambda s: "device tree blob" in s.lower())
+
+
+def match_dtb_magic(data: bytes):
+    if len(data) < 4:
+        return False
+    return data == struct.pack("<I", DTB_MAGIC_SIGNATURE)
```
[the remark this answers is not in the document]
whyitfor: Updated dtb (https://github.com/redballoonsecurity/ofrak/pull/492/commits/4d95dd277d0e9fe52ea542721ff908f726af9ea8) as well as uf2 (https://github.com/redballoonsecurity/ofrak/pull/492/commits/f79f876723708542a9676ce999a0dcd1b2e01fea)

rbs-jacob (changes requested):
Two small typos introduced in the last commits. Pending those fixes, any outstanding threads, and the changelog merge conflict resolution, this should be good to go.

docs/contributor-guide/component/identifier.md
```diff
@@ -1,17 +1,50 @@
...
+## Implement a New Identifier
+Additionally, it is possible to implement a new [Identifier][ofrak.component.identifier.Identifier].
+Doing so should be reserved for situations where extending the [MagicIdentifier][ofrak.core.magic.MagicIdentifier] is impractical.
+The [ApkIdentifier][ofrak.core.apk.ApkIdentifier] is an example of a custom identifier implementation.
 
-### Handling External Dependencies
+!!! warning
+    Adding new identifiers should be done with care to minimize overall performance impact to OFRAK workflows.
+    Try to carefully select the resource tags the identifier targets to minimize the frequency with which
+    it is run: generally speaking, targeting `GenericBinary` with result in this identifier getting run on the largest
```
rbs-jacob, line 44:
```suggestion
    it is run: generally speaking, targeting `GenericBinary` will result in this identifier getting run on the largest
```

rbs-jacob (changes requested):

rbs-jacob (approved):
Looks good to me. Excited about this change!

PR #496, opened 2024-09-09
Ghidra instruction decode fails if the VLE register is available but the value is null.

dannyp303:

Fall back to PPC instruction encoding when the VLE register is available but the value is null. 
**Link to Related Issue(s)**

### DIFF
--- a/disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/GetInstructions.java
+++ b/disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/GetInstructions.java
@@ -118,7 +118,9 @@ public class GetInstructions extends HeadlessScript {
             Register vle_register = instruction.getRegister("vle");
             if (vle_register != null) {
                 BigInteger vle_val = instruction.getValue(vle_register, false);
-                this.instruction_mode = vle_val.equals(BigInteger.ONE) ? "VLE" : this.instruction_mode ;
+                if (vle_val != null){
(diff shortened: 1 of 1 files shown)


disassemblers/ofrak_ghidra/ofrak_ghidra/ghidra_scripts/GetInstructions.java
```diff
@@ -118,7 +118,9 @@ class ResultInstruction {
             Register vle_register = instruction.getRegister("vle");
             if (vle_register != null) {
                 BigInteger vle_val = instruction.getValue(vle_register, false);
-                this.instruction_mode = vle_val.equals(BigInteger.ONE) ? "VLE" : this.instruction_mode ;
+                if (vle_val != null){
+                    this.instruction_mode = vle_val.equals(BigInteger.ONE) ? "VLE" : this.instruction_mode ;
+                }
```
rbs-jacob, line 123: I would change it to either entirely use the ternary, or not use it at all.

```suggestion
                if (vle_val != null && vle_val.equals(BigInteger.ONE)){
                    this.instruction_mode = "VLE";
                }
```

or 

```suggestion
                this.instruction_mode = vle_val != null && vle_val.equals(BigInteger.ONE) ? "VLE" : this.instruction_mode ;
```

rbs-jacob (changes requested):

rbs-jacob:
Also if there is a specific binary that triggers this null pointer exception, should we write a test to catch regressions? No need to block this PR on that if it's complicated, but if it's easy to do, we should add it in.

whyitfor:
Marked as draft to make it clear this is a WIP

PR #498, opened 2024-09-09
Add `package-lock.json` to version control and use `npm ci` for consistent frontend dependencies. Fixes #493

### LINKED ISSUE #493: Frontend does not include `package-lock.json`
Currently, the `package-lock.json` file in frontend is `.gitignore`d. This opens the door to every developer getting a different combination of dependency versions, which is different than the ones being run in production or (in the future) checked in the CI.

This file [is intended to be committed to source repositories](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json). v. https://stackoverflow.com/questions/44297803/what-is-the-role-of-the-package-lock-json

In non development (production, CI) contexts [`npm ci`](https://docs.npmjs.com/cli/v10/commands/npm-ci) should be used instead of `npm install`. This command installs the exact dependencies specified in `package-lock.json`.

`frontend/.gitignore`, [docker files and scripts using `npm install`](https://github.com/search?q=repo%3Aredballoonsecurity%2Fofrak%20npm%20install&type=code) 

No.

Sure.

alchzh:

**Link to Related Issue(s)**
#493

See #493

@rbs-jacob

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install frontend linting dependencies
         run: |
           cd frontend
-          npm install
+          npm ci
           sudo apt-get install shellcheck
           python3 -m pip install black==23.3.0
       - name: Lint frontend
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -6,7 +6,6 @@
 .vscode
 
 node_modules
-package-lock.json
 **/.DS_Store
 src/resource.js
 dist
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,0 +1,999 @@
+{
+  "name": "ofrak-app",
+  "version": "3.2.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "ofrak-app",
+      "version": "3.2.0",
+      "dependencies": {
+        "@splinetool/runtime": "^0.9.30",
+        "highlight.js": "^11.7.0",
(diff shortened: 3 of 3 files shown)


frontend/Dockerstage
```diff
@@ -3,4 +3,4 @@ FROM node:latest AS svelte
 COPY --chown=node:node $PACKAGE_DIR /home/node/frontend
 
 WORKDIR /home/node/frontend
-RUN su node -c "npm install && npm run build"
+RUN su node -c "npm ci && npm run build"
```
rbs-jacob, line 6: Using this (and a `package-lock.json`) makes sense to me in general, but I don't know if it makes sense to use here specifically. 

This is used not just for CI, but also for building regular Docker images. I can imagine a world in which someone building a Docker image that uses the front end doesn't want the "fail fast" approach of the `npm ci` command and would instead prefer `npm install`. 

Same comment applies to the other change in this PR to switch `npm install` to `npm ci` in the `ofrak_core/Makefile`.

rbs-jacob: I agree that it should cause an error in CI and before code is merged. But it seems overly strict to impose on local development. 

In my estimation, we are more likely to get a bug report from someone who cannot build their image due to hard failures than we are from someone who has GUI bugs due to minor dependency version mismatches. 

Our main users are security researchers, not software engineers. Sometimes, allowing them to play fast and loose during development is critical for reducing friction when using OFRAK (which is already higher than I would like). 

I still think we should not use `npm ci` everywhere. We should use it only in CI.

rbs-jacob (approved):

PR #505, opened 2024-10-24
Improved allocation of .bss(-like) sections

alchzh:
The existing strategy for allocating `.bss` in RAM has a few problems
 - only one contiguous `.bss` region per FEM is supported.
 - alignment was not properly calculated for `.bss`, leading to linker errors

This commit generalizes our process for allocating `.text`, `.data`, `.rodata`, etc. to `.bss` and like as well. It also improves our support for different platforms by reducing the use of hardcoded section names for determining behavior in favor of attributes defined by the binary format like `sh_type` and `sh_flags`

With the new system, instead of passing `unsafe_bss_segment` to `make_fem`, `.bss` sections are allocated during the `allocate_bom` step. This is supported by free space regions that don't map to any data on the parent resource. The new code allows `.bss` regions to be allocated in existing `RW` regions that are mapped to data by zeroing out those areas.

Changes (incomplete):
- Use SHT_NOBITS type to detect .bss(-like) sections on ELF toolchains
- Create `FreeSpaceWithoutData` tag for allocating regions for .bss. These are created from `MemoryRegion` children of resources with `data_range=None`
- Use SHF_ALLOC flag to determine whether section needs to be patched or not instead of hard coded list of names.
- Pass flag to compiler to prevent `.eh_frame` generation
- Backwards compatible with old bss allocation for now, emits deprecation warnings.


ofrak_core/ofrak/core/free_space.py
```diff
@@ -360,22 +422,47 @@ async def modify(self, resource: Resource, config: FreeSpaceAllocation) -> None:
 
         for fs in wholly_allocated_resources:
             fs.resource.remove_tag(FreeSpace)
+            fs.resource.remove_tag(RuntimeFreeSpace)
+
+            # Without committing the changes to our tags here, removing both the child and
+            # parent tags will cause a KeyError on update.
+            await fs.resource.save()
+            fs.resource.remove_tag(AnyFreeSpace)
```
alchzh, line 430: see #507 for why this hack is needed

rbs-jacob: Can this be removed now that #513 is merged, or is it still necessary?

alchzh: yes, removed the workaround

alchzh:
test cases done

dannyp303:
This is good to merge IMO, after merge conflicts are done, sorry for the delayed response.

dannyp303 (approved):
LGT$

PR #508, opened 2024-10-30
Add LLM analyzers

rbs-jacob:

Add AI-enhanced LLM analyzers.

@whyitfor


rbs-jacob:
Outstanding to do items:

- [ ] Push LLM-enhanced program/function analyzers
- [ ] Update changelog

ofrak_core/ofrak/core/llm.py
```diff
@@ -0,0 +1,172 @@
...
+                result.append(f"{k}: {s}")
+        return "\n".join(result)
+
+
+async def dump_attributes(resource: Resource) -> str:
+    model = resource.get_model()
+    # The data summary is un-informative and verbose
+    if DataSummary in model.attributes:
+        del model.attributes[DataSummary]
+    return serialize(make_serializable(model.attributes))
```
whyitfor, line 167: Isn't this essentially a re-implementation of what the `PJSONSerializationService` does? Is there a reason to not use it?

rbs-jacob: PJSON has types alongside the values. This is useful for deserializing PJSON values, but would make this code overly complicated, since we never need to deserialize the resulting text and thus would end up adding logic to drop the types anyway. 

Moreover, PJSON is nice as a wire format, but is not optimized for human-readable text "serialization." We want to deviate from its representation of some values to make the resulting text easier for the LLM to understand.

I also (personally) just don't like the PJSON serialization service.

ofrak_core/test_ofrak/components/test_llm_components.py
```diff
@@ -0,0 +1,71 @@
...
+    await proc.communicate()
+    yield model_name
+
+
+@pytest.fixture(autouse=True)
+def ghidra_components(ofrak_injector):
+    ofrak_injector.discover(ofrak_ghidra)
+
+
+async def test_llm_component(ofrak_context: OFRAKContext, model: str):
```
whyitfor, line 45: Can we use the https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/test_ofrak/unit/component/analyzer/analyzer_test_case.py#L34 to run these tests?

Right now these tests are not validating that the analyzer successfully generates the needed attributes.

ofrak_core/test_ofrak/components/test_llm_components.py
```diff
@@ -0,0 +1,71 @@
...
+        LlmAnalyzer,
+        LlmAnalyzerConfig(
+            "http://localhost:11434/api/chat",
+            model,
+        ),
+    )
+    root.get_attributes(LlmAttributes)
+
+
+async def test_llm_program_component(ofrak_context: OFRAKContext, model: str):
```
whyitfor, line 60: In addition, let's create a test case for the function analyzer.

ofrak_core/ofrak/core/llm.py
whyitfor: Are we sure that the serialized PJSON would hinder the LLM analysis? This sounds like a hypothesis and not something that was tested.

We should, whenever possible, avoid writing ~80 lines of code if it can be avoided.

whyitfor (commented):
See commenda

ofrak_core/Makefile
```diff
@@ -28,5 +28,6 @@ ofrak/gui/public:
 		npm install && \
 		npm run build && \
 		cd ../ofrak_core && \
+		rm -rf ofrak/gui/public ; \
```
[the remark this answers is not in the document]
rbs-jacob: You can force a make target to run, even if the directory exists. In that case the public directory will be copied to `ofrak/gui/public/public` which is not what we want. If the directory doesn't already exist, the `rm -rf` should not cause any problems, so I don't really see a drawback to adding it here.

```
jacob@jacob % rm -rf /tmp/fake/does/not/exist && echo true || echo false
true
```

ofrak_core/ofrak/core/llm.py
rbs-jacob: I could surely get something working with PJSON. The (admittedly untested) hypothesis is that the code to do so would be as long as this code, less clear than this code, and dependent on undocumented serialized PJSON structure, which may change in the future, necessitating changes here.

If you still think I should rewrite this to use PJSON, I will.

ofrak_core/Dockerstub
```diff
@@ -75,3 +76,6 @@ RUN cd /tmp && \
     make install && \
     cd /tmp && \
     rm -r UEFITool-A68
+
+# Install Ollama
+RUN curl -L "https://ollama.com/download/ollama-linux-""$TARGETARCH"".tgz" | tar -C /usr/ -xzv
```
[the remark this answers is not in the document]
rbs-jacob: How would you propose doing it instead?

It would be ideal to have this whole thing in a separate package so that users could conditionally include the AI components. Then we could install Ollama and pull a model without worrying about unnecessarily inflating the core image. But one of the original constraints was that this should be in OFRAK core.

If we don't have Ollama installed, then we can't run the tests without having the tests themselves pull down and install the binary. As far as I know, in other OFRAK tests, there is no precedent for installing a dependency like this. It would be doable, but seems a little weird.

On the other hand, we could pull a model in the OFRAK core Dockerstub after installing Ollama, but this seems like a bad idea. We don't want to add another several-hundred-MB dependency to the OFRAK core images. We especially don't want to add models this large to the base image if the user is just going to use OpenAI anyway. 

Installing Ollama, but not pulling a model, seemed like a reasonable tradeoff. If it seems wrong to you, I would appreciate guidance on a better way to do it.

ofrak_core/test_ofrak/components/test_llm_components.py
rbs-jacob: I can do that validation in these tests. It does not make sense to use `AnalyzerTests` subclass here since the analyzer must be run with `resource.run`. It requires a config, and thus cannot be called with `resource.analyze`, as happens in `AnalyzerTests` methods.

rbs-jacob: Also the tests do call `get_attributes` at the end, so they're validating that the attributes are added.

ofrak_core/test_ofrak/components/test_llm_components.py
rbs-jacob: The function analyzer is tested implicitly since it is called directly by the program analyzer component. But I have added an explicit test for it.

PR #515, opened 2024-11-13
Auto-enable Docker buildkit

ANogin:

Default to forcing Docker buildkit to be enabled.

**Link to Related Issue(s)**
N/A; supersedes #392 

Per @EdwardLarson:
> Some features in the build require Docker buildkit to be enabled, and we say as much in the docs to build the image. However, this is a point of failure (for some reason I could not get the buildkit to be enabled in my machine) and a "gotcha" that can be removed entirely by just setting the Docker buildkit enabled in the subprocess we spin up to do the Docker build. In case there is a desire to specifically NOT use the buildkit, an option is added to not alter the subprocess environment at all (buildkit may or may not be enabled depending on user's environment).

@whyitfor

### DIFF
--- a/build_image.py
+++ b/build_image.py
@@ -73,7 +73,7 @@ def main():
         f.write(dockerfile_finish)
     print(f"{FINISH_DOCKERFILE} built.")
 
-    env = {k: v for k, v in os.environ.items()}
+    env = os.environ.copy()
     if not config.no_forced_buildkit:
         env["DOCKER_BUILDKIT"] = "1"
 
--- a/build_image.py
+++ b/build_image.py
@@ -33,7 +33,6 @@ class OfrakImageConfig:
     install_target: InstallTarget
     cache_from: List[str]
     entrypoint: Optional[str]
-    no_forced_buildkit: bool
 
     def validate_serial_txt_existence(self):
         """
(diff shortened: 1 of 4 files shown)


build_image.py
```diff
@@ -33,6 +33,7 @@ class OfrakImageConfig:
     install_target: InstallTarget
     cache_from: List[str]
     entrypoint: Optional[str]
+    no_forced_buildkit: bool
```
whyitfor, line 36: Do we need this option? The Dockerstub files expect buildkit -- this process doesn't work without it.

ANogin: @whyitfor this came from Edward's original implementation. No clue whether there are any cases where it might be needed or desired...

whyitfor: As we have it, the Docker build process requires buildkit, so using `no_forced_buildkit` will not work. Based on this, I think we should just remove this option -- it's not needed.

@rbs-jacob, is there something here I am missing?

rbs-jacob: As far as I know, we should be able to remove it. 

My suspicion is that this may have been related to getting OFRAK to build with old versions of Docker running on internal blades, but I don't have any specific memories or notes indicating that – just a vibe.

Either way, even if the `DOCKER_BUILDKIT` environment variable is set, on systems too old to have it, that environment variable being set shouldn't break anything.

whyitfor:
@rbs-jacob, since I've updated this PR, can you review?

rbs-jacob (approved):

PR #520, opened 2024-11-19
Run `ofrak_patch_maker_test/test_symbol_parsing.py`

alchzh:
This test was incorrectly marked as `async def` so it never actually run in CI or locally in my experience.

See example logs here: https://github.com/redballoonsecurity/ofrak/actions/runs/11921517683/job/33225804260#step:6:380

Maintenance only change

- Removes a spurious `async def` preventing this test case from running
- Renames the `TestCase` class so pytest doesn't think it's a class with tests

@whyitfor @rbs-jacob


whyitfor (approved):
LGTM

PR #523, opened 2024-11-21
Include `node` and `npm` in Docker images with frontend (Fixes #497)

### LINKED ISSUE #497: Install `node` and `npm` into development Docker images
Currently, the Docker images include a pre-built copy of the frontend, and don't have the tooling to build the frontend themselves.

Something like

```Dockerfile
COPY --from=node:lts-bullseye-slim /usr/local/bin /usr/local/bin
COPY --from=node:lts-bullseye-slim /usr/local/lib/node_modules /usr/local/lib/node_modules
```

would be nice.

`build_image.py`?

No

Not familiar with where this should go.

alchzh:

Include `node` and `npm` in Docker images with frontend

**Link to Related Issue(s)**
#497 

- Uses `node:lts-bullseye-slim` to build frontend (can use a specific version pin as well)
- Copies `node`, `npm` and a couple of other tools into the resulting image for development

@rbs-jacob

### DIFF
--- a/frontend/Dockerstage
+++ b/frontend/Dockerstage
@@ -1,4 +1,4 @@
-FROM node:latest AS svelte
+FROM node:lts-bullseye-slim AS svelte
 
 COPY --chown=node:node $PACKAGE_DIR /home/node/frontend
 
--- a/frontend/Dockerstub
+++ b/frontend/Dockerstub
@@ -1,6 +1,8 @@
 RUN apt-get update && apt-get install --yes nginx
 
 COPY --from=svelte --chown=root:root /home/node/frontend/dist /ofrak_gui
+COPY --from=svelte /usr/local/bin /usr/local/bin
+COPY --from=svelte /usr/local/lib/node_modules /usr/local/lib/node_modules
 
 COPY $PACKAGE_PATH/nginx.conf /etc/nginx/sites-enabled/default
 


rbs-jacob:
I don't understand why we need to be able to build a new front end in Docker containers that are "for production." I also think that if we wanted to do that, it would make more sense to just install Node properly than to copy the `/usr/local/bin` and `node_modules` directories from another image.

Pinning the image that builds the front end to a specific tag is a really good idea that should have been done a while ago.

alchzh:
We also build development images off of these docker files, and I thought it'd be good to have a consistent node/npm tooling. I've been building including these lines own development docker files for months since the node version in debian repositories for bullseye is horribly out of date and EoL. Copying from their upstream /usr/local/bin only includes the basic node tooling, so we can basically merge the python and node images

rbs-jacob:
I think there are better ways to make dev images. We should not include optional developer dependencies in the main images. I agree that for developer UX, it would be nice to rebuild the GUI from within the images, but I do not think this is the right way to do it. We could instead, for example, have an additional package with its own Dockerstub that installs dev dependencies.

[Using `nvm`, it is very easy to install specific versions of Node](https://nodejs.org/en/download/package-manager/), which is an officially-supported (and much less brittle) way of installing it. It works interactively at the command line, and it also works from scripts and Dockerfiles. If we are going to install Node and npm in developer images, this should be the way to do it.

PR #559, opened 2025-01-06
Data.Offset refers to offset in data root

whyitfor:

Update `Data.Offset` such that its value is always the equivalent to a resource's start offset in the data root:

```
Range.from_size(Data.Offset, Data.Length) == await resource.get_data_range_within_root()
```

**Link to Related Issue(s)**

Before this change, the `Data` attribute was created with the data range of a parent resource, which was not always equal to the data root range.

During patching, however, the value of `Data` were updated to correspond to the root data range. This created inconsistencies when sorting by `Data`--for example, in the OFRAK GUI tree. 

`TestGrandchildrenDataAttributes` asserts that the behavior is correct.

@rbs-jacob.
@dannyp303 could review for general awareness.


rbs-jacob (approved):
Looks good to me, great catch! Merge when you're ready.

rbs-jacob (changes requested):
Very minor tweak based on the additional updates. Once that is in, this should be good to merge.

rbs-jacob (approved):
Looks good to me

Issue #549, opened 2024-12-29
Add multiarch binaries and aarch64 variants of toolchains for full ARM support


Jepson2k:
**Which files would be affected?**
Mainly dockerstubs. ARM support for ofrak dependencies have greatly increase over the year. Although I'm not finished, it looks like only the tests that deal with the BCC spark toolchain will need to be skipped on ARM. Otherwise everything else appears to have ARM variants.

**Does the proposed maintenance include non-doc string _functional_ changes to the Python code?**
Yes but not much likely just a couple of pytest.marker.skipif(platform != x86_64) for toolchains that do not support ARM.

**Are you interested in implementing it yourself?**
Yes

PR #562, opened 2025-01-07
Support for running OFRAK on arm64 Linux

alchzh:
Authored by @Jepson2k

OFRAK Docker image builds on arm64 platforms (M1, etc.).

**Link to Related Issue(s)**
#549 

This PR ensures that OFRAK Docker images build on arm64 platforms, such as the M1. All of the `ofrak_patch_maker` tests (except SPARC, which are skipped) pass on both x86_64, aarch64 platforms with this change.

**ofrak-capstone**: 1.1.0rc1
- Bump underlying `capstone` package to 5.0.3 to use pre-built aarch64 binaries

**ofrak_core**: 3.3.0rc3
- Bumped version of lief to `0.16.1`, orjson to `3.10.12` to take advantage of pre-built aarch64 binaries
- Bumped version of angr to `9.2.133` for arm binaries

**ofrak-patch-maker**: 4.1.0rc0
- Dockerstub works for x86_64, aarch64 platforms for all toolchains except SPARC

@rbs-jacob


ofrak_core/Makefile
```diff
@@ -18,7 +18,11 @@ test: inspect
 	$(PYTHON) -m pytest -n auto test_ofrak --cov=ofrak --cov-report=term-missing
 	(sleep 2; echo 1; sleep 2; echo i agree) \
 		| python3 -m coverage run --append --source ofrak -m ofrak license --force
-	fun-coverage --cov-fail-under=100
+	@if [ "$(shell uname -m)" = "x86_64" ]; then \
+		fun-coverage --cov-fail-under=100; \
+	else \
+		fun-coverage; \
+	fi
```
rbs-jacob, line 25:
```suggestion
	fun-coverage --cov-fail-under=100 || ! (shell uname -m | grep x86)
```

Same suggestion applies to all other instances of this change across the various package Makefiles below. 

I think doing it this way is a bit less brittle, and will have slightly more descriptive console output (since we'll still see coverage result failures, they just won't stop future tests from running).

alchzh:
test currently failing because the bumped angr version requires python >=3.10. I guess we can set a complex requirement for it

ofrak_patch_maker/ofrak_patch_maker/toolchain/utils.py
```diff
@@ -48,10 +51,18 @@ def get_repository_config(section: str, key: Optional[str] = None):
...
+                raw_value = config.get(section, key)
+
+                if "*" in raw_value:
+                    matches = glob.glob(raw_value)
+                    if matches:
+                        return matches[0]
+                    else:
+                        raise NotFoundError(f"No file matches wildcard {raw_value}")
+                else:
+                    return raw_value
```
rbs-jacob, line 63: I don't feel very positively about this change to use globbing. I personally feel that it makes the config file harder to read and understand. It also seems kind of error-prone for anyone not running OFRAK inside of Docker, who may have their own toolchains installed on their system.

@Jepson2k @alchzh what alternatives were considered here instead of globbing?

rbs-jacob (changes requested):

ofrak_core/Makefile
rbs-jacob: I think I may have made an error here – I don't think the `shell` part at the beginning of the subshell should be necessary...

alchzh: d'oh! should have caught that by testing it... fixed now

alchzh:
Changed it so that toolchain.conf can be read from the `/etc/ofrak/toolchain.conf`, and the Dockerstub copies over the correct file based on the platform architecture

Left the current toolchain.conf in place to avoid breaking anyone's setup if they install a newer version. I want to update it to one without absolute paths in the future though so users can use executables that exist on their local PATH

ofrak_patch_maker/Dockerstub
```diff
@@ -22,13 +24,11 @@ RUN cd /tmp && \
 
 #X64-64 toolchain for arm64 Docker images
 RUN if [ "$TARGETARCH" = "arm64" ]; then \
-    apt-get update && apt-get install -y gcc-10-x86-64-linux-gnu; \
+    apt-get update && apt-get install -y gcc-12-x86-64-linux-gnu; \
```
rbs-jacob, line 27: We should probably download an ARM version of the GCC toolchain from a link that pins the version as opposed to apt installing it. Doesn't need to block this PR if the change will take too long, but if we don't do it here, we should add an issue tracking it.

ofrak_core/pytest_ofrak/elf/assets/Makefile
```diff
@@ -1,5 +1,5 @@
-
-CC=gcc
+# Specifying the full name is necessary for some tests to work on ARM
+CC=x86_64-linux-gnu-gcc
```
rbs-jacob, line 2: In my opinion, this is not a good change, because the premise is faulty. We should not be building these ELF files dynamically in the first place. They should instead be Git LFS committed to the repo like all of our other test assets. 

We don't gain anything by building them dynamically except for requiring additional dependencies to run the tests.

ofrak_core/pytest_ofrak/elf/fixtures.py
```diff
@@ -6,7 +6,7 @@
 
 
 # MAKEFILE_CONTENTS = """
-# CC=gcc
+# CC=x86_64-linux-gnu-gcc
```
rbs-jacob, line 9: Same comment as above about dynamically building ELF test assets.

Also this stuff is entirely commented out anyway and probably shouldn't even be in this file?

ofrak_core/Dockerstub
```diff
@@ -68,3 +68,17 @@ RUN cd /tmp && \
...
+      apt-get -y install --no-install-recommends \
+      libc6:amd64 \
+      zlib1g:amd64 \
+      libselinux1:amd64 \
+      libacl1:amd64 \
+      libmpc3:amd64 \
+      libisl23:amd64 \
+      libstdc++6:amd64 \
+      crossbuild-essential-amd64; \
+fi;
```
rbs-jacob, line 84: If we install GCC for amd64 from a tarball instead of apt, we should be able to drop these multiarch packages, right?

disassemblers/ofrak_angr/Makefile
```diff
@@ -16,4 +16,4 @@ inspect:
 .PHONY: test
 test: inspect
 	$(PYTHON) -m pytest -n auto --cov=ofrak_angr ofrak_angr_test
-	fun-coverage --cov-fail-under=100
+	fun-coverage --cov-fail-under=100 || ! (shell uname -m | grep x86_64)
```
rbs-jacob, line 19: I see the value in having a way for ARM people to run all of the tests without them stopping on every package due to function coverage failures. I do not think adding test coverage exceptions to each package is the way to do it.

Partly I think it sets a bad precedent in the code. Partly I think it might complicate our safety cert situation. 

I'd rather have a top-level Makefile target in the repo root Makefile that does "run all tests, skip failures" and could look something like:

```makefile
# I haven't actually tested this, it's just off the dome
test-skip-failure:
    find . -name Makefile -maxdepth 1 -execdir sh -c '$(MAKE) test || true'
```

disassemblers/ofrak_angr/requirements.txt
```diff
@@ -1,2 +1,2 @@
-angr==9.2.93
+angr>=9.2.93
```
rbs-jacob, line 1: I get that `angr` changes often, but that's exactly why we pin versions. I think we should bump it to a better version, but using any version at all seems irresponsible given how often they push breaking changes.

ofrak_patch_maker/ofrak_patch_maker/toolchain/utils.py
```diff
@@ -70,6 +78,34 @@ def get_repository_config(section: str, key: Optional[str] = None):
...
+    :param section: section name in config file
+    :param key: key in `config[section]`
+
+    :raises SystemExit: If `config[section]` or `config[section][key]` not found.
+    :return Union[str, List[Tuple[str, str]]]: the result of ``config.get(section, key)`` or
+        ``config.items(section)``
+    """
+
+    exec_path = get_repository_config(section, key)
+    if shutil.which(exec_path) is not None:
```
rbs-jacob, line 95: I don't understand why we need `shutil.which` here when all of the paths in the configuration file are absolute and fully qualified?

rbs-jacob (changes requested):
More changes required before I'm comfortable merging this.

ofrak_patch_maker/Dockerstub
paulnoalhyt: https://www.kernel.org/pub/tools/crosstool/files/bin/arm64/10.3.0/arm64-gcc-10.3.0-nolibc-x86_64-linux.tar.xz might be a good candidate.

disassemblers/ofrak_angr/Makefile
whyitfor: I've removed all of these for now. This PR simply allows the ARM64 build to complete (getting all the tests to pass is out of scope, will be addressed in subsequent PR).

ofrak_core/pytest_ofrak/elf/fixtures.py
whyitfor: Removed these lines

ofrak_patch_maker/Dockerstub
whyitfor: Good pick!

rbs-jacob (changes requested):
Two very small questions/comments. Then, this should be good to go!

rbs-jacob (approved):

rbs-jacob:
@whyitfor Why did you end up breaking the toolchain config back out into two files in https://github.com/redballoonsecurity/ofrak/pull/562/commits/31ae6547ac529201f0bd9501c9de566330f5960c ?

rbs-jacob (approved):

Issue #547, opened 2024-12-27
Fix ofrak-dev, ofrak-ghidra, ofrak-tutorial entrypoints


Jepson2k:
**What is the problem? (Here is where you provide a complete Traceback.)**
The addition of ofrak licensing broke the entrypoint commands for ofrak-dev, ofrak-ghidra, and ofrak-tutorial.

**Please provide some information about your environment.**
At minimum we would like the following information on your platform and Python environment:
  - Platform: redballoonsecurity/ofrak/dev, .../ghidra, .../tutorial

**If you've discovered it, what is the root cause of the problem?**
Ofrak license needs to be run first

**How often does the issue happen?**
Everytime using those configs.

**What are the steps to reproduce the issue?**
Ideally, give us a short script that reproduces the issue.
Build those images, start those images.

**How would you implement this fix?**
Add `ofrak license --community --i-agree` command to the entrypoint commands.

**Are there any (reasonable) alternative approaches?**
No

**Are you interested in implementing it yourself?**
Yes

ANogin:
@Jepson2k , we should not be submitting either `--community` or `--i-agree` options on behalf of the user - we need the user to possess the appropriate license (e.g. either explicitly confirming their use qualifies for the Community License or purchasing an appropriate license authorizing Commercial Use), and explicitly agreeing to the terms of the license. Off the top of my head, one possibility would be to provide a web UI for this, another is to enable the user to provide the license inputs at the docker build time.

Jepson2k:
@ANogin Took a different approach:
- Upon the first start of a docker container it checks for the existence of the ofrak.license file.
  - If it finds it, it attaches it to the container and proceeds as normal.
  - If it does not find it, it first brings up the freshly build container running `ofrak license` command. Prompts user for input and if the user selects the license and agrees to the terms it copies the license file out of the container and then proceeds as normal.
 
Thoughts?

PR #563, opened 2025-01-07
Docker image improvements and Makefile targets

alchzh:
Authored by @Jepson2k

Fix small issues with Docker images and add Makefile targets for creating/starting them.

**Link to Related Issue(s)**
#547 

- Reorder packages so ofrak_patch_maker comes first for better caching.
- Add Makefile targets ofrak-*, start-ofrak-*, and super-start-ofrak-* for creating/starting Docker images
- Change entrypoints in some docker images to auto accept the license (is this what we want to do? See #547)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ super-start-ofrak-%: ## Start OFRAK image with mounted volumes, profiling tools,
 	  --cap-add SYS_PTRACE \
 	  $(call check_for_binja_license,) \
 	  $(call volume_mounts,$(PACKAGES)) --entrypoint bash \
-	  redballoonsecurity/ofrak/dev:latest \
+	  redballoonsecurity/ofrak/$*:latest \
 	  -c 'make develop \
 	  && ofrak license --community --i-agree \
 	  && (sleep infinity)'
--- a/Makefile
+++ b/Makefile
@@ -121,8 +121,8 @@ $(foreach pkg,$(1),--volume "$$(pwd)"/$(pkg):/$(shell basename $(pkg)) \$(newlin
 endef
 
 .PHONY: super-start-ofrak-%
-super-start-ofrak-%: IMAGE_NAME = redballoonsecurity/ofrak/$*:latest
-super-start-ofrak-%: ensure-ofrak-license ## Start OFRAK image with mounted volumes, profiling tools, misc dev utils, and a modified entrypoint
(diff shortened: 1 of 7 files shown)


rbs-jacob:
I do not think we should move forward with most of the changes in this pull request.

- The Makefile comments almost exclusively add noise, and are redundant – we don't need a comment for a single command that is the same length as the comment describing it
  - In general, comments should document why decisions were made, not what the code is doing
- The license terms need to be manually agreed by a user – there should not be code in the Makefile or entrypoint to automatically agree
- Many of the complex Makefile functions should not be in the Makefile at all – part of the reason we have `build_image.py` is to avoid complex, hard-to-maintain logic in Makefiles and shell scripts

The only change that *might* be alright is changing the order of `ofrak_patch_maker` in the YAML files, and I'm hesitant about that. It would be good for caching, but the order in which OFRAK packages are installed in Docker impacts whether the OFRAK version we are trying to build from source gets overridden by the `pip` version. We need to manually confirm that this is not happening in the images that get built before I am okay with moving ahead on this particular subset of the proposed changeset.

ANogin:
> The only change that _might_ be alright is changing the order of `ofrak_patch_maker` in the YAML files, and I'm hesitant about that. It would be good for caching, but the order in which OFRAK packages are installed in Docker impacts whether the OFRAK version we are trying to build from source gets overridden by the `pip` version. 

Yes. the order in yaml must agree with the dependencies. 

> We need to manually confirm that this is not happening in the images that get built before I am okay with moving ahead on this particular subset of the proposed changeset.

We need to finally land https://github.com/redballoonsecurity/ofrak/pull/218, then the check would no longer have to be manual.

Jepson2k:
I can remove the comments but otherwise I think the rest of the changes are worthwhile. There is currently a large barrier to adoption with ofrak in the sense that there is no easy way to download, build, and run it. Users shouldn't have to have a deep understanding of the ofrak build system in order to simply build and run it in a container.

I would also argue that the logic added to the makefile is neither complex nor obscure. It's a simple way to add fuzz matching to allow users to quickly build and run the container using the predefined ofrak configuration files.

Moreover the order of the yaml is not a clearly documented requirement so if this is a restriction then we should document it. 

As for the licensing it requires the user to have generated the ofrak.license file, a step that requires the user to agree to the license. If this method is not satisfactory please propose a new method that doesn't require the user to type "I agree" every time they start a container.

ANogin (changes requested):

Issue #550, opened 2024-12-29
Bump Mypy


Jepson2k:
**Which files would be affected?**
Mypy static checking has grown a lot and provides more useful analysis in later versions, but with the introduction of PEP 484 and the banning-ish of implicit optionals there are a lot of type hints that will need to be added to many files. Mypy recommends a program that auto converts the code base so it shouldn't be hard to do but it'll affect many files.

**Does the proposed maintenance include non-doc string _functional_ changes to the Python code?**
Yes

**Are you interested in implementing it yourself?**
Yes

whyitfor:
This should be tackled in a PR that only changes this and nothing else, to isolate these changes

PR #565, opened 2025-01-07
Configure mypy skip following imports for some dependencies to workaround bugs

alchzh:
Authored by @Jepson2k

Workaround for issues such as  https://github.com/python/mypy/issues/13499
in the version of mypy version we're currently using which are causing
spurious type checking failures.

These workarounds should be removed when we bump the mypy version to
one without these bugs.

Configure mypy skip following imports for some dependencies to workaround bugs.

**Link to Related Issue(s)**
#550 

Add `follow_imports = skip` to `tempfile312`, `lief` and `numpy` to workaround mypy bugs causing spurious type checking failures.


whyitfor:
Is this an issue with `mypy==0.942`? I'm trying to understand why this change is needed -- we don't seem to have failures in CI.

alchzh:
the change is needed when mypy is installed with a python version >3.8, but because of the linked bug reports spurious failures in stub files

disassemblers/ofrak_angr/mypy.ini
```diff
@@ -3,3 +3,18 @@ files = ofrak_angr
...
+follow_imports = skip
+follow_imports_for_stubs = true
+ignore_missing_imports = True
+
+[mypy-lief.*] # TODO remove the next two lines when upgrading mypy. Current version has bug.
+follow_imports = skip
+follow_imports_for_stubs = true
+ignore_missing_imports = True
+
+[mypy-numpy.*] # TODO remove the next two lines when upgrading mypy. Current version has bug.
```
whyitfor, line 17: I'm not aware of `numpy` being a dependency of OFRAK (at least not directly).

disassemblers/ofrak_angr/mypy.ini
```diff
@@ -3,3 +3,18 @@ files = ofrak_angr
...
 ignore_missing_imports = True
+
+[mypy-tempfile312.*] # TODO remove the next two lines when upgrading mypy. Current version has bug.
+follow_imports = skip
+follow_imports_for_stubs = true
+ignore_missing_imports = True
+
+[mypy-lief.*] # TODO remove the next two lines when upgrading mypy. Current version has bug.
+follow_imports = skip
+follow_imports_for_stubs = true
```
whyitfor, line 14: Do I understand correctly that `ignore_missing_imports` alone here is not sufficient?

I'm not tracking why this is needed for lief but not for any of the other libraries in this file.

disassemblers/ofrak_angr/mypy.ini
```diff
@@ -3,3 +3,18 @@ files = ofrak_angr
 
 [mypy-angr.*]
 ignore_missing_imports = True
+
+[mypy-tempfile312.*] # TODO remove the next two lines when upgrading mypy. Current version has bug.
+follow_imports = skip
+follow_imports_for_stubs = true
```
whyitfor, line 9: Have we tried just using `ignore_missing_imports` here?

disassemblers/ofrak_angr/mypy.ini
```diff
@@ -3,3 +3,18 @@ files = ofrak_angr
 
 [mypy-angr.*]
 ignore_missing_imports = True
+
+[mypy-tempfile312.*] # TODO remove the next two lines when upgrading mypy. Current version has bug.
```
whyitfor, line 7: What version of mypy? What bug?

This comment isn't really useful. I'd recommend just putting a comment in this PR to document whatever this is trying to communicate and removing this comment. We should use the issue tracker to track issues, not comments.

disassemblers/ofrak_angr/mypy.ini
Jepson2k: I believe its a dependency of a dependency of angr. I was confused as well as to why mypy was scanning it. I'll look in to it some more.

disassemblers/ofrak_angr/mypy.ini
Jepson2k: Yes ignore_missing_imports alone does not fix the issue. I believe there was a bug report in mypy. But if we switch to legacy mode for the editable installs then this may not be an issue.

alchzh:
The bug this workaround is needed for was fixed in mypy version 0.981. This change should be reverted after mypy to that version or later in the future. Tracked in #550

disassemblers/ofrak_angr/mypy.ini
alchzh: it's a dependency of pyformlang which is a dependency of angr

whyitfor (approved):
LGTM

PR #569, opened 2025-01-13
Fix bug with resource data offset in `FreeSpaceModifier` introduced in #505

alchzh:

Fix bug with resource data offset in free_space.py introduced in #505. (No changelog necessary, fixes a bug recently introduced)

**Link to Related Issue(s)**

- Fix a typo where `(await resource.get_data_range_within_parent()).start` was accidentally replaced with `(await parent.get_data_range_within_parent()).start` in `FreeSpaceModifier`
 - Change the free space modifier tests to use a child memoryregion with non-zero offset to catch this bug in the future.
 
### DIFF
--- a/ofrak_core/test_ofrak/components/test_free_space.py
+++ b/ofrak_core/test_ofrak/components/test_free_space.py
@@ -18,12 +18,12 @@ from ofrak_type.range import Range
 @pytest.fixture
 async def resource_under_test(ofrak_context: OFRAKContext) -> Resource:
     resource = await ofrak_context.create_root_resource(
-        "mock_memory_region",
-        b"\xff" * 0x110,
+        "mock_program",
+        b"\xff" * 0x100,
         (Program,),
     )
     memory_region = await resource.create_child_from_view(
-        MemoryRegion(0x10, 0x100), data_range=Range(0x10, 0x110)
+        MemoryRegion(0x100, 0xF0), data_range=Range(0x10, 0x100)
     )
     await resource.save()
     await memory_region.save()
@@ -37,7 +37,9 @@ async def dataless_resource_under_test(ofrak_context: OFRAKContext) -> Resource:
         b"",
         (Program,),
     )
-    memory_region = await resource.create_child_from_view(MemoryRegion(0x0, 0x100), data_range=None)
+    memory_region = await resource.create_child_from_view(
+        MemoryRegion(0x100, 0xF0), data_range=None
+    )
     await resource.save()
     await memory_region.save()
     return memory_region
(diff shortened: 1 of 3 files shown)


ofrak_core/test_ofrak/components/test_free_space.py
```diff
@@ -19,11 +19,11 @@
 async def resource_under_test(ofrak_context: OFRAKContext) -> Resource:
     resource = await ofrak_context.create_root_resource(
         "mock_memory_region",
-        b"\xff" * 0x100,
+        b"\xff" * 0x110,
         (Program,),
     )
     memory_region = await resource.create_child_from_view(
-        MemoryRegion(0, 0x100), data_range=Range(0, 0x100)
+        MemoryRegion(0x10, 0x100), data_range=Range(0x10, 0x110)
```
whyitfor, line 26: Does the memory region need to be 0x100 bytes long?

This is a small nit, but I think this bugfix would be easier to understand if only the start offset changed, and not the end offset with it.

ofrak_core/test_ofrak/components/test_free_space.py
```diff
@@ -19,11 +19,11 @@
 async def resource_under_test(ofrak_context: OFRAKContext) -> Resource:
     resource = await ofrak_context.create_root_resource(
         "mock_memory_region",
-        b"\xff" * 0x100,
+        b"\xff" * 0x110,
```
whyitfor, line 22: Line 21 is confusing -- should this be renamed to "mock_program"? (I realize this is my fault :) ).

ofrak_core/test_ofrak/components/test_free_space.py
```diff
@@ -63,11 +63,12 @@ async def test_partial_free_modifier(resource_under_test: Resource):
     """
     Test that the PartialFreeSpaceModifier returns expected results.
     """
-    partial_start_offset = 4
-    partial_end_offset = 10
-    parent = await resource_under_test.get_parent()
+    partial_start_offset = 0x14
+    partial_end_offset = 0x1A
```
whyitfor, line 67: I don't believe this update is needed.

This offset _should_ be the offset in the memory region, not in the parent resource, so 4 and 10 should still work here.

What this test should be doing is replacing bytes 4-10 with the stub (which is not free space) plus `\x00` * remaining length of the modification, which would be 2 bytes in this case. Do you think we can refactor it to make this intention clearer?

whyitfor (changes requested):
Thanks for catching this. Please see my suggestions -- they mainly revolve around making this test case a little easier to understand.

While I agree that this doesn't require a separate change log entry, i wonder if we should update the existing one to list both 505 and 509? @rbs-jacob, we should probably have a standard approach to things like this.

ofrak_core/test_ofrak/components/test_free_space.py
alchzh: These numbers shouldn't be offsets at all, they're addresses in the `MemoryRegion` resource. I've updated the variable names to make this clearer and changed the next line to make sense.

I've also changed the `MemoryRegion` virtual address to `0x100` so we don't run into more bugs hidden by only testing with `0`

rbs-jacob:
> While I agree that this doesn't require a separate change log entry, i wonder if we should update the existing one to list both 505 and 509? @rbs-jacob, we should probably have a standard approach to things like this.

I'm in favor of standardizing on listing all relevant PRs for a changelog entry. I'll make a PR that updates the contributor guidelines accordingly to codify this.

whyitfor (approved):
@alchzh, per https://github.com/redballoonsecurity/ofrak/pull/569#issuecomment-2590646518, can you update the CHANGELOG to list this PR in addition to 505?

Other than that this looks good to me, feel free to merge in once the CHANGELOG is updated and you are ready.

Issue #548, opened 2024-12-27
Add test commands to makefile


Jepson2k:
**What is the use case for the feature?**
Currently there is no easy way to test locally that mimics the testing setup by the github runners. Ideally there should be a `make test-all` command that runs the same tests which are run by the github runners. This would allow new developers to easily test their changes without copying the commands one by one from the test-all.yml workflow file.

**Does the feature contain any proprietary information about another company's intellectual property?**
No

**How would you implement this feature?**
Essentially add the same sequence of commands used in the test-all.yml workflow file to the Makefile under a command titled "test-all".

**Are there any (reasonable) alternative approaches?**
Yes

**Are you interested in implementing it yourself?**
Yes

whyitfor:
When you use `build-image.py` to create a Docker image, it builds a Makefile that includes a test target for every OFRAK package. Here is an example:

```
# cat /Makefile
develop:
	$(MAKE) -C ofrak_type develop
	$(MAKE) -C ofrak_io develop
	$(MAKE) -C ofrak_patch_maker develop
	$(MAKE) -C ofrak_core develop
	$(MAKE) -C ofrak_ghidra develop
	$(MAKE) -C frontend develop

.PHONY: test test_ofrak_type test_ofrak_io test_ofrak_patch_maker test_ofrak_core test_ofrak_ghidra test_frontend
test: test_ofrak_type test_ofrak_io test_ofrak_patch_maker test_ofrak_core test_ofrak_ghidra test_frontend
test_ofrak_type:
	$(MAKE) -C ofrak_type test
test_ofrak_io:
	$(MAKE) -C ofrak_io test
test_ofrak_patch_maker:
	$(MAKE) -C ofrak_patch_maker test
test_ofrak_core:
	$(MAKE) -C ofrak_core test
test_ofrak_ghidra:
	$(MAKE) -C ofrak_ghidra test
test_frontend:
	$(MAKE) -C frontend test
```

Is there a reason you cannot use this Makefile target in your testing?

rbs-jacob:
It is perhaps also worth recommending that folks who want to run the exact GitHub Actions jobs use the following tool that does just that:

https://github.com/nektos/act

Jepson2k:
> When you use `build-image.py` to create a Docker image, it builds a Makefile that includes a test target for every OFRAK package. Here is an example:
> 
> ```
> # cat /Makefile
> develop:
> 	$(MAKE) -C ofrak_type develop
[...]

PR #571, opened 2025-01-14
Fix circular dependencies by making `ofrak_angr` and `ofrak_capstone` optional for `ofrak_core` tests

### LINKED ISSUE #419: Circular dependency cases a `make develop` weirdness.

ANogin:

Optmizes the build of the `--target=development` docker containers by removing the `ofrak_core` "forward" test dependencies on `ofrak_angr` and `ofrak_capstone`

**Link to Related Issue(s)**
Fixes #419 

https://github.com/redballoonsecurity/ofrak/pull/226 introduced an unfortunate issue (which I discovered using PR https://github.com/redballoonsecurity/ofrak/pull/218 - IMHO we really out to land that one, and use in CI) - when the docker build runs make develop, it would first go to ofrak_core, where the "test": "ofrak_angr"] dependency would cause it to download ofrak_angr from PyPI, then it would go into `ofrak_angr `and then the make develop there would overwrite the downloaded one with the local one (and similar with `ofrak_capstone`). With https://github.com/redballoonsecurity/ofrak/pull/417, this means that not only the `ofrak_angr` itself is downloaded from PyPI, but also the angr itself and all its dependencies installed in `base.Dockerfile` are updated to the versions required by the `ofrak_angr` from PyPI. Then `make develop` would put them all back.

A potentially even bigger issue is that the `ofrak-ghidra.yaml` container would use the `ofrak_angr` and `ofrak_capstone` from pypi, not the latest, so the testing in CI would not necessaily be accurately representative of what is actually going on!

This appempts to fix the issues by:

- Removing the extra dependencies
- Skipping the test that depend on  `ofrak_angr` and `ofrak_capstone`  when those are not installed
- Using a new `ofrak-angr-ghidra.yml` container for all tests

@whyitfor


whyitfor (changes requested):
@ANogin, please see comments -- once this is done, we can merge this in.

The comments suggest a short-term work-around to this issue.

Longer term, I'd like to have the following true:

1. `ofrak-dev.yml` allows a user to build, test, and use the core ofrak components only. We need to remove the `ofrak_angr` and `ofrak_capstone` test dependencies here.
2. `ofrak-ghidra.yml` allows a user to build, test, and use ofrak with Ghidra.
3. `ofrak-angr.yml` allows a users to build, test, and use ofrak with angr.

ANogin:
> * `ofrak-dev.yml` allows a user to build, test, and use the core ofrak components only. We need to remove the `ofrak_angr` and `ofrak_capstone` test dependencies here.

Isn't that what `ofrak-core-dev.yml` is for?

ofrak_core/setup.py
```diff
@@ -83,17 +83,7 @@ def read_requirements(requirements_path):
     + read_requirements("requirements.txt"),
     extras_require={
         "docs": read_requirements("requirements-docs.txt"),
-        "test": [
-            "importlib-resources",  # Needed because of https://github.com/redballoonsecurity/ofrak/issues/398
-            "ofrak_angr~=1.0",
```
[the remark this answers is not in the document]
ANogin: @whyitfor This does not make sense to me, not sure what you are asking for
- What do you mean by "lead it commented out (see ofrak_capstone)" - there is nothing commented out in ofrak_capstone?
- "bump `ofrak_angr to "ofrak_angr>=1.1.0rc0, and list that is the requirement here" - but the goal here is to remove this requirement althogether?

ofrak_core/setup.py
```diff
@@ -83,17 +83,7 @@ def read_requirements(requirements_path):
...
-            # Running the tests requires this version of capstone.
-            #  If building a Docker image, it should already be installed on the system.
-            #  If you are running the tests in another environment, you will need to make sure
-            #  that this version is installed from GitHub, since it depends on ofrak==3.3.0rc0
-            #  and neither of these are yet released on PyPI.
-            # "ofrak_capstone>=1.1.0rc0",
-        ]
-        + read_requirements("requirements-test.txt"),
+        "test": read_requirements("requirements-test.txt"),
```
whyitfor, line 86:
```suggestion
        "test": [
            "importlib-resources",  # Needed because of https://github.com/redballoonsecurity/ofrak/issues/398
            # Running the tests requires unreleased versions of ofrak_angr and ofrak_capstone.
            #  If building a Docker image, it should already be installed on the system.
            #  If you are running the tests in another environment, you will need to make sure
            #  that this version is installed from GitHub, since it depends on ofrak==3.3.0rc0
            #  and neither of these are yet released on PyPI.
            # "ofrak_angr>=1.1.0rc0",
            # "ofrak_capstone>=1.1.0rc0",
        ]
        + read_requirements("requirements-test.txt"),
```

ofrak_core/setup.py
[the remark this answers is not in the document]
whyitfor: See https://github.com/redballoonsecurity/ofrak/pull/571/files#r1917187442. The point is to document the version that the test does currently rely on.

Once that is accepted, if you update `ofrak_angr/setup.py` and `ofrak_angr/CHANGELOG.md` to be 1.1.0rc0.

ofrak_core/Makefile
```diff
@@ -18,7 +18,7 @@ test: inspect
 	$(PYTHON) -m pytest -n auto test_ofrak --cov=ofrak --cov-report=term-missing
 	(sleep 2; echo 1; sleep 2; echo i agree) \
 		| python3 -m coverage run --append --source ofrak -m ofrak license --force
-	fun-coverage --cov-fail-under=100
+	if $(PYTHON) -m pip show -qqq ofrak_angr; then fun-coverage --cov-fail-under=100; fi
```
whyitfor, line 21:
```suggestion
	fun-coverage --cov-fail-under=100
```

The point of fun-coverage here is to assert that all functions are tested. This test should fail if we are skipping tests -- that is a feature, not a bug.

ofrak_core/setup.py
ANogin: Will apply, but will also comment out importlib-resources (only needed when ofrak_angr is there)

ofrak_core/Makefile
ANogin: @whyitfor , how would one then do a complete regression testing? My go-to approach when I am messing with docker configurations and want to be extra sure I did not break anything is to build both versions (install and develop) of every .yml configuration, and run tests in the develop version of each of them (https://github.com/redballoonsecurity/ofrak/issues/548#issuecomment-2591575019). Does not quite work right now for `ofrak-core-dev.yml` because of python11 incomatibility in some tests, but once that is fixed, wouldn't it be good to retain the ability to test all configurations and have nothing fail?

ANogin: P.S. combined with #572 I just created, this ensures that `make test` in `ofrak-core-dev` develop container succeeds.

whyitfor:
> > * `ofrak-dev.yml` allows a user to build, test, and use the core ofrak components only. We need to remove the `ofrak_angr` and `ofrak_capstone` test dependencies here.
> 
> Isn't that what `ofrak-core-dev.yml` is for?

Sure, `ofrak-core-dev.yml` allows a user to build, test, and use the core components only.

ofrak_core/Makefile
whyitfor: `make -C ofrak_core test` should pass only if the following are true:

- All of the tests that are run pass
- The tests that are run provide 100% function coverage for the package under test

I guess the implication of this is that it will not pass in an `ofrak-core-dev.yml` container. This is regrettable, but I'm fine with that for now since, as I've mentioned, we will eventually refactor the tests to not rely on disassembler back ends.

ANogin:
@whyitfor does this require a Changelog entry?

whyitfor:
> @whyitfor does this require a Changelog entry?

I think so. I'm generally thinking the following:

1. A `Fixed` entry here saying that `ofrak` drops circular test dependencies.
2. We should bump (in changelog and in setup.py) ofrak's version to `3.3.0rc1`. In the changelog, this will just change the `Unreleased` line. The idea behind this is then we can use these `rc` versions internally to specify "this is the unreleased version of ofrak I'm using" (this can be helpful for projects that rely on ofrak "master" -- they can specify through versioning that they are on an older version of master. 

@rbs-jacob, does this make sense to you?

rbs-jacob:
> @rbs-jacob, does this make sense to you?

Yup, sounds good to me. I think bumping versions before release is something we probably should have been doing the whole time, but I don't think anyone thought of it until recently.

whyitfor (approved):
LGTM

PR #590, opened 2025-05-28
Fix compiler flag for PIE relocatable patches

rbs-afflitto:

Fix compiler flag for PIE relocatable patches - the GNU toolchain used the flag `-pie`, which has no effect. The correct flag is `-fPIE`.

**Link to Related Issue(s)**

N/A

- Change GNU toolchain to use `-fPIE` instead of `-pie` when building relocatable patches
- Add relocatable test for x86 and AARCH64
  - Note: I'm having trouble getting this test to work for LLVM toolchains, so I haven't added this test to the ARM target yet. The userspace LLVM patch maker does not provide a custom linker script so we cannot provide symbols via the `base_symbols` patch maker feature.

@rbs-jacob

### DIFF
--- a/ofrak_patch_maker/ofrak_patch_maker/toolchain/gnu.py
+++ b/ofrak_patch_maker/ofrak_patch_maker/toolchain/gnu.py
@@ -372,7 +372,7 @@ class GNU_10_Toolchain(Abstract_GNU_Toolchain):
 
         if self._config.relocatable:
             self._compiler_flags.append("-fPIE")
-            self._linker_flags.append("--pic-executable")
+            self._linker_flags.append("-pie")
         else:
             self._compiler_flags.append("-fno-plt")
             self._compiler_flags.append("-fno-pic")


ofrak_patch_maker/ofrak_patch_maker/toolchain/gnu.py
```diff
@@ -371,7 +371,7 @@ def __init__(
         )
 
         if self._config.relocatable:
-            self._compiler_flags.append("-pie")
+            self._compiler_flags.append("-fPIE")
             self._linker_flags.append("--pic-executable")
```
rbs-jacob, line 375: Hard to tell from the docs, but it looks like the compiler wants `-fPIE` if the linker will be using `-pie`. Documentation for `ld` indicates that `--pic-executable` is equivalent to `-pie`, but documentation for GCC omits any mention of `--pic-executable`... I wonder if we should be setting `pie` instead of `--pic-executable`?

- <https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#index-fpie>
- <https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-pie>
- <https://man7.org/linux/man-pages/man1/ld.1.html>

ofrak_patch_maker/ofrak_patch_maker_test/test_arm_toolchain.py
```diff
@@ -116,6 +120,12 @@ def test_hello_world(toolchain_under_test: ToolchainUnderTest):
     run_hello_world_test(toolchain_under_test)
 
 
+def test_relocatable(toolchain_under_test: ToolchainUnderTest):
+    # LLVM userspace test can't supply external symbols and LLVM non-userspace test can't link .got for some reason
+    if toolchain_under_test.toolchain != LLVM_12_0_1_Toolchain:
+        run_relocatable_test(toolchain_under_test)
```
rbs-jacob, line 126: Instead of having this test pass, maybe we should [use `pytest.skip`](https://stackoverflow.com/questions/38442897/how-do-i-disable-a-test-using-pytest)? 

We seem to use it in one or two places in the code, though our usage differs from the suggestions in the linked StackOverflow post.

https://github.com/redballoonsecurity/ofrak/blob/5ca2c3f6bc8b80e647bfa4d9e81c175d6d0b14ee/ofrak_core/pytest_ofrak/patterns/register_usage_analyzer.py#L444-L454

ofrak_patch_maker/ofrak_patch_maker_test/toolchain_c.py
```diff
@@ -169,3 +169,83 @@ def run_hello_world_test(toolchain_under_test: ToolchainUnderTest):
...
     assert get_file_format(exec_path) == tc_config.file_format
+
+
+def run_relocatable_test(toolchain_under_test: ToolchainUnderTest):
+    """
+    Use patchmaker to link on a relocatable binary
+    """
+    source_dir = os.path.join(os.path.dirname(__file__), "example_5")
+    source_path = os.path.join(source_dir, "patch.c")
+    build_dir = tempfile.mkdtemp()
```
rbs-jacob, line 180: This should probably use the [pytest `tmpdir` fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html).

rbs-jacob (changes requested):
Couple of small comments. If those are addressed and the tests pass, this otherwise looks like a good bug fix. Thanks!

ofrak_patch_maker/ofrak_patch_maker/toolchain/gnu.py
rbs-afflitto: As far as I can tell, `--pic-executable` is just the long-form alias for `-pie`. If `-pie` seems more readable I'll change it otherwise I'll leave it as-is.

rbs-afflitto: Changed this to `-pie` in case we decide to use gcc for linking in the future (gcc only supports `pie` not `pic-executable`)

rbs-jacob (approved):
Looks good to me!

PR #591, opened 2025-06-04
Added some tests, added some error checking in range.py

paulnoalhyt:

Added some tests in ofrak_type, added error checking in `range.py`

**Link to Related Issue(s)**

- Added tests to `ofrak_type_test`
- Added error checking in `range.py`

No

### DIFF
--- a/ofrak_type/CHANGELOG.md
+++ b/ofrak_type/CHANGELOG.md
@@ -5,11 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 
 ## [Unreleased](https://github.com/redballoonsecurity/ofrak/tree/master)
 
+### Added
+- Added tests to the `ofrak_type_test` package ([#591](https://github.com/redballoonsecurity/ofrak/pull/591))
+
 ### Changed
 - Minor update to OFRAK Community License, add OFRAK Pro License ([#478](https://github.com/redballoonsecurity/ofrak/pull/478))
 
 
 ### Fixed
+- `intersect` method in `Range` now raises a `ValueError` when ranges touch at a single point but do not intersect ([#591](https://github.com/redballoonsecurity/ofrak/pull/591))
 - Fix ValueError after analyzing ELF with invalid MemoryPermissions ([#581](https://github.com/redballoonsecurity/ofrak/pull/581))
 
 ## [2.2.0](https://github.com/redballoonsecurity/ofrak/compare/ofrak-type-v2.1.0...ofrak-type-v2.2.0)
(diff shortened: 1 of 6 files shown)


ofrak_type/ofrak_type/range.py
```diff
@@ -68,7 +68,7 @@ def intersect(self, range: "Range") -> "Range":
         """
         start = max(range.start, self.start)
         end = min(range.end, self.end)
-        if start > end:
+        if start >= end:
```
rbs-jacob: I'm not convinced that this was a real bug before. 

We defined ranges to always assume `Range.start <= Range.end`, so `Range(10, 3)` is meaningless, whereas `Range(10, 10)` is just the empty set. (See the `Range.__post_init__` method, where we specifically say that `start` may equal `end`.) The empty range is still valid in general – computing its length makes sense, as does taking its intersection, union, and all of the other range operations.

This change is equivalent to saying that "returning the empty range here is an error." I am not sure that makes sense for all cases we use ranges for. I have to look closer at the failing tests, but they may indicate that there are times where we *do* let this method return an empty range, and it works fine. 

On the other hand, there is an argument to be made that empty ranges could be unintuitive for a user. Personally, I think errors instead of empty ranges are less intuitive, but I could be convinced otherwise.

rbs-jacob (changes requested):

ofrak_type/ofrak_type/range.py
paulnoalhyt: I guess I was confused by the empty range. But if that's valid I guess intersect returning an empty range is also valid if there's not intersection between 2 ranges. I reverted that change and modified the tests accordingly.

ofrak_type/CHANGELOG.md
```diff
@@ -5,11 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 
+### Added
+- Added tests to the `ofrak_type_test` package ([#591](https://github.com/redballoonsecurity/ofrak/pull/591))
+
 ### Changed
 - Minor update to OFRAK Community License, add OFRAK Pro License ([#478](https://github.com/redballoonsecurity/ofrak/pull/478))
 
 
 ### Fixed
+- `intersect` method in `Range` now raises a `ValueError` when ranges touch at a single point but do not intersect ([#591](https://github.com/redballoonsecurity/ofrak/pull/591))
```
paulnoalhyt:
```suggestion
```

rbs-jacob (approved):
As far as I can tell, these additional tests look good

PR #556, opened 2025-01-03
Feature/cached disasm

dannyp303:
- [x] I have reviewed the [OFRAK contributor guide](https://ofrak.com/docs/contributor-guide/getting-started.html) and attest that this pull request is in accordance with it.

**One sentence summary of this PR (This should go in the CHANGELOG!)**
Creates two new disassemblers for OFRAK, ofrak_pyghidra using the [PyGhidra feature](https://github.com/NationalSecurityAgency/ghidra/tree/master/Ghidra/Features/PyGhidra) in the ghidra repository, and ofrak_cached for cacheing binary analysis. 

**Link to Related Issue(s)**
None

**Please describe the changes in your request.**
`ofrak_pyghidra.components`: One single unpacker that unpacks the entire binary, down to instruction level, with an option switch to not unpack basic blocks into instructions.

`ofrak_pyghidra.standalone`(name pending): A utility completely seperate from OFRAK that uses pyghidra to generate the cache file used in `ofrak_cached`

`ofrak_cached.components`: OFRAK components that ingest a cached analysis file, associates it with a resource id, and uses it to unpack the binary as requested. 

**Anyone you think should look at this, specifically?**
@rbs-jacob

PR #604, opened 2025-07-25
Handle escape character in json

paulnoalhyt:

Handle escape (`\`) character in the JSON going from Java (Ghidra) to Python.

**Link to Related Issue(s)**

None

I noticed that when the decompilation output from Ghidra contained something like `printf("This is bob\'s key.")`, then the Python side would raise an exception: `json.decoder.JSONDecodeError: Invalid \escape: line ... column ... (char ...)` because of the `\` character. I added a case in the list of special chars we replace on both ends.

No


rbs-jacob:
Hey Paul! Thanks for looking at this. 

We're almost ready to merge a different (huge) PR that would migrate us to using PyGhidra instead of the current mess: #556 

Even though your bug fix here is good, hopefully it's soon obsoleted by that other PR. I'll leave this open until that PR lands, and then probably close this one, unless you have any objections.

whyitfor:
Since `ofrak_ghidra` still exists, does it make sense to just land this change? @paulnoalhyt, the changelog and setup.py will need to be updated to be 0.2.0rc2 if we go that route....

paulnoalhyt:
Updated that version number, if we decide to go with this PR.
Regarding keeping the old java version: is there a reason why we didn't just swap the java version with the python version, keeping the same package name and interfaces?

rbs-jacob (approved):

PR #607, opened 2025-08-26
Fix ofrak_pyghidra base addresses

kmjones42:
- [X] I have made or updated a changelog entry for the changes in this pull request.

Ensure we are only unpacking CodeRegions and that we are using the correct base address.
**Link to Related Issue(s)**

This fixes two bugs. One where not every child of a PyGhidraProject will be a CodeRegion, so we filter to only unpack code regions. The other is ofrak_pyghidra stores the decompilation using the virtual address so we need to add the base_address.
@rbs-jacob


disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -169,12 +172,22 @@ async def analyze(self, resource: Resource, config=None):
...
                 except NotFoundError:
                     program_attrs = await program_r.analyze(ProgramAttributes)
+
+                base_address = None
+
+                if not any(program_r.has_tag(tag) for tag in _GHIDRA_AUTO_LOADABLE_FORMATS):
+                    code_regions = await program_r.get_children_as_view(
+                        CodeRegion, r_filter=ResourceFilter.with_tags(CodeRegion)
+                    )
+                    base_address = min(r.virtual_address for r in code_regions)
```
rbs-jacob: I don't think it's necessarily true that the minimum load address of any code region is the base address... Maybe we should be taking the min of the memory regions instead? Even that is error-prone, but probably slightly less so.

disassemblers/ofrak_pyghidra/setup.py
```diff
@@ -42,7 +42,8 @@ def read_requirements(requirements_path):
         "Programming Language :: Python :: 3",
         "Operating System :: OS Independent",
     ],
-    install_requires=read_requirements("requirements.txt") + ["ofrak", "ofrak_cached_disassembly"],
+    install_requires=read_requirements("requirements.txt")
+    + ["ofrak", "ofrak_cached_disassembly", "pyghidra~=2.1.0"],
```
rbs-jacob: If there is no specific reason this is here, we should move `pyghidra` requirement to the requirements file that we're already reading for dependencies.

disassemblers/ofrak_pyghidra/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -49,12 +49,20 @@ def unpack(program_file, decompiled, language=None, base_address=None):
             main_dictionary[seg_key] = code_region
             func_cbs = _unpack_code_region(code_region, flat_api)
             code_region["children"] = []
+
+            from ghidra.app.decompiler import DecompInterface
```
rbs-jacob: I really don't like imports inside of functions without a good reason, let alone repeatedly importing this inside a loop body! 

Not sure how I missed this on the initial review of this code, but we should not repeat this mistake without a really good reason.

rbs-jacob (changes requested):

disassemblers/ofrak_pyghidra/ofrak_pyghidra/standalone/pyghidra_analysis.py
kmjones42: I believe the ghidra imports have to come after pyghidra starts the JVM, so after the call to `pyghidra.open_program` in this case.

disassemblers/ofrak_pyghidra/requirements.txt
```diff
@@ -1 +1,4 @@
-importlib-resources  # A workaround for https://github.com/redballoonsecurity/ofrak/issues/398
+ofrak
+ofrak_cached_disassembly
```
rbs-jacob: Should these have their versions pinned?

disassemblers/ofrak_pyghidra/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -18,6 +18,12 @@ def _parse_offset(java_object):
 
 def unpack(program_file, decompiled, language=None, base_address=None):
     with pyghidra.open_program(program_file, language=language) as flat_api:
+        from ghidra.app.decompiler import DecompInterface
+        from ghidra.util.task import TaskMonitor
+        from ghidra.program.model.block import BasicBlockModel
+        from ghidra.program.model.symbol import RefType
+        from java.math import BigInteger  #  Java packages must be imported after pyghidra.start()
+
```
rbs-jacob:
```suggestion
        # Java packages must be imported after pyghidra.start or pyghidra.open_program
        from ghidra.app.decompiler import DecompInterface
        from ghidra.util.task import TaskMonitor
        from ghidra.program.model.block import BasicBlockModel
        from ghidra.program.model.symbol import RefType
        from java.math import BigInteger
```

disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -108,8 +109,11 @@ async def analyze(self, resource: Resource, config: PyGhidraAutoAnalyzerConfig =
                     return PyGhidraProject()
 
             program_attrs = resource.get_attributes(ProgramAttributes)
-            code_regions = await resource.get_children_as_view(CodeRegion)
-            base_address = min(code_region.virtual_address for code_region in code_regions)
+            regions = await resource.get_children_as_view(
+                MemoryRegion, r_filter=ResourceFilter.with_tags(MemoryRegion)
+            )
+            base_address = min(code_region.virtual_address for code_region in regions)
```
rbs-jacob, line 115:
```suggestion
            # Guess that the base address is the min start address of any memory region
            regions = await resource.get_children_as_view(
                MemoryRegion, r_filter=ResourceFilter.with_tags(MemoryRegion)
            )
            base_address = min(code_region.virtual_address for code_region in regions)
```

I think a clarifying comment here is important

rbs-jacob (changes requested):

disassemblers/ofrak_pyghidra/setup.py
kmjones42: Due to the docker build script, ofrak and ofrak_cached_disassembly cannot be included in the requirements.txt and need to reside in the setup.py. The docker build image installs requirements.txt files before any of the packages.

disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -169,12 +173,22 @@ async def analyze(self, resource: Resource, config=None):
                     program_attrs = program_r.get_attributes(ProgramAttributes)
                 except NotFoundError:
                     program_attrs = await program_r.analyze(ProgramAttributes)
+
+                base_address = None
+
+                if not any(program_r.has_tag(tag) for tag in _GHIDRA_AUTO_LOADABLE_FORMATS):
```
rbs-jacob:
```suggestion
                base_address = None
                if not any(program_r.has_tag(tag) for tag in _GHIDRA_AUTO_LOADABLE_FORMATS):
```

PR #611, opened 2025-09-15
Draft: Clean up OFRAK for python distribution and testing

whyitfor:
- [ ] I have reviewed the [OFRAK contributor guide](https://ofrak.com/docs/contributor-guide/getting-started.html) and attest that this pull request is in accordance with it.
- [ ] I have made or updated a changelog entry for the changes in this pull request.

**One sentence summary of this PR (This should go in the CHANGELOG!)**

**Link to Related Issue(s)**

**Please describe the changes in your request.**

**Anyone you think should look at this, specifically?**

PR #618, opened 2025-09-16
Add the packaging workflow and push to test pypi workflow

whyitfor:

**Link to Related Issue(s)**

### DIFF
--- a/.github/workflows/build-and-test-distributions.yml
+++ b/.github/workflows/build-and-test-distributions.yml
@@ -1,0 +1,897 @@
+name: Build and test ofrak packages
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash -euxo pipefail {0}  # NOTE: On Windows, this uses Git Bash which provides Unix compatibility (for example for ofrak-io's StreamCapture
+
+jobs:
+  build-ofrak-type-sdist:
+    name: build ofrak-type sdist
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install build tools
+        run: |
+          python -m pip install -r pip-requirements.txt
+          pip install build
+          pip freeze
+
+      - name: Build ofrak-type source distribution
+        run: |
+          cd ofrak_type
+          python -m build --sdist --outdir ../dist-ofrak-type
+
+      - name: List built packages
+        run: |
+          echo "Built OFRAK Type packages:"
+          ls -la dist-ofrak-type/
+
+      - name: Upload source distribution
+        uses: actions/upload-artifact@v4
+        with:
+          name: ofrak-type-sdist
+          path: dist-ofrak-type/
+          retention-days: 30
+
+  test-ofrak-type:
+    name: test ofrak-type on ${{ matrix.os }} with Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    needs: build-ofrak-type-sdist
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download source distribution
+        uses: actions/download-artifact@v4
+        with:
+          name: ofrak-type-sdist
+          path: dist-ofrak-type/
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+
+      - name: Create virtual environment
+        run: |
+          python -m venv venv
+        
+      - name: Activate virtual environment (Linux/macOS)
+        if: runner.os != 'Windows'
+        run: |
+          source venv/bin/activate
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
+          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
+          
+      - name: Activate virtual environment (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          venv\Scripts\activate
+          echo "VIRTUAL_ENV=$env:VIRTUAL_ENV" >> $env:GITHUB_ENV
+          echo "$env:VIRTUAL_ENV\Scripts" >> $env:GITHUB_PATH
+        shell: pwsh
+
+      - name: Install pinned pip version
+        run: |
+          python -m pip install -r pip-requirements.txt
+
+      - name: Install development requirements
+        run: |
+          pip install -r dev-requirements.txt
+
+      - name: Install OFRAK Type from sdist
+        run: |
+          pip install dist-ofrak-type/*.tar.gz
+
+      - name: Run tests (Linux/macOS)
+        if: runner.os != 'Windows'
+        run: |
+          cd ofrak_type
+          make test
+
+      - name: Run tests (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          cd ofrak_type
+          make test PYTHON=python
+
+  build-ofrak-io-sdist:
+    name: build ofrak-io sdist
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install build tools
+        run: |
+          python -m pip install -r pip-requirements.txt
+          pip install build
+
+      - name: Build OFRAK IO source distribution
+        run: |
+          cd ofrak_io
+          python -m build --sdist --outdir ../dist-io
+
+      - name: List built packages
+        run: |
+          echo "Built OFRAK IO packages:"
+          ls -la dist-io/
+
+      - name: Upload OFRAK IO source distribution
+        uses: actions/upload-artifact@v4
+        with:
+          name: ofrak-io-sdist
+          path: dist-io/
+          retention-days: 30
+
+  test-ofrak-io:
+    name: test ofrak-io on ${{ matrix.os }} with Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    needs: [build-ofrak-type-sdist, build-ofrak-io-sdist]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download ofrak_type source distribution
+        uses: actions/download-artifact@v4
+        with:
+          name: ofrak-type-sdist
+          path: dist-ofrak-type/
+
+      - name: Download ofrak_io source distribution
+        uses: actions/download-artifact@v4
+        with:
+          name: ofrak-io-sdist
+          path: dist-io/
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+
+      - name: Create virtual environment
+        run: |
+          python -m venv venv
+
+      - name: Activate virtual environment (Linux/macOS)
+        if: runner.os != 'Windows'
+        run: |
+          source venv/bin/activate
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
+          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
+
+      - name: Activate virtual environment (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          venv\Scripts\activate
+          echo "VIRTUAL_ENV=$env:VIRTUAL_ENV" >> $env:GITHUB_ENV
+          echo "$env:VIRTUAL_ENV\Scripts" >> $env:GITHUB_PATH
+        shell: pwsh
+
+      - name: Install pinned pip version
+        run: |
+          python -m pip install -r pip-requirements.txt
+
+      - name: Install development requirements
+        run: |
+          pip install -r dev-requirements.txt
+
+      - name: Install ofrak_type from sdist
+        run: |
+          pip install dist-ofrak-type/*.tar.gz
+
+      - name: Install OFRAK IO from sdist
+        run: |
+          pip install dist-io/*.tar.gz
+
+      - name: Run OFRAK IO tests (Linux/macOS)
+        if: runner.os != 'Windows'
+        run: |
+          cd ofrak_io
+          make test
+
+      - name: Run ofrak-io tests (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          cd ofrak_io
+          make test PYTHON=python
+
+  build-ofrak-patch-maker-sdist:
+    name: build ofrak-patch-maker sdist
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install build tools
+        run: |
+          python -m pip install -r pip-requirements.txt
+          pip install build
+
+      - name: Build ofrak-patch-maker source distribution
+        run: |
+          cd ofrak_patch_maker
+          python -m build --sdist --outdir ../dist-patch-maker
+
+      - name: List built packages
+        run: |
+          echo "Built ofrak-patch-maker packages:"
+          ls -la dist-patch-maker/
+
+      - name: Upload ofrak-patch-maker source distribution
+        uses: actions/upload-artifact@v4
+        with:
+          name: ofrak-patch-maker-sdist
+          path: dist-patch-maker/
+          retention-days: 30
+
+  build-pytest-ofrak-sdist:
+    name: build pytest-ofrak sdist
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          lfs: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install build tools
+        run: |
+          python -m pip install -r pip-requirements.txt
+          pip install build
+
+      - name: Build pytest-ofrak source distribution
+        run: |
+          cd pytest_ofrak
+          python -m build --sdist --outdir ../dist-pytest-ofrak
+
+      - name: List built packages
+        run: |
+          echo "Built pytest-ofrak packages:"
+          ls -la dist-pytest-ofrak/
+
+      - name: Upload pytest-ofrak source distribution
+        uses: actions/upload-artifact@v4
+        with:
+          name: pytest-ofrak-sdist
+          path: dist-pytest-ofrak/
+          retention-days: 30
+
+  build-ofrak-wheels:
+    name: build ofrak wheels on ${{ matrix.os }} for Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          lfs: true
+
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
+
+      - name: Build frontend
+        run: |
+          if [ -d "frontend" ]; then
+            cd frontend
+            npm install
+            npm run build
+            cd ../ofrak_core
+            cp -r ../frontend/dist src/ofrak/gui/public
+          else
+            echo "No frontend directory found, creating empty public directory"
+            mkdir -p ofrak_core/src/ofrak/gui/public
+          fi
+
+      - name: Install build tools
+        run: |
+          python -m pip install -r pip-requirements.txt
+          pip install cibuildwheel
+
+      - name: Set Python version for cibuildwheel
+        id: py-version
+        run: |
+          PY_VERSION="${{ matrix.python-version }}"
+          PY_VERSION_NO_DOT=${PY_VERSION//./}
+          echo "version=${PY_VERSION_NO_DOT}" >> $GITHUB_OUTPUT
+
+      - name: build ofrak wheels with cibuildwheel
+        env:
+          CIBW_BUILD: "cp${{ steps.py-version.outputs.version }}-*"
+          CIBW_SKIP: "*-win32 *-manylinux_i686"
+          CIBW_ARCHS_WINDOWS: "AMD64"  # Simplified for now
+          CIBW_ARCHS_MACOS: "x86_64 arm64"
+          CIBW_ARCHS_LINUX: "x86_64"  # Simplified for now
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_BEFORE_BUILD: "pip install setuptools wheel"
+        run: |
+          cd ofrak_core
+          cibuildwheel --output-dir ../wheelhouse
+
+      - name: List built wheels
+        run: |
+          echo "Built OFRAK wheels:"
+          ls -la wheelhouse/ || echo "No wheelhouse found"
+
+      - name: Upload wheels
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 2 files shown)


ofrak_core/tests/components/assets/binwalk_assets/README.md
```diff
@@ -1,5 +1,11 @@
...
+version https://git-lfs.github.com/spec/v1
+oid sha256:c5aa78d2c446cdd44caff1d2b43196e556336f9d486e665aeafaadda7c942f45
+size 396
+=======
 # Binwalk assets
 
 These files are taken from the [binwalk test suite](https://github.com/ReFirmLabs/binwalk/tree/master/testing/tests/input-vectors) (MIT-licensed), specifically on commit [fa0c0bd59b](https://github.com/ReFirmLabs/binwalk/commit/fa0c0bd59b8588814756942fe4cb5452e76c1dcd).
 
 Not all binwalk tests need to be duplicated here, only a few of them (the fastest-running) were selected.
+>>>>>>> master
```
rbs-jacob: I don't think this diff is supposed to be left in the file.

Also is this README supposed to be LFS-committed? I wouldn't think so.

.github/workflows/publish-to-test-pypi.yml
```diff
@@ -46,8 +46,13 @@ jobs:
   publish-to-test-pypi:
     name: Publish distributions to Test PyPI
     runs-on: ubuntu-latest
-    
+
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
rbs-jacob:
```suggestion
      - name: Checkout code
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
          lfs: false
```

If we don't need LFS assets, we should specify it. Also, if we don't actually need the code and are just pulling archived wheels, then we shouldn't even run the checkout step.

.github/workflows/publish-packages.yml
```diff
@@ -11,32 +11,32 @@ on:
...
         required: false
         type: boolean
-        default: true
+        default: false
       publish_ofrak_capstone:
         description: 'Publish ofrak-capstone'
         required: false
         type: boolean
-        default: true
+        default: false
```
rbs-jacob, line 47: Why change the defaults from true to false?

.github/workflows/publish-packages.yml
rbs-jacob, line 1: I have other comments for the parts of this file you pushed without review, but it will not let me make the suggestions because the lines were not changed in this PR.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash -euxo pipefail {0}  # NOTE: On Windows, this uses Git Bash which provides Unix compatibility (for example for ofrak-io's StreamCapture
```
rbs-jacob:
```suggestion
    # On Windows, this uses Git Bash which provides Unix compatibility for ofrak-io's StreamCapture
    shell: bash -euxo pipefail {0}
```

Can we get rid of StreamCapture so that we don't need the dependency on UNIX-like systems and shells?

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+
+jobs:
+  build-ofrak-type-sdist:
+    name: build ofrak-type sdist
+    runs-on: ubuntu-latest
```
rbs-jacob, line 24: Should we be pinning the Ubuntu version? GitHub periodically changes the images (and what is preinstalled on them).

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
rbs-jacob: As I commented on the other YAML file, we should think about whether this should pull with or without LFS. If without, we should specify it. If with, we should add the caching logic from the other workflows.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
rbs-jacob: We can add lines here to cache pip dependencies as part of the `actions/setup-python` step. See the other workflows in the repo.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
+          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
+          
+      - name: Activate virtual environment (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          venv\Scripts\activate
+          echo "VIRTUAL_ENV=$env:VIRTUAL_ENV" >> $env:GITHUB_ENV
+          echo "$env:VIRTUAL_ENV\Scripts" >> $env:GITHUB_PATH
+        shell: pwsh
```
rbs-jacob: I already commented this on the other PR, but I am skeptical that the virtual environment adds anything here. And even if it does, why are we using PowerShell for the Windows one when we implicitly use Git Bash everywhere else?

.github/workflows/build-and-test-distributions.yml
rbs-jacob: This comment applies to all of the jobs in the workflow.

.github/workflows/build-and-test-distributions.yml
rbs-jacob: This comment applies to all of the jobs in the workflow.

.github/workflows/build-and-test-distributions.yml
rbs-jacob: This applies to all of the other jobs in this workflow.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+        run: |
+          echo "Built ofrak-patch-maker packages:"
+          ls -la dist-patch-maker/
+
+      - name: Upload ofrak-patch-maker source distribution
+        uses: actions/upload-artifact@v4
+        with:
+          name: ofrak-patch-maker-sdist
+          path: dist-patch-maker/
+          retention-days: 30
```
rbs-jacob: There must be a way to combine all of these into one parameterized job instead of a ton of separate ones that all have to be updated in tandem. Perhaps using the structured GitHub Actions variables (via JSON)

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+        with:
+          name: pytest-ofrak-sdist
+          path: dist-pytest-ofrak/
+          retention-days: 30
+
+  build-ofrak-wheels:
+    name: build ofrak wheels on ${{ matrix.os }} for Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
rbs-jacob: Why this strategy?

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          lfs: true
```
rbs-jacob: Add LFS caching here.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+          if [ -d "frontend" ]; then
+            cd frontend
+            npm install
+            npm run build
+            cd ../ofrak_core
+            cp -r ../frontend/dist src/ofrak/gui/public
+          else
+            echo "No frontend directory found, creating empty public directory"
+            mkdir -p ofrak_core/src/ofrak/gui/public
+          fi
```
rbs-jacob: I already made these comments on the other PR. There should always be a frontend directory, so we don't need this error handling here. We never want to just create an empty frontend directory.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,897 @@
...
+#        if: runner.os != 'Windows'
+#        run: |
+#          cd disassemblers/ofrak_capstone
+#          ofrak license --community --i-agree && make test
+#
+#      - name: Run ofrak-capstone tests (Windows)
+#        if: runner.os == 'Windows'
+#        run: |
+#          cd disassemblers/ofrak_capstone
+#          ofrak license --community --i-agree && make test PYTHON=python
```
rbs-jacob: Do we want to keep this in here?

.github/workflows/build-and-test-distributions.yml
rbs-jacob, line 1: I already made many other comments on this file in the other (now closed) PR #611 that were not addressed here.

.github/workflows/build-and-test-distributions.yml
whyitfor: Perhaps, but let's do this outside of this MR. Added an issue #628.

.github/workflows/build-and-test-distributions.yml
whyitfor: Changed to use the `make -C ofrak_core src/ofrak/gui/public` strategy

.github/workflows/publish-packages.yml
whyitfor: This is a manual trigger. The idea here is that users need to "opt in", to specify which packages they want to build. Hoping this will prevent accidental pushes

.github/workflows/build-and-test-distributions.yml
whyitfor: My thinking here is:
- Default start with latest, as failing tests could catch issues with compatibility with new systems as they come out
- If there is a specific need to test a specific version, we could do that / add it as needed

WDYT?

.github/workflows/build-and-test-distributions.yml
whyitfor: Removed the venvs.

.github/workflows/build-and-test-distributions.yml
whyitfor: Removed fail-fast: false.

.github/workflows/build-and-test-distributions.yml
whyitfor: I think most of these have been addressed. At this point though it will probably be most efficient to just comment directly on this MR, enough has changed that going back and forth is not going to be productive.

.github/workflows/build-and-test-distributions.yml
rbs-jacob: Sounds good to me.

.github/workflows/publish-packages.yml
```diff
@@ -7,46 +7,59 @@ on:
...
         required: true
         type: string
+      publish_target:
+        description: 'Where to publish packages'
+        required: true
+        type: choice
+        options:
+          - test-pypi
+          - pypi
+        default: test-pypi
```
rbs-jacob: The conditionals later in this workflow could probably be simplified if this was a boolean. Specifically, they could avoid brittle string equality checking. Any reason to have it be a two-choice enumeration instead of a boolean?

.github/actions/lfs-cache/action.yml
rbs-jacob, line 1: I think this action should also do the checkout. Then we can just replace every use of the checkout action with this, rather than having separate checkout and LFS cache steps.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,152 @@
...
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+    inputs:
```
rbs-jacob:
```suggestion
```

I'm not 100% sure, but I think that if this is empty, we don't need this key.

.github/actions/build-sdist/action.yml
```diff
@@ -0,0 +1,62 @@
...
+  source-dir:
+    description: 'Source directory containing the package'
+    required: true
+  lfs:
+    description: 'Whether this package requires LFS'
+    required: false
+    default: 'false'
+  python-version:
+    description: 'Python version to use for building'
+    required: true
```
rbs-jacob: It seems like we only use Python 3.9 for this workflow (as far as I can tell). Should we just set a default here and make this input optional?

.github/actions/build-sdist/action.yml
```diff
@@ -0,0 +1,62 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: pip
+        cache-dependency-path: |
+          requirements-*.txt
+          **/setup.py
+
+    - name: Install build tools
+      shell: bash -euxo pipefail {0}
```
rbs-jacob, line 36: Why set this for each step instead of as a default at the top? Does that not work here?

.github/actions/build-wheels/action.yml
```diff
@@ -0,0 +1,73 @@
...
+          requirements-*.txt
+          **/setup.py
+
+    - name: Set up Node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: '18'
+
+    - name: Build frontend
+      shell: bash -euxo pipefail {0}
```
rbs-jacob: Same as above: why manually specify the shell instead of setting defaults?

.github/actions/build-wheels/action.yml
```diff
@@ -0,0 +1,73 @@
...
+        python -m pip install -r requirements-pip.txt
+        python -m pip install -r requirements-build.txt
+
+    - name: Set Python version for cibuildwheel
+      id: py-version
+      shell: bash -euxo pipefail {0}
+      run: |
+        PY_VERSION="${{ inputs.python-version }}"
+        PY_VERSION_NO_DOT=${PY_VERSION//./}
+        echo "version=${PY_VERSION_NO_DOT}" >> $GITHUB_OUTPUT
```
rbs-jacob: Why make this a separate step? This could probably be folded into the step below and avoid passing variables between workflow steps.

.github/actions/build-wheels/action.yml
```diff
@@ -0,0 +1,73 @@
...
+        CIBW_BUILD: "cp${{ steps.py-version.outputs.version }}-*"
+        CIBW_SKIP: "*-win32 *-manylinux_i686"
+        CIBW_ARCHS_WINDOWS: "AMD64"
+        CIBW_ARCHS_MACOS: "x86_64 arm64"
+        CIBW_ARCHS_LINUX: "x86_64"
+        CIBW_BUILD_VERBOSITY: 1
+      run: |
+        cd ofrak_core
+        cibuildwheel --output-dir ../wheelhouse
+        ls -la wheelhouse/ || echo "No wheelhouse found"
```
rbs-jacob: This will succeed even if the wheelhouse is not found. But if that directory is not created, do we still want to continue? Or would it be better to fail here?

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+        path: dist-ofrak-capstone/
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: pip
+        cache-dependency-path: |
+          requirements-*.txt
+          **/setup.py
```
rbs-jacob: Do we have to worry about this caching the sdists that we're trying to test? Or are they not copied into `site-packages`?

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+
+    - name: Install cmake (macOS)
+      if: runner.os == 'macOS' && inputs.package-under-test == 'ofrak-angr'
+      shell: bash -euxo pipefail {0}
+      run: |
+        brew install cmake libmagic
```
rbs-jacob: Does this mean that we will not install `libmagic` unless we're testing with angr? Do we need to install libmagic on Linux, or is it already present in the runners?

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+        cmake --version
+
+    - name: Install packages
+      shell: bash -euxo pipefail {0}
+      run: |
+        python -m pip install -r requirements-pip.txt
+        python -m pip install -r requirements-dev.txt
+        # Install packages, in order of packages list
+        for package in ${{ inputs.packages }}; do
+          if [[ "$package" == "ofrak" ]]; then
```
rbs-jacob:
```suggestion
          if [ "$package" = "ofrak" ]; then
```

We'll be better off using portable POSIX shell conditionals.

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+            echo "Installing ofrak wheel. Available wheels"
+            ls -la wheelhouse/
+            echo ""
+            echo "Platform info:"
+            python -c "import sysconfig; print(f'Platform: {sysconfig.get_platform()}')"
+            python -c "import platform; print(f'Machine: {platform.machine()}')"
+            echo ""
```
rbs-jacob: It seems weird that this check is inside of the loop body?

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+            WHEEL_PATH=$(pip install --dry-run --find-links wheelhouse/ --no-index --no-deps ofrak 2>&1 | grep "Processing.*\.whl" | sed 's/Processing //' | head -1)
```
rbs-jacob:
```suggestion
            WHEEL_PATH="$(pip install --dry-run --find-links wheelhouse/ --no-index --no-deps ofrak 2>&1 | grep "Processing.*\.whl" | sed 's/Processing //' | head -1)"
```

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+            echo "Installing: $WHEEL_PATH"
+            pip install "$WHEEL_PATH"
```
rbs-jacob, line 115: For this to work and not install dependencies, does that mean that `ofrak` must come after others like `ofrak-type` and `ofrak-io` in the input list? If that's the case, should this be documented somewhere?

.github/actions/test-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+          else
+            echo "Installing ${package}..."
+            pip install dist-${package}/*.tar.gz
+          fi
+        done
+
+    - name: Run tests
+      shell: bash -euxo pipefail {0}
+      run: |
+        make -C ${{ inputs.source-dir }} ${{ inputs.test-target }} PYTHON=python
```
rbs-jacob, line 127:
```suggestion
        make -C ${{ inputs.source-dir }} ${{ inputs.test-target }}
```

Is the `PYTHON=python` just for Windows? I don't understand why it's included.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,152 @@
...
+  build-ofrak-capstone:
+    name: Build ofrak-capstone sdist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/build-sdist
+        with:
+          package-name: ofrak-capstone
+          source-dir: disassemblers/ofrak_capstone
+          python-version: '3.9'
```
rbs-jacob: I think I commented this elsewhere, but should the Python version just be a default value instead of being repeated over and over?

.github/actions/build-wheels/action.yml
```diff
@@ -0,0 +1,73 @@
...
+      env:
+        CIBW_BUILD: "cp${{ steps.py-version.outputs.version }}-*"
+        CIBW_SKIP: "*-win32 *-manylinux_i686"
+        CIBW_ARCHS_WINDOWS: "AMD64"
+        CIBW_ARCHS_MACOS: "x86_64 arm64"
+        CIBW_ARCHS_LINUX: "x86_64"
+        CIBW_BUILD_VERBOSITY: 1
+      run: |
+        cd ofrak_core
+        cibuildwheel --output-dir ../wheelhouse
```
rbs-jacob, line 59: Does this automatically find all packages to build wheels for? Is that why we need a bunch of different workflows and jobs to build source distributions, but only one workflow/job (spread across the matrix) to build wheels?

.github/workflows/test-package-matrix.yml
```diff
@@ -0,0 +1,57 @@
+name: Test Package Matrix
```
rbs-jacob, line 1: I think that this workflow just takes a package as input, and then tests it across a matrix? But I'm honestly not sure. A docstring might help explain why this job with one step isn't inlined above.

.github/workflows/build-and-test-distributions.yml
```diff
@@ -0,0 +1,152 @@
...
+    with:
+      package-under-test: ofrak
+      source-dir: ofrak_core
+      packages: ofrak-type ofrak-io ofrak-patch-maker ofrak pytest-ofrak
+      lfs: true
+      test-target: test-wheel
+
+  test-ofrak-angr:
+    uses: ./.github/workflows/test-package-matrix.yml
+    needs: [build-ofrak-type, build-ofrak-io, build-ofrak-patch-maker, build-pytest-ofrak, build-ofrak-wheels, build-ofrak-angr]
```
rbs-jacob:
```suggestion
    needs: 
      - build-ofrak-type
      - build-ofrak-io
      - build-ofrak-patch-maker
      - build-pytest-ofrak
      - build-ofrak-wheels
      - build-ofrak-angr
```

They don't have to be written like this, but just FYI that they can

rbs-jacob (changes requested):

.github/workflows/publish-packages.yml
whyitfor: As discussed offline, I've updated this to use the repository-url as options, but left explicit options for clarity on what the possible steps are.

.github/actions/lfs-cache/action.yml
whyitfor: As discussed, the checkout is needed in the workflows in order for GitHub to access this action itself. We can work around this by publishing this action in the future.

.github/workflows/build-and-test-distributions.yml
whyitfor: True. Removed here and in test-all.yml as well

.github/actions/build-sdist/action.yml
whyitfor: Updated to this effect.

.github/actions/test-package/action.yml
whyitfor: Yes -- added a comment to document this.

.github/actions/build-wheels/action.yml
whyitfor: Yes, cibuildwheel builds wheels for the current directory (in this case ofrak_core). The different matrix workflows allow us to build wheels that are compatible with different architecture.

.github/workflows/test-package-matrix.yml
whyitfor: Added a description for this for clarity.

.github/workflows/build-and-test-distributions.yml
whyitfor: Ok, god to know!

.github/actions/build-sdist/action.yml
whyitfor: As discussed, you cannot set a default shell for actions. It looks like we need a shell arugment for each of these steps -- shoudl we leave as is or make it `shell:bash`?

.github/actions/build-wheels/action.yml
whyitfor: Done.

.github/actions/build-wheels/action.yml
whyitfor: Updated so it fails here.

.github/actions/test-package/action.yml
whyitfor: Looked into this a bit. It should be fine but it's unclear. For now I've removed the `**/setup.py` from the cache in an abundance of caution. If we need to speed this up later we can revisit this.

.github/actions/test-package/action.yml
whyitfor: Simplified this -- always install cmake and libmagic.

.github/actions/test-package/action.yml
whyitfor: Removed.

.github/actions/test-package/action.yml
whyitfor: Yes. The description already says this:

```
description: 'Space-separated list of packages to install (in dependency order)'
```

.github/actions/build-sdist/action.yml
whyitfor: I've switched most back to "shell: bash"

.github/actions/build-sdist/action.yml
```diff
@@ -30,26 +27,25 @@ runs:
...
     - name: Build source distribution
-      shell: bash -euxo pipefail {0}
+      shell: bash
       run: |
         cd ${{ inputs.source-dir }}
         python -m build --sdist --outdir ${{ github.workspace }}/dist
 
     - name: List built packages
-      shell: bash -euxo pipefail {0}
+      shell: bash
```
rbs-jacob: We should revert these changes to `shell:`. If we're going to have a line for specifying the shell anyway, we definitely want to take full advantage of it. My recommendation is to either keep the previous thing or get rid of it entirely.

.github/actions/test-package/action.yml
```diff
@@ -129,6 +121,7 @@ runs:
         done
 
     - name: Run tests
-      shell: bash -euxo pipefail {0}
+      shell: bash
       run: |
-        make -C ${{ inputs.source-dir }} ${{ inputs.test-target }} PYTHON=python
+        # PYTHON=python needed for Windows compatibility where 'python3' is not available
+        make -C ${{ inputs.source-dir }} ${{ inputs.test-target }}
```
rbs-jacob:
```suggestion
        make -C ${{ inputs.source-dir }} ${{ inputs.test-target }} PYTHON=python
```

You said it was needed, but didn't add it back in?

.github/actions/build-sdist/action.yml
whyitfor: Ok, making all of them `bash -euxo pipefail {0}`

PR #623, opened 2025-09-18
Reduce the decompilation time of PyGhidra

NicolaasWeideman:

Reduce the decompilation time of PyGhidra by reusing cached unpacking results.

**Link to Related Issue(s)**
N/A

1. The `PyGhidraDecompilationAnalyzer` performs unpacking during decompilation ([src](https://github.com/redballoonsecurity/ofrak/blob/a51531ca1182fa1fab2aacee1e46bd85899d4ff7/disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py#L186)).
So, if the resource has already been unpacked recursively, unpacking will be performed a second time.
This pull request reuses cached unpacking results ([src](https://github.com/NicolaasWeideman/ofrak/blob/58a64996ef8ce4a44bd1f0ae450c0d472375aa24/disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py#L170)) and extends these with decompilation results ([src](https://github.com/NicolaasWeideman/ofrak/blob/58a64996ef8ce4a44bd1f0ae450c0d472375aa24/disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py#L186)).

1. The pull request uses persistent temporary directory ([src](https://github.com/NicolaasWeideman/ofrak/blob/58a64996ef8ce4a44bd1f0ae450c0d472375aa24/disassemblers/ofrak_pyghidra/ofrak_pyghidra/components/pyghidra_components.py#L94)) to allow Ghidra to create and reuse a single project for the binary.

These changes should reduce decompilation time.

@rbs-jacob

### DIFF
--- a/disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
+++ b/disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from tempfile import mkdtemp
+from tempfile312 import mkdtemp
 import os
 from typing import Dict
 from xml.etree import ElementTree
--- a/disassemblers/ofrak_pyghidra/requirements.txt
+++ b/disassemblers/ofrak_pyghidra/requirements.txt
@@ -1,2 +1,3 @@
 pyghidra~=2.1.0
 importlib-resources  # A workaround for https://github.com/redballoonsecurity/ofrak/issues/398
+tempfile312~=1.0.1


paulnoalhyt:
This PR looks good to me. I noticed significant decompilation speed improvements as advertised.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from tempfile import TemporaryDirectory
+from tempfile import mkdtemp
```
rbs-jacob:
```suggestion
from tempfile312 import mkdtemp
```

Might be a good idea to import from [Albert's backported version of `tempfile`](https://github.com/rbs-forks/tempfile312), even though I am not sure we explicitly need any of his backported features here. This suggestion may necessitate changing one of the requirements files. See the other places it's imported in OFRAK for reference.

https://github.com/redballoonsecurity/ofrak/blob/627cb732e3bf7409b820bd12431aabb6e38d364e/ofrak_core/requirements.txt#L22

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -380,6 +381,23 @@ def _decompile(func, decomp_interface, task_monitor):
...
+        from ghidra.app.decompiler import DecompInterface
+        from ghidra.util.task import TaskMonitor
+
+        decomp = DecompInterface()
+        decomp.openProgram(flat_api.getCurrentProgram())
+        program = flat_api.getCurrentProgram()
+        function_manager = program.getFunctionManager()
+        func_to_decomp = {}
+        for func in function_manager.getFunctions(True):
+            cb_key = f"func_{func.getEntryPoint().getOffset()}"
```
rbs-jacob, line 395: The offset doesn't need to be converted to hex, right? (Just checking.)

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -164,32 +164,27 @@ class PyGhidraDecompilationAnalyzer(CachedDecompilationAnalyzer):
...
+            try:
+                program_attrs = program_r.get_attributes(ProgramAttributes)
+            except NotFoundError:
+                program_attrs = await program_r.analyze(ProgramAttributes)
+            analysis = unpack(
+                program_file, False, language=_arch_info_to_processor_id(program_attrs)
+            )
+        if "decompilation" not in analysis[cb_key]:
+            program_file = analysis["metadata"]["path"]
+            await program_r.flush_data_to_disk(program_file)
```
rbs-jacob: Since we're changing the program file on disk during analysis, we should probably document somewhere that we're assuming the path is in a temporary directory. We don't want to end up with a situation where `unpack` or `analyze` is called on a non-temporary file, and it gets modified unexpectedly.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -91,39 +91,39 @@ def __init__(
...
-                resource.get_id(),
-                unpack(
-                    program_file,
-                    decomp,
-                    language=_arch_info_to_processor_id(program_attrs),
-                    base_address=base_address,
-                ),
-            )
-            return PyGhidraProject()
+        tempdir = mkdtemp(prefix="rbs-pyghidra-bin")
```
rbs-jacob, line 94: These calls to `mkdtemp` will leak tempfiles, right? 

If so, probably not an issue if we can't work around it while still using the caching results across runs. But we should consider if there is an easy way to clean up when OFRAK terminates. Not worth blocking the PR over; just worth thinking about.

rbs-jacob (changes requested):
Some minor changes, and this should be good to go.

Thanks for the attention to OFRAK speed!

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
NicolaasWeideman: Addressed in [this](https://github.com/redballoonsecurity/ofrak/pull/623/commits/7cf009d54a7130781b07d8506881b5cde0551ba8) commit.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
NicolaasWeideman: I think this specific call to `program_r.flush_data_to_disk(program_file)` (that executes every time a ComplexBlock is decompiled) was a mistake. It should only be necessary to create the temporary file `rbs-pyghidra-bin*/program` once before calling `unpack`. I pushed [this](https://github.com/redballoonsecurity/ofrak/pull/623/commits/31f2978b59dbfa1334b99e669a44318b4ba045a2) commit to make the necessary changes.

I don't think it'll be an issue if `unpack` or `analyze` are called with a non-temporary file. I don't believe they are changing the file in any way. But it might be something to keep in mind if we later decide to remove this file after OFRAK completes.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
NicolaasWeideman: That's right, the temporary file `rbs-pyghidra-bin*/program` will persist after OFRAK completes. The only place I can think of to put the cleanup code would be at the end of the `OFRAK`'s run function ([here](https://github.com/redballoonsecurity/ofrak/blob/627cb732e3bf7409b820bd12431aabb6e38d364e/ofrak_core/src/ofrak/ofrak_context.py#L200)). But, I'm not sure how to collect everything that needs cleanup here.

rbs-jacob: Yeah, not worth adding any complicated cleanup code. Thanks for confirming my understanding.

PR #631, opened 2025-09-22
Add ofrak_pyghidra to GUI script builder

paulnoalhyt:

Add `ofrak_pyghidra` to OFRAK GUI script builder.

**Link to Related Issue(s)**

None.

Add `ofrak_pyghidra` to OFRAK GUI script builder: add relevant imports and modify the tests accordingly.


ofrak_core/src/ofrak/gui/script_builder.py
```diff
@@ -84,6 +84,11 @@ async def main(ofrak_context: OFRAKContext, root_resource: Optional[Resource] =
 
             ofrak.discover(ofrak_ghidra)
 
+        if False:
+            import ofrak_pyghidra
+
+            ofrak.discover(ofrak_pyghidra)
+
```
rbs-jacob, line 91: How do you feel about replacing the following with PyGhidra instead of just appending the PyGhidra stuff?

```
if False:
    import ofrak_ghidra

    ofrak.discover(ofrak_ghidra)
```

If we get rid of the above, it might be a bit less confusing.

rbs-jacob (changes requested):
Thanks for this change, I totally didn't think of this when the original PyGhidra PR was going in!

Let's discuss my one small comment, then we can get this merged quickly.

ofrak_core/src/ofrak/gui/script_builder.py
paulnoalhyt: My opinion is that the `ofrak_ghidra` import should be removed when we phase out the whole `ofrak_ghidra` backend.

rbs-jacob: In that case maybe a comment pointing to which is the "old" back end and which is the "new" one? I think that would be a reasonable compromise.

paulnoalhyt: Just added the comments

rbs-jacob (approved):
Looks great!

PR #634, opened 2025-09-24
MCP Endpoints

rbs-afflitto:

Add endpoints to the OFRAK GUI backend to support OFRAK MCP

**Link to Related Issue(s)**

N/A

Adds the following endpoints to the GUI backend:

- `get_view_schema`: Gets the schema for a ResourceView (lists attributes and their types)
- `add_view_to_resource`: Adds a ResourceView to an existing resource
- `remove_tag`: Removes a ResourceTag from a resource
- `remove_component`: Removes a component (and attributes if applicable) from a resource

Change the behavior of the following endpoints:
- `get_components`: Adds a new parameter incl_docstrings, if true it includes components' docstrings, targets, and outputs with the response
- `get_config_for_component`: Specifies whether each argument for the config is optional
- `run_component`: Fix calling run_component with no config (for components with optional configs)

The PatchMaker Toolchain now includes stdout and stderr in the raised ValueError upon toolchain failure.

@rbs-jacob

### DIFF
--- a/ofrak_core/src/ofrak/gui/server.py
+++ b/ofrak_core/src/ofrak/gui/server.py
@@ -753,10 +753,7 @@ class AiohttpOFRAKServer:
             body["view_type"], Type[ResourceViewInterface]
         )
         field_values = body["fields"]
-
-        # Create the view instance with the provided field values
         view_instance = view_type(**field_values)
-
         script_str = (
             """
         {resource}"""
@@ -767,7 +764,6 @@ class AiohttpOFRAKServer:
         )
         await self.script_builder.add_action(resource, script_str, ActionType.MOD)
 
-        # Add the view to the resource
         resource.add_view(view_instance)
         await resource.save()
         return json_response(self._serialize_resource(resource))
--- a/ofrak_core/tests/unit/test_ofrak_server.py
+++ b/ofrak_core/tests/unit/test_ofrak_server.py
@@ -584,6 +584,9 @@ async def test_add_tag(ofrak_client: TestClient, hello_elf):
         json="ofrak.core.apk.Apk",
     )
     assert resp.status == 200
+    # Check tag is added
+    add_tag_resp = await ofrak_client.get(f"/{resource_id}/")
+    assert "ofrak.core.apk.Apk" in (await add_tag_resp.json())["tags"]
 
 
 async def test_remove_tag(ofrak_client: TestClient, hello_elf):
@@ -597,12 +600,19 @@ async def test_remove_tag(ofrak_client: TestClient, hello_elf):
         f"/{resource_id}/add_tag",
         json="ofrak.core.apk.Apk",
     )
+    # Check tag is added
+    add_tag_resp = await ofrak_client.get(f"/{resource_id}/")
+    assert "ofrak.core.apk.Apk" in (await add_tag_resp.json())["tags"]
+
     # Then remove it
     resp = await ofrak_client.post(
         f"/{resource_id}/remove_tag",
         json="ofrak.core.apk.Apk",
     )
     assert resp.status == 200
+    # Check tag is removed
+    rem_tag_resp = await ofrak_client.get(f"/{resource_id}/")
+    assert "ofrak.core.apk.Apk" not in (await rem_tag_resp.json())["tags"]
 
 
 async def test_get_view_schema(ofrak_client: TestClient):
@@ -632,6 +642,17 @@ async def test_add_view_to_resource(ofrak_client: TestClient, hello_elf):
         },
     )
     assert resp.status == 200
+    add_attribute_resp = await ofrak_client.get(f"/{resource_id}/")
+    add_attribute_dict = {
+        attr[0]: attr[1][1] for attr in (await add_attribute_resp.json())["attributes"]
+    }
+    assert (
+        "ofrak.model._auto_attributes.AttributesType[FilesystemEntry]" in add_attribute_dict.keys()
+    )
+    fields = add_attribute_dict["ofrak.model._auto_attributes.AttributesType[FilesystemEntry]"]
+    assert fields["name"] == "test_file"
+    assert fields["stat"] == None
+    assert fields["xattrs"] == None
 
 
 async def test_remove_component(ofrak_client: TestClient, hello_elf):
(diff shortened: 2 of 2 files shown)


ofrak_core/tests/unit/test_ofrak_server.py
```diff
@@ -586,6 +586,72 @@ async def test_add_tag(ofrak_client: TestClient, hello_elf):
...
+    await ofrak_client.post(
+        f"/{resource_id}/add_tag",
+        json="ofrak.core.apk.Apk",
+    )
+    # Then remove it
+    resp = await ofrak_client.post(
+        f"/{resource_id}/remove_tag",
+        json="ofrak.core.apk.Apk",
+    )
+    assert resp.status == 200
```
rbs-jacob, line 612: Maybe worth adding tests here to confirm that the tag was added, and that the tag was actually removed?

ofrak_core/tests/unit/test_ofrak_server.py
```diff
@@ -586,6 +586,72 @@ async def test_add_tag(ofrak_client: TestClient, hello_elf):
...
+    # First run a component to add it
+    await ofrak_client.post(f"/{resource_id}/analyze")
+    # Then try to remove a component
+    resp = await ofrak_client.post(
+        f"/{resource_id}/remove_component",
+        params={"component": "DataSummaryAnalyzer"},
+    )
+    assert resp.status == 200
+    resp_body = await resp.json()
+    assert "success" in resp_body
```
rbs-jacob, line 678: Worth adding a check to confirm that the component was actually removed?

ofrak_core/tests/unit/test_ofrak_server.py
```diff
@@ -586,6 +586,72 @@ async def test_add_tag(ofrak_client: TestClient, hello_elf):
...
+    create_body = await create_resp.json()
+    resource_id = create_body["id"]
+    resp = await ofrak_client.post(
+        f"/{resource_id}/add_view_to_resource",
+        json={
+            "view_type": "ofrak.core.filesystem.File",
+            "fields": {"name": "test_file", "stat": None, "xattrs": None},
+        },
+    )
+    assert resp.status == 200
```
rbs-jacob, line 644: Add a check that confirms that the attributes appear when you view the child's metadata in the GUI? (Of course the check itself should just use the API calls that populate that part of the GUI.)

ofrak_core/src/ofrak/gui/server.py
```diff
@@ -689,6 +697,81 @@ async def create_mapped_child(self, request: Request) -> Response:
...
+        schema = {
+            "name": view_type.__name__,
+            "type": self._convert_to_class_name_str(view_type),
+            "fields": fields_info,
+            # Optionally include the composed attribute types for reference
+            "composed_attributes": self._serializer.to_pjson(
+                list(getattr(view_type, "composed_attributes_types", [])),
+                List[Type[ResourceAttributes]],
+            ),
+        }
```
rbs-jacob, line 740: Is the schema format the same just for simplicity, or for a technical reason? I ask because if we want to keep them synchronized, it may make sense to also add a comment pointer in the `get_config_for_component` noting that it should match this.

ofrak_core/src/ofrak/gui/server.py
```diff
@@ -689,6 +697,81 @@ async def create_mapped_child(self, request: Request) -> Response:
...
+        """
+        resource = await self._get_resource_for_request(request)
+        body = await request.json()
+
+        view_type: Type[ResourceViewInterface] = self._serializer.from_pjson(
+            body["view_type"], Type[ResourceViewInterface]
+        )
+        field_values = body["fields"]
+
+        # Create the view instance with the provided field values
```
rbs-jacob:
```suggestion
```

This comment doesn't really add anything in my opinion.

ofrak_core/src/ofrak/gui/server.py
```diff
@@ -689,6 +697,81 @@ async def create_mapped_child(self, request: Request) -> Response:
...
+            """
+        {resource}"""
+            f""".add_view({view_instance.__repr__()})
+        """
+            """await {resource}.save()
+        """
+        )
+        await self.script_builder.add_action(resource, script_str, ActionType.MOD)
+
+        # Add the view to the resource
```
rbs-jacob:
```suggestion
```

ofrak_core/src/ofrak/gui/server.py
```diff
@@ -872,6 +955,28 @@ async def add_tag(self, request: Request) -> Response:
...
         return json_response(self._serialize_resource(resource))
 
+    @exceptions_to_http(SerializedError)
+    async def remove_tag(self, request: Request) -> Response:
+        resource = await self._get_resource_for_request(request)
+        tag = self._serializer.from_pjson(await request.json(), ResourceTag)
+        script_str = (
+            """
+        {resource}.remove_tag"""
+            f"""({tag.__name__})"""
```
rbs-jacob, line 961: Should we add the `resource.save()` to this as well?

rbs-jacob (changes requested):

ofrak_core/src/ofrak/gui/server.py
rbs-afflitto: This is just for simplicity, I don't see a problem if these get out of sync (as long as the MCP docstrings remain in sync with whatever format this returns)

ofrak_core/src/ofrak/gui/server.py
rbs-afflitto: The .save() is a few lines below. This follows the convention in add_tag where the .save() is added as a separate script action

ofrak_core/tests/unit/test_ofrak_server.py
rbs-afflitto: Added checks

ofrak_core/tests/unit/test_ofrak_server.py
rbs-afflitto: Added checks - the `DataSummaryAnalyzer` wasn't actually running, so this test wasn't doing much before checking the attributes. I changed it to explicitly run the MD5 analyzer and then check the attribute is removed.

rbs-jacob (approved):
Looks good to me!

PR #637, opened 2025-09-26
Add `ProgramAttributes` from the GUI

paulnoalhyt:

Add ProgramAttributes from the OFRAK GUI. Rewrote the ihex components.

##### Allow users to add `ProgramAttributes` from the GUI:
Allow users to add `ProgramAttributes` to a resource in the OFRAK GUI. This is particularly useful when dealing with Intel Hex files which do not contain info about the CPU architecture, making them impossible to analyze in the GUI before.

- added 2 endpoints in the server to:
    - get all possible `ProgramAttributes` fields (`InstructionSet`, `SubInstructionSet`, `BitWidth`, `Endianness`, `ProcessorType`).
    - add `ProgramAttributes` to a resource.
- added tests for these 2 endpoints.
- `test_pyghidra_components.py`: added test for unpacking ihex file with pyghidra, after adding `ProgramAttributes`.

###### some context about how ELF files are handled in OFRAK:

- ELF file has the `Program` tag, so when we analyze an ELF with a backend, we give the backend the full ELF with headers, etc.
- The ELF unpacker creates `CodeRegions` that can be unpacked by the backend.

###### some context about how ihex files were handled before this PR:
- the `Ihex` is just considered as a `GenericFile`.
- the `IhexProgram` gets the `Program` tag. It's the binary content of all sections present in the ihex (binary blob of concatenated sections). This is given to the backend. The issue is that the backend doesn't receive the ihex file, so it has no idea about the sections present in this binary blob.
- the `IhexUnpacker` doesn't create `CodeRegions` but an `IhexProgram` (binary blob mentioned above).
    - the `IhexProgramUnpacker` then creates `ProgramSections` which are not really `CodeRegions`.
- it's harder to map the backend segments to these `ProgramSections`. For example, the `PyGhidraCodeRegionUnpacker` is made to run on `CodeRegions` and not `ProgramSections`.

###### proposed rewriting:
- the `Ihex` file gets the `Program` tag. That way the full ihex file is given to the backend for analysis.
- the `IhexUnpacker` created `CodeRegions`, like the `ElfUnpacker`. This makes sense because when loading an ihex file in Ghidra, it assumes that all sections are RWX.
- remove the binary blob intermediary representation, which doesn't add any value. The binary content is contained in the `CodeRegions` data.
- tests were updated.

Happy to discuss this rewriting in comments!

##### Bug fixes/improvements:
- `pyghidra_components.py`:
    - when running the `PyGhidraAutoAnalyzer`, the resource was flushed to disk, so it was packed. This removes all child resources that were created before running the `PyGhidraAutoAnalyzer`.  This was an issue if for example, you load an ihex file, unpack it (so it creates `CodeRegions`), then run `unpack_recursively()`. As the binary was packed by the `PyGhidraAutoAnalyzer`, the `CodeRegions` were be deleted, and the `unpack_recursively` failed because the `CodeRegions` it was supposed to unpack disapeared. To avoid that, I put `pack=False` in `flush_data_to_disk`. This solves my issue, but is not ideal (what if the user modified some data in the ihex before running the `PyGhidraAutoAnalyzer`?!). Happy to discuss other ideas to solve this issue!
    - the `PyGhidraAutoAnalyzer` now checks for `ProgramAttributes` before running `unpack`.
- `pyghidra_analysis.py`: added better error message when hitting `ghidra.app.util.opinion.LoadException: No load spec found`, to give the user a hint that they should add `ProgramAttributes`.
- `AddTagView.svelte` and `AddProgramAttributesView.svelte`: updated `refreshResource()`  so that the attribute panel in the GUI is actually refreshed with the new tag/ProgramAttributes.

No

### DIFF
--- a/ofrak_core/tests/unit/test_ofrak_server.py
+++ b/ofrak_core/tests/unit/test_ofrak_server.py
@@ -2184,6 +2184,9 @@ async def test_get_all_program_attributes(ofrak_client: TestClient):
     - Endianness options are included
     - Processor type options are included
     - Specific known values exist for each category (ARM, ARMv4T, etc.)
+
+    Requirements Mapping:
+    - REQ2.2
     """
     resp = await ofrak_client.get(f"/get_all_program_attributes")
     assert resp.status == 200
@@ -2215,6 +2218,9 @@ async def test_add_program_attributes(ofrak_client: TestClient, hello_elf):
     - Optional fields can be set to None
     - Multiple program attribute updates can be applied to the same resource
     - The add_program_attributes endpoint handles both complete and minimal configurations
+
+    Requirements Mapping:
+    - REQ2.2
     """
     create_resp = await ofrak_client.post(
         "/create_root_resource", params={"name": "hello_elf"}, data=hello_elf
--- a/disassemblers/ofrak_pyghidra/tests/test_pyghidra_components.py
+++ b/disassemblers/ofrak_pyghidra/tests/test_pyghidra_components.py
@@ -1,5 +1,8 @@
 """
 This module tests the PyGhidra component, including unpackers, disassembly and decompilation components.
+
+Requirements Mapping:
+- REQ1.2
 """
 from ofrak.core.instruction import Instruction
 import os
@@ -260,6 +263,9 @@ async def test_PIE_code_regions(program_resource):
     - Code regions are correctly created at expected virtual addresses
     - PIE binary base address (0x100000) is properly handled
     - Code region addresses and sizes match expected values
+
+    Requirements Mapping:
+    - REQ1.2
     """
     await program_resource.unpack()
     code_regions = await program_resource.get_descendants_as_view(
@@ -341,6 +347,14 @@ async def ihex_resource(ofrak_context: OFRAKContext):
 async def test_ihex_unpacking(ihex_resource):
     """
     Test that adding ProgramAttributes to an Ihex file allows for unpacking with ofrak_pyghidra.
+
+    This test verifies that:
+    - Intel HEX files can be unpacked with PyGhidra when ProgramAttributes are provided
+    - Complex blocks are correctly identified after recursive unpacking
+    - Expected function names are discovered in the unpacked binary
+
+    Requirements Mapping:
+    - REQ1.2
     """
     program_attributes = ProgramAttributes(
         InstructionSet.X86,


frontend/src/views/AddProgramAttributesView.svelte
```diff
@@ -0,0 +1,218 @@
...
+
+  onMount(async () => {
+    try {
+      await fetch(`${$settings.backendUrl}/get_all_program_attributes`).then(
+        async (r) => {
+          if (!r.ok) {
+            throw Error(JSON.stringify(await r.json(), undefined, 2));
+          }
+
+          //console.log(r.json())
```
rbs-jacob:
```suggestion
```

frontend/src/views/AddProgramAttributesView.svelte
```diff
@@ -0,0 +1,218 @@
...
+          r.json().then((ofrakProgramAttributes) => {
+            ofrakProgramAttributesPromise = ofrakProgramAttributes;
+          });
```
rbs-jacob: Since this whole function is `async`, you can await `r.json()` here without using a `.then` callback.

frontend/src/views/AddProgramAttributesView.svelte
```diff
@@ -0,0 +1,218 @@
...
+        },
+      ]);
+      await $selectedResource.add_program_attributes(program_attributes);
+      await refreshResource();
+    }
+  }
+
+  onMount(async () => {
+    try {
+      await fetch(`${$settings.backendUrl}/get_all_program_attributes`).then(
```
rbs-jacob: Since this whole function is async, you can await `fetch` here without using a `.then` callback.

frontend/src/views/AddProgramAttributesView.svelte
```diff
@@ -0,0 +1,218 @@
...
+    ) {
+      modifierView = undefined;
+      let program_attributes = JSON.stringify([
+        "ofrak.core.architecture.ProgramAttributes",
+        {
+          isa: selected_attribute["isa"],
+          sub_isa: selected_attribute["sub_isa"],
+          bit_width: selected_attribute["bit_width"],
+          endianness: selected_attribute["endianness"],
+          processor: selected_attribute["processor"],
```
rbs-jacob: In JavaScript, object field access is equivalent to dictionary-style field access. So `selected_attribute.isa` is the same as `selected_attribute["isa"]`.

Might be worth changing these to the more concise syntax?

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -17,107 +17,120 @@ def _parse_offset(java_object):
...
+                        main_dictionary[dw_key] = dw
+                    main_dictionary[cb_key] = cb
+    except Exception as e:
+        if "toString" in dir(e) and "No load spec found" in e.toString():
+            raise Exception(
+                e.toString()
+                + "\nTry adding ProgramAttributes to you binary before running a Ghidra analyzer/unpacker!"
+            )
+        else:
+            raise e
```
rbs-jacob:
```suggestion
    except <SOMETHING MORE SPECIFIC> as e:
        if "toString" in dir(e) and "No load spec found" in e.toString():
            raise <SOMETHING MORE SPECIFIC>(
                e.toString()
                + "\nTry adding ProgramAttributes to you binary before running a Ghidra analyzer/unpacker!"
            )
        else:
            raise
```

We want to avoid catching raw `Exception`s if possible. We also want to avoid raising them. We should probably replace these with a more specific `Exception` instance such as `RuntimeError`, `ValueError`, or something even more specific.

Worth changing in other parts of the code where you see this, as well.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -17,107 +17,120 @@ def _parse_offset(java_object):
...
-                    except Exception as e:
-                        print(e, traceback.format_exc())
-                        decompilation = ""
-                    cb["decompilation"] = decompilation
-                bb_model = BasicBlockModel(flat_api.getCurrentProgram())
-                basic_blocks, data_words = _unpack_complex_block(
-                    func, flat_api, bb_model, BigInteger.ONE
+    try:
+        with pyghidra.open_program(program_file, language=language) as flat_api:
+            # Java packages must be imported after pyghidra.start or pyghidra.open_program
```
rbs-jacob, line 26: Not worth blocking this PR for this, but I wonder if this function can be refactored to be a bit smaller? I think that might make it a bit easier to read.

There are helper functions like `_unpack_basic_block` that get called, but there is still a lot of logic before and after those calls. Can any of that logic before and after the helper function calls be moved into the helper functions themselves?

disassemblers/ofrak_pyghidra/CHANGELOG.md
```diff
@@ -3,13 +3,14 @@ All notable changes to `ofrak-pyghidra` will be documented in this file.
...
 
-## [Unreleased 0.2.0rc3](https://github.com/redballoonsecurity/ofrak/tree/master)
+## [Unreleased 0.2.0rc4](https://github.com/redballoonsecurity/ofrak/tree/master)
 
 ### Fixed
 - Fix Ghidra and pyghidra CodeRegion unpacker to take into account the base address that ghidra sets for PIE executables.([#627](https://github.com/redballoonsecurity/ofrak/pull/627))
 ### Changed
 - Reduce the decompilation time of PyGhidra by reusing cached unpacking results. ([#623](https://github.com/redballoonsecurity/ofrak/pull/623))
 - Improve `ofrak_pyghidra` decompilation: more strings and symbol names for cross-references in decompilation. ([#633](https://github.com/redballoonsecurity/ofrak/pull/633))
+- Look for `ProgramAttributes` before unpacking. Do not pack when flushing to disk. Better error message when Ghidra raises "No load spec found". Added test for unpacking ihex with `ofrak_pyghidra` ([#637](https://github.com/redballoonsecurity/ofrak/pull/637))
```
rbs-jacob:
```suggestion
- Improve unpacking logic, error messages, and testing for `ofrak_pyghidra` auto analyzer ([#637](https://github.com/redballoonsecurity/ofrak/pull/637))
```

ofrak_core/src/ofrak/core/ihex.py
```diff
@@ -32,106 +31,70 @@ class Ihex(GenericBinary):
...
+                - binfile.minimum_address : segment.maximum_address
+                - binfile.minimum_address
+            ]
             await resource.create_child_from_view(
-                ProgramSection(seg_vaddr_range.start, seg_vaddr_range.length()),
-                data_range=segment_data_range,
+                CodeRegion(
+                    segment.minimum_address, segment.maximum_address - segment.minimum_address
+                ),
+                data=segment_data,
```
rbs-jacob, line 71: Is it necessarily the case that all unpacked IHEX segments are code regions? No data segments? I don't know enough about IHEX to say for sure, but wanted to confirm. 

I generally like that `Ihex` now inherits from `Program`, but am uncertain about this particular change.

ofrak_core/CHANGELOG.md
```diff
@@ -83,6 +84,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Update `DataSummary` handling so that it is no longer stored in `ResourceModel`, but rather retrievable via `DataSummaryAnalyzer.get_data_summary` ([#598](https://github.com/redballoonsecurity/ofrak/pull/598))
 - Update dependencies: aiohttp>=3.12.14, beartype~=0.20.0, ubi-reader==0.8.12 ([#613](https://github.com/redballoonsecurity/ofrak/pull/613))
 - Add `ofrak_pyghidra` to OFRAK GUI script builder ([#631](https://github.com/redballoonsecurity/ofrak/pull/631))
+- rewrote intel hex components ([#637](https://github.com/redballoonsecurity/ofrak/pull/637))
```
rbs-jacob:
```suggestion
- Rewrote intel hex components ([#637](https://github.com/redballoonsecurity/ofrak/pull/637))
```

Might be worth adding a couple words on what justified the rewrite and/or what changed. Thinking from the perspective of someone reading the CHANGELOG, they're going to wonder whether and how this affects them, so we want to clarify that as much as possible up-front.

ofrak_core/tests/unit/test_ofrak_server.py
```diff
@@ -1540,3 +1540,63 @@ async def test_get_project_by_resource_id(ofrak_client: TestClient, test_project
     project_resp = await ofrak_client.get(f"/{resource_id}/get_project_by_resource_id")
     project = await project_resp.json()
     assert project["session_id"] == id
+
+
+async def test_get_all_program_attributes(ofrak_client: TestClient):
```
whyitfor, line 2174: We need docstrings for these tests. Look at examples.

ofrak_core/tests/unit/test_ofrak_server.py
```diff
@@ -1540,3 +1540,63 @@ async def test_get_project_by_resource_id(ofrak_client: TestClient, test_project
...
+    assert "endianness" in resp_body_dict
+    assert "processor" in resp_body_dict
+    assert "ofrak_type.architecture.InstructionSet.ARM" in resp_body_dict["isa"]
+    assert "ofrak_type.architecture.SubInstructionSet.ARMv4T" in resp_body_dict["sub_isa"]
+    assert "ofrak_type.bit_width.BitWidth.BIT_16" in resp_body_dict["bit_width"]
+    assert "ofrak_type.endianness.Endianness.LITTLE_ENDIAN" in resp_body_dict["endianness"]
+    assert "ofrak_type.architecture.ProcessorType.ARM926EJ_S" in resp_body_dict["processor"]
+
+
+async def test_add_program_attributes(ofrak_client: TestClient, hello_elf):
```
whyitfor: Docstring needed explaining what test does

ofrak_core/src/ofrak/core/ihex.py
rbs-afflitto: IHEX doesn't have a concept of memory permissions or code vs data like Elf does. (See https://developer.arm.com/documentation/ka003292/latest/). We usually find IHEX used for devices where all memory is executable, so it's simpler to treat everything as CodeRegions. This change also plays better with `ofrak_cached_disassembly` as it allows autoanalysis to load the cached CodeRegion information.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
rbs-afflitto: Hm this is tricky. The exception raised here is a Ghidra `LoadException` which is raised from Java. We can't easily import it outside of the try/except block though, because the `ghidra` import is only importable inside of the `with ghidra.open_program()` block. I think it's best to catch `Exception` here but to raise a more specific exception.

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
rbs-afflitto: See #667

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
rbs-jacob: That makes sense to me! If we do that, maybe worth adding a comment explaining why? Always helpful to have comments like that when looking back so that I (or someone else) doesn't go changing things that have good rationale

rbs-afflitto: Ok I added a comment here.

ofrak_core/tests/unit/test_ofrak_server.py
```diff
@@ -1543,6 +2172,19 @@ async def test_get_project_by_resource_id(ofrak_client: TestClient, test_project
...
+    This test verifies that:
+    - The get_all_program_attributes endpoint is accessible
+    - All program attribute categories are returned
+    - ISA (Instruction Set Architecture) options are included
+    - Sub-ISA options are included
+    - Bit width options are included
+    - Endianness options are included
+    - Processor type options are included
+    - Specific known values exist for each category (ARM, ARMv4T, etc.)
+    """
```
rbs-jacob, line 2190: According to the OFRAK contributor guide, the docstrings for the tests should map back to what specific requirements are verified by the test.

<https://ofrak.com/docs/contributor-guide/getting-started.html#requirements-and-test-documentation>

@whyitfor can weigh in further on the how and why

ofrak_core/tests/unit/test_ofrak_server.py
whyitfor: @rbs-afflitto, has this been addressed?

Issue #616, opened 2025-09-16
Unpacking CPIO filesystem fails on MacOS because of permissions


paulnoalhyt:
**What is the problem? (Here is where you provide a complete Traceback.)**

On MacOS, unpacking a `CpioFilesystem` fails with a Permission Denied error:

```
python3 -m ofrak gui -b angr -H 0 -p 8877
GUI is being served on http://0:8877/
[     iso9660.py:  194] El Torito images are not currently supported
[      server.py:  122] Exception raised in aiohttp endpoint
ofrak.service.job_service_i.ComponentAutoRunFailure: Component CpioUnpacker failed when running on bea2e5503bad4d95a53155fcb8562211. Component was chosen because it matched filters (((ComponentTypeFilter(Unpacker) or ComponentTypeFilter(Identifier))) and ((ComponentTypeFilter(Identifier) and ComponentTargetFilter(FilesystemRoot, CpioFilesystem)) or (ComponentTypeFilter(Analyzer) and ComponentTargetFilter(FilesystemRoot, CpioFilesystem)) or (not ComponentTypeFilter(Identifier) and not ComponentTypeFilter(Analyzer) and (ComponentTargetFilter(CpioFilesystem) then ComponentTargetFilter(FilesystemRoot)))))

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/ofrak/ofrak_core/ofrak/gui/server.py", line 120, in wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ofrak/ofrak_core/ofrak/gui/server.py", line 486, in unpack
    raise e
  File "/ofrak/ofrak_core/ofrak/gui/server.py", line 482, in unpack
    result = await resource.unpack()
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/ofrak/ofrak_core/ofrak/resource.py", line 461, in unpack
    return await self.auto_run(all_identifiers=True, all_unpackers=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/ofrak/ofrak_core/ofrak/resource.py", line 434, in auto_run
    components_result = await self._job_service.run_components(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #647, opened 2025-10-03
Bugfix/cpio libarchive

paulnoalhyt:

Reimplement the CPIO components using `libarchive`.

**Link to Related Issue(s)**

Addresses https://github.com/redballoonsecurity/ofrak/issues/616

Updated the CPIO components to use `libarchive` instead of the `cpio` command. The advantage is that unpacking is done in memory, files aren't actually created on disk. This fixes the 2 issues documented in https://github.com/redballoonsecurity/ofrak/issues/616. I added regression tests for the 2 aforementioned issues, plus a full unpack-pack-unpack test to make sure all metadata is preserved.

No

### DIFF
--- a/ofrak_core/src/ofrak/core/cpio.py
+++ b/ofrak_core/src/ofrak/core/cpio.py
@@ -1,30 +1,54 @@
 import asyncio
 import logging
-import tempfile312 as tempfile
+import os
+import stat
 from dataclasses import dataclass
 from enum import Enum
-from subprocess import CalledProcessError
 
 from ofrak.component.analyzer import Analyzer
 from ofrak.component.packer import Packer
 from ofrak.component.unpacker import Unpacker
 from ofrak.core.binary import GenericBinary
-from ofrak.core.filesystem import File, Folder, FilesystemRoot, SpecialFileType
+from ofrak.core.filesystem import (
+    File,
+    Folder,
+    FilesystemRoot,
+    SpecialFileType,
+    SymbolicLink,
+    CharacterDevice,
+    BlockDevice,
+    FIFOPipe,
+)
 from ofrak.core.magic import MagicMimePattern, MagicDescriptionPattern, Magic
 from ofrak.model.component_model import ComponentExternalTool
 from ofrak.resource import Resource
+from ofrak.service.resource_service_i import ResourceFilter
 from ofrak_type.range import Range
+from ofrak.core.filesystem import FilesystemEntry
+
+try:
+    import libarchive
+    LIBARCHIVE_INSTALLED = True
+except OSError:
+    LIBARCHIVE_INSTALLED = False
+
+class LibarchiveTool(ComponentExternalTool):
+    def __init__(self):
+        super().__init__(
+            "libarchive",
+            "https://www.libarchive.org/",
+            install_check_arg="",
+            apt_package="libarchive-dev",
+            brew_package="libarchive",
+            choco_package="libarchive",
+        )
+
+    async def is_tool_installed(self) -> bool:
+        return LIBARCHIVE_INSTALLED
+
+LIBARCHIVE_TOOL: LibarchiveTool = LibarchiveTool()
 
 LOGGER = logging.getLogger(__name__)
-
-CPIO_TOOL = ComponentExternalTool(
-    "cpio",
-    "https://www.gnu.org/software/cpio/",
-    install_check_arg="--help",
-    apt_package="cpio",
-    brew_package="cpio",
-    choco_package="cpio",
-)
 
 
 class CpioArchiveType(Enum):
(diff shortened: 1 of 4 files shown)


ofrak_core/Dockerstub
```diff
@@ -23,6 +23,7 @@ RUN apt-get -y update && \
       unar \
       unzip \
       zstd \
+      libarchive-dev \
```
whyitfor: Might be nice to add a link /explanation here or in the CPIO file, that the compiled library needs to be installed in order for python bindings to work: https://github.com/dsoprea/PyEasyArchive?tab=readme-ov-file#notes

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -1,30 +1,56 @@
...
-    "cpio",
-    "https://www.gnu.org/software/cpio/",
-    install_check_arg="--help",
-    apt_package="cpio",
-    brew_package="cpio",
-    choco_package="cpio",
-)
+    LIBARCHIVE_INSTALLED = True
+except OSError:
+    LIBARCHIVE_INSTALLED = False
```
whyitfor, line 36: As you can see in CI testing this isn't actually working for Windows, where there is a `TypeError`. Is `OSError` raised in other cases?

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -76,79 +102,197 @@ async def analyze(self, resource: Resource, config=None) -> CpioFilesystem:
...
+            int(entry.mtime) if entry.mtime else 0,  # st_mtime
+            int(entry.ctime) if entry.ctime else 0,  # st_ctime
+        )
+    )
+
+
 class CpioUnpacker(Unpacker[None]):
     """
-    Unpack a CPIO archive.
+    Unpack a CPIO archive using libarchive.
```
whyitfor:
```suggestion
    Unpack a CPIO archive.
```

This is an implementation detail

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -76,79 +102,197 @@ async def analyze(self, resource: Resource, config=None) -> CpioFilesystem:
...
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
-                cwd=temp_flush_dir,
-            )
-            await proc.communicate(input=resource_data)
-            if proc.returncode:
-                raise CalledProcessError(returncode=proc.returncode, cmd=cmd)
-            await cpio_v.initialize_from_disk(temp_flush_dir)
+
+        with libarchive.memory_reader(resource_data) as archive:
```
whyitfor, line 163: How does this compare with using CPIO in terms of speed?

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -76,79 +102,197 @@ async def analyze(self, resource: Resource, config=None) -> CpioFilesystem:
...
+        return stat.S_IFIFO
+    else:
+        raise ValueError("Unknown file type")
+
+
+def _get_libarchive_format(archive_type: CpioArchiveType) -> str:
+    """
+    Note: libarchive supports reading many CPIO variants but only supports
+    writing a subset. We map to the closest supported write format.
+    Available write formats: 'cpio' (generic), 'cpio_newc' (SVR4 newc)
```
whyitfor: - Is this a limitation of libarchive or the python bindings?
- Does this make packing lossy (are we encountering changes when repacking)?
- Can we help add support to handle this?

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -76,79 +102,197 @@ async def analyze(self, resource: Resource, config=None) -> CpioFilesystem:
...
+        CpioArchiveType.HPBIN: "cpio",
+        CpioArchiveType.HPODC: "cpio",
+    }
+    return format_map.get(archive_type, "cpio_newc")
 
 
 class CpioPacker(Packer[None]):
     """
-    Pack files into a CPIO archive.
+    Pack files into a CPIO archive using libarchive.
```
whyitfor:
```suggestion
    Pack files into a CPIO.
```

Implementation detail not needed here.

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -76,79 +102,197 @@ async def analyze(self, resource: Resource, config=None) -> CpioFilesystem:
...
-        )
-        cpio_pack_output, stderr = await cpio_pack_proc.communicate(input=list_files_list)
-        if cpio_pack_proc.returncode:
-            raise CalledProcessError(returncode=cpio_pack_proc.returncode, cmd=cpio_pack_cmd)
-        # Passing in the original range effectively replaces the original data with the new data
-        resource.queue_patch(Range(0, await resource.get_data_length()), cpio_pack_output)
+
+        format_str = _get_libarchive_format(cpio_v.archive_type)
+
+        entries = await resource.get_descendants(r_filter=ResourceFilter(tags=(FilesystemEntry,)))
```
whyitfor, line 258: Wondering if path sorting can be done here.

ofrak_core/tests/components/test_cpio_component.py
```diff
@@ -58,20 +53,266 @@ async def repack(self, cpio_resource: Resource) -> None:
...
-                patched_data = f.read()
-            assert patched_data == EXPECTED_DATA
+        await repacked_cpio_resource.unpack_recursively()
+        cpio_v = await repacked_cpio_resource.view_as(CpioFilesystem)
+        child_textfile = await cpio_v.get_entry(CPIO_ENTRY_NAME)
+        patched_data = await child_textfile.resource.get_data()
+        assert patched_data == EXPECTED_DATA
+
+
+async def test_unpacking_root(ofrak_context: OFRAKContext):
```
whyitfor: Could this (and other) tests be written using the OFRAK testing patterns?

ofrak_core/tests/components/test_cpio_component.py
```diff
@@ -58,20 +53,266 @@ async def repack(self, cpio_resource: Resource) -> None:
...
+            continue
+
+        m1 = metadata1[path]
+        m2 = metadata2[path]
+
+        # Check if this is a symlink using stat module
+        is_symlink = stat.S_ISLNK(m1["mode"])
+
+        # For symlinks, skip size check - libarchive CPIO writer doesn't preserve symlink size
+        # This is a known limitation of libarchive's add_file_from_memory for CPIO symlinks
```
whyitfor: Link for this? Is this something we could fix?

ofrak_core/src/ofrak/core/cpio.py
paulnoalhyt: Path sorting is already implemented in the following lines

ofrak_core/src/ofrak/core/cpio.py
paulnoalhyt: I ran this simple test:
```
from ofrak import *
from ofrak.core import *
import time

input_file = "core_tinycore"
async def main(ofrak_context: OFRAKContext):
    unpack = []
    repack = []
    for i in range(0, 10):
        root_resource = await ofrak_context.create_root_resource_from_file(input_file)
        start_time = time.time()
        await root_resource.unpack()
        unpack.append(time.time()-start_time)
        start_time = time.time()
        await root_resource.pack_recursively()
        repack.append(time.time()-start_time)
    print(f"Took {sum(unpack)/len(unpack)} seconds on average to unpack")
    print(f"Took {sum(repack)/len(repack)} seconds on average to repack")


if __name__ == "__main__":
    ofrak = OFRAK()
    ofrak.run(main)
```
With old implementation (`cpio` command):
```
Took 1.5415567398071288 seconds on average to unpack
Took 0.7365199565887451 seconds on average to repack
```
With this new implementation (`libarchive`):
```
Took 4.62700207233429 seconds on average to unpack
Took 0.25534915924072266 seconds on average to repack
```

So the `libarchive` implementation is about 3 times slower for unpacking, and 3 times faster to repack. This test used a 19Mb CPIO archive from the Linux TinyCore ISO.

ofrak_core/Dockerstub
paulnoalhyt: Changed that to `libarchive13`, the dev package wasn't actually required. Added a note at the top of the `Dockerstub`

ofrak_core/src/ofrak/core/cpio.py
paulnoalhyt: `OSError` was raised on MacOS when `libarchive` wasn't found. Made that check take into account all OSes.

ofrak_core/src/ofrak/core/cpio.py
paulnoalhyt: This is a limitation from `libarchive` itself (not the python bindings). See https://linux.die.net/man/5/libarchive-formats:
> The libarchive library can read a number of common cpio variants and can write ''odc'' and ''newc'' format archives

Worst case scenario an CPIO could be repacked to a different format (header's format), but the actual files data and metadata would be the same. I don't think there is much we can do.

ofrak_core/tests/components/test_cpio_component.py
paulnoalhyt: I removed some tests that were overlapping with others. For the ones I left, the issue is that the `FilesystemPackUnpackVerifyPattern` expect files being flushed to disk, which is exactly what I'm trying to avoid with this move to `libarchive`.

ofrak_core/tests/components/test_cpio_component.py
paulnoalhyt: This took me a while to investigate! The issue was a combination of `libarchive-c` python bindings, and libarchive code itself. The issue is that symlinks are created as hardlinks, then ignored during archive writing. I created an issue on the `libarchive-c` python bindings repo: https://github.com/Changaco/python-libarchive-c/issues/143. I found a fix for this, until this is fixed in the bindings.

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -1,30 +1,71 @@
...
+    # Linux throws an AttributeError, MacOS an OSError, Windows a TypeError
+    LIBARCHIVE_INSTALLED = False
+    try:
+        # On Windows, if libarchive is not installed, use the DLL from the `extractcode-libarchive` python package.
+        import sys
+
+        os.environ["LIBARCHIVE"] = os.path.join(
+            sys.prefix, "lib", "site-packages", "extractcode_libarchive", "lib", "libarchive-13.dll"
+        )
+        import libarchive
```
paulnoalhyt, line 44: As `choco` didn't provide a `libarchive` package, I found a dirty solution: install the `extractcode-libarchive` python package, which comes with `libarchive-13.dll`. When the `libarchive-c` python bindings are loaded, there is a check for the `LIBARCHIVE` environment variable. So by setting that, the bindings can use that DLL. That appears to me like the easiest solution if we don't want to compile the libarchive library ourselves, or ship a DLL in ofrak. Let me know what you think.

rbs-jacob: Since `libarchive` may presumably become a core part of many of our packers and unpackers after this change is merged, I wonder if we should cross-build the DLLs and push Python wheels with them to PyPI ourselves? If it takes less than a day to get our own builds pushed up to PyPI, I think it is probably worthwhile. (@whyitfor can weigh in on this.)

The cross-compiling part isn't difficult with GitHub actions. As a quick experiment, I did it already [here](https://github.com/rbs-jacob/libarchive-zig) using the Zig toolchain. What remains to be done there is just the "build the wheel and push it" part, which I think might not be that hard. To see what is generated, look at the [Actions build log](https://github.com/rbs-jacob/libarchive-zig/actions/runs/18386089531/job/52384763842#step:5:25).

whyitfor: @rbs-jacob, I like this idea, but let's keep it separate from the PR. Can you make an issue? Possible options would include a) pushing our own wheels b) helping the libarchive-c maintainer do this with the library.

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -1,30 +1,71 @@
...
+class LibarchiveTool(ComponentExternalTool):
+    def __init__(self):
+        super().__init__(
+            "libarchive",
+            "https://www.libarchive.org/",
+            install_check_arg="",
+            apt_package="libarchive",
+            brew_package="libarchive",
+            choco_package="",  # libarchive is not available on choco. If it is not installed, it will use the DLL provided by the `extractcode-libarchive` python package.
+        )
```
whyitfor:
```suggestion
            choco_package="",  # libarchive is not available on choco. If libarchive is not installed, OFRAK will use the DLL provided by the `extractcode-libarchive` python package.
        )
```

ofrak_core/tests/components/test_cpio_component.py
```diff
@@ -20,30 +27,19 @@
...
+        cpio_v = await cpio_r.view_as(CpioFilesystem)
+        # This also tests packing and unpacking a root file
+        await cpio_v.add_file(
+            path=CPIO_ENTRY_NAME,
+            data=INITIAL_DATA,
+            file_stat_result=os.stat_result((0o644, 0, 0, 1, 0, 0, 0, 0, 0, 0)),
+            file_xattrs=None,
+        )
+        await cpio_r.pack_recursively()
+        return cpio_r
```
whyitfor: The point of this function was to create a file without OFRAK to test OFRAK. Can you build this file once, using the cpio utility and then add it as a test asset to gitlfs?

ofrak_core/src/ofrak/core/cpio.py
rbs-jacob: Happy to make an issue. 

To be clear: you're saying that, for now, we *should* stick with the `extractcode-libarchive` Python package hack used here? If so, I agree. Just want to confirm.

ofrak_core/CHANGELOG.md
```diff
@@ -89,6 +89,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - When the user selects the "Decompilation" tab in the GUI, the pane is updated with the decompiled code automatically, without having to click "Analyze" first. ([#639](https://github.com/redballoonsecurity/ofrak/pull/639))
 - Use a single source of truth for the package version ([#640](https://github.com/redballoonsecurity/ofrak/pull/640))
 - Update the behavior of `get_only_descendant_as_view`, `get_descendants_as_view`, `get_ancestors_as_view`, and `get_only_ancestor_as_view` to retrieve all resources that match the filter. ([#642](https://github.com/redballoonsecurity/ofrak/pull/642))
+- Reimplement the CPIO components using `libarchive` ([#647](https://github.com/redballoonsecurity/ofrak/pull/647))
```
whyitfor: This should be added to unreleased! You also need to bump the VERSION in `ofrak_core/version.py`.

ofrak_core/src/ofrak/core/cpio.py
```diff
@@ -76,79 +117,224 @@ async def analyze(self, resource: Resource, config=None) -> CpioFilesystem:
...
+    format_map = {
+        CpioArchiveType.NEW_ASCII: "cpio_newc",
+        CpioArchiveType.OLD_ASCII: "cpio",
+        CpioArchiveType.CRC_ASCII: "cpio_newc",
+        CpioArchiveType.BINARY: "cpio",
+        CpioArchiveType.TAR: "cpio",
+        CpioArchiveType.USTAR: "ustar",
+        CpioArchiveType.HPBIN: "cpio",
+        CpioArchiveType.HPODC: "cpio",
+    }
```
whyitfor, line 234: @paulnoalhyt, I think I'd feel more comfortable if we didn't do this mapping here.

I think the behavior we want is the following:
1. If someone tries to pack a CPIO archive, and the archive type is not supported, we raise an error explaining that to the user.
2. This error can tell them how to do this mapping themselves (they can update the attributes on the CPIO archive to the desired output format before calling pack. We could even implement an archive format modifier to do this change.

This still gives the users the ability to pack, but they need to explicitly know they are changing the archive format.

What do you think?

ofrak_core/src/ofrak/core/cpio.py
rbs-jacob: Documented pushing our own `libarchive` binaries to PyPI in an issue here: #669.

ofrak_core/tests/components/test_cpio_component.py
paulnoalhyt: Done in https://github.com/redballoonsecurity/ofrak/pull/647/commits/b8138f62c8b4e37f70ac48727f1c0b83bd7eea5c

ofrak_core/CHANGELOG.md
paulnoalhyt: Done in https://github.com/redballoonsecurity/ofrak/pull/647/commits/3e2275d7810ee58d3fd4d379e5f4a4aa32d6ff77

ofrak_core/src/ofrak/core/cpio.py
paulnoalhyt: Done in https://github.com/redballoonsecurity/ofrak/pull/647/commits/9d108d67b807aeba1009284c204b2caa354cb849

rbs-jacob:
Now that tests are passing, I defer to @whyitfor as to when this is ready to merge.

PR #650, opened 2025-10-03
Add id for PyGhidraDecompilationAnalyzer

rbs-afflitto:
- [] I have made or updated a changelog entry for the changes in this pull request.

`PyGhidraDecompilationAnalyzer` is not discovered automatically. Adding the `id` lets it be auto discovered.

**Link to Related Issue(s)**

N/A

- Add id for PyGhidraDecompilationAnalyzer

@rbs-jacob


rbs-jacob:
> I'll push a change to fix the test, and leave it to @rbs-jacob and @whyitfor if we should accept this PR, or close it and change the filtering logic to allow components with different names and IDs.

My vote is for "both." I think we should merge this PR that will improve functionality in the short-term. We should also get to the bottom of why the API server filters out components where the name and ID are not equal, so that we can either remove that filter or fix the underlying issue.

PR #652, opened 2025-10-05
Add --version, -V flag to ofrak cli

### LINKED ISSUE #643: Add an `ofrak --version` command.
**What is the use case for the feature?**

**Does the feature contain any proprietary information about another company's intellectual property?**

**How would you implement this feature?**

whyitfor:

Add --version, -V flag to OFRAK CLI.

**Link to Related Issue(s)**
Closes #643.

1. Add version command to ofrak cli so users can easily check the version.
2. Move the version file inside of ofrak package to support this, update scripts that process version number.

@rbs-jacob


ofrak_core/src/ofrak/version.py
```diff
@@ -0,0 +1 @@
+VERSION = "3.3.0"
```
rbs-jacob: What is the benefit of removing the ability to print the version if this is called directly?

ofrak_core/src/ofrak/cli/ofrak_cli.py
```diff
@@ -12,11 +12,13 @@
 
 from importlib_metadata import entry_points
 
+from ofrak import __version__
```
rbs-jacob: Totally fine to have set `__version__` for this module, but why import the dunder one instead of the original one?

ofrak_core/src/ofrak/version.py
whyitfor: It seems unnecessary and more precise to pass "VERSION" in rather than whatever this file prints.

ofrak_core/tests/unit/test_ofrak_cli.py
```diff
@@ -318,7 +318,7 @@ def test_version_flag(cli_commands, capsys):
     captured = capsys.readouterr()
     output = captured.out + captured.err
     assert "ofrak" in output.lower()
-    assert VERSION in output
+    assert importlib.metadata.version("ofrak") in output
```
rbs-jacob, line 321: Should we assert that the `importlib` version is present *and* that `VERSION` is present? We probably want to confirm that both are present, and that both are consistent with each other so we know it's pulling the version from the right place.

ofrak_core/src/ofrak/__init__.py
```diff
@@ -16,6 +16,3 @@
     ResourceFilterCondition,
     ResourceSort,
 )
-from ofrak.version import VERSION
-
-__version__ = VERSION
```
rbs-jacob, line 21: Why remove?

ofrak_core/src/ofrak/cli/ofrak_cli.py
```diff
@@ -237,4 +237,4 @@ def parse_and_run(self, args: Sequence[str]):
 
 
 def get_ofrak_version() -> str:
-    return f"ofrak version {__version__}"
+    return f"ofrak version {importlib.metadata.version('ofrak')}"
```
rbs-jacob, line 240: I know we don't officially support Python < 3.9, but I just wanted to point out that I think this won't be supported on those versions (though it might actually be < 3.8). For those versions we'd need to use `pkg_resources`.

ofrak_core/tests/unit/test_ofrak_cli.py
whyitfor: Not sure I understand this.

With this changeset, `VERSION` is no longer part of the ofrak package. This test essentially tests that:
- OFRAK CLI's --version command returns a version
- This version matches whatever `importlib.metadata.version("ofrak")` returns

This seems reasonable to me. What do you think is missing here?

ofrak_core/src/ofrak/cli/ofrak_cli.py
whyitfor: Since we don't support < 3.9, we need to provide backwards compatibility? A lot of other things won't work (we use dependencies that require >= 3.9, pypi packages won't install for anything < 3.9, etc.

rbs-jacob: Not necessarily something we need to address. I just wanted to point out that this is a breaking change for anyone trying to get this version of OFRAK to work on older Python. We don't support < 3.9 mainly because of dependencies not supporting it, but I don't recall any parts of OFRAK proper that need >= 3.9 until this. Feel free to resolve.

ofrak_core/tests/unit/test_ofrak_cli.py
rbs-jacob:
```suggestion
    assert importlib.metadata.version("ofrak") == VERSION
    assert VERSION in output
```

There is a world in which the wrong version of the package is installed. Importlib will get *some* version of OFRAK, and we make sure that's in the CLI output, but my point is that it's worth confirming for certain that it's the version we expect.

I admit it's not 100% necessary, but I also don't see a downside of adding the extra check. Unless for editable installs importlib lists a different version number than the one we'll be importing?

ofrak_core/src/ofrak/__init__.py
whyitfor: There seem to be two ways ppl get versions: defining `__version__` and importlib. Using the importlib approach, this is no longer needed, so I removed it.

ofrak_core/src/ofrak/cli/ofrak_cli.py
whyitfor: Ok, good note.

ofrak_core/tests/unit/test_ofrak_cli.py
whyitfor: Testing wheels is a good reason why checking VERSION doesn't make sense. That's an even stronger reason to leave the check as is!

PR #672, opened 2025-11-14
Pyghidra Analysis Improvements

rbs-afflitto:

- Add detailed logging output and progress indicators to standalone analysis script
- Fix redundant re-analysis of complex blocks in the standalone analysis script 

**Link to Related Issue(s)**
#673 

- Fix an issue causing redundant unpacking of complex blocks
- Add logging
- Add progress bar when running the analysis tool via CLI

### DIFF
--- a/disassemblers/ofrak_pyghidra/CHANGELOG.md
+++ b/disassemblers/ofrak_pyghidra/CHANGELOG.md
@@ -3,14 +3,17 @@ All notable changes to `ofrak-pyghidra` will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased 0.2.0rc4](https://github.com/redballoonsecurity/ofrak/tree/master)
+## [Unreleased 0.2.0rc5](https://github.com/redballoonsecurity/ofrak/tree/master)
 
 ### Added
 - Add a PyGhidra custom load analyzer to allow for loading programs with a custom layout ([#677](https://github.com/redballoonsecurity/ofrak/pull/677))
+- Add detailed logging output and progress indicators to standalone analysis script ([#672](https://github.com/redballoonsecurity/ofrak/pull/672))
 
 ### Fixed
 - Fix Ghidra and pyghidra CodeRegion unpacker to take into account the base address that ghidra sets for PIE executables.([#627](https://github.com/redballoonsecurity/ofrak/pull/627))
 - Fix auto discovery of `PyGhidraDecompilationAnalyzer` ([#650](https://github.com/redballoonsecurity/ofrak/pull/650))
+- Fix redundant re-analysis of complex blocks in the standalone analysis script ([#672](https://github.com/redballoonsecurity/ofrak/pull/672))
+
 ### Changed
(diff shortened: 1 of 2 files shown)


disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -2,11 +2,16 @@
...
 import hashlib
 import traceback
+from typing import Any, Dict, Optional, Union
 import pyghidra
 import argparse
 import time
 import re
 import json
+import logging
+from tqdm import tqdm
```
rbs-jacob, line 13: Not sure how I feel about introducing this dependency for just this analysis...

whyitfor, line 13: @rbs-afflitto looks like you need to add this to dependencies. Does this print to stdout? I'm wondering if there is a way for us to have this in logging.

whyitfor: @rbs-jacob, I had the same thought, but understanding % done for long-running processes is useful. Do you have a suggestion here?

rbs-jacob: My suggestion is that if we add it here, we also add it elsewhere in the code! And we make sure that multiple concurrent/async tqdm progress bars running at the same time don't compete with each other and write garbage to stdout/stderr. 

I'd be okay removing it from here, or keeping it for now and adding other progress bars later. If we do the latter option, we should just document our intent to add it elsewhere in an issue.

rbs-afflitto: I'd like to keep it in, it's helpful to know that the script is still progressing, and adding the progress bar is what led to the discovery of the redundant unpacking. By default, the progress bar is only shown when this script is executed directly (passes `show_progress=True` to unpack()) so there shouldn't be an issue with logging or competing progress bars in this case.

rbs-jacob: Sure, leaving it in is no problem. 

We should still test multiple progress bars. We want to know whether multiple will work or not before we add them elsewhere. It's still possible that `show_progress` will be set to `True` while some other progress bar is running if we sprinkle `tqdm` throughout the code, so checking before we use it is still a good idea. I'm fine to put that in the issue documenting our intent to add more progress bars. 

To summarize: fine to leave this here as-is, as long as we open an issue documenting that we want more progress bars as soon as we confirm they will work together. Still need to fix failing tests associated with this change though.

rbs-afflitto: I've opened a feature request, #680, to track this.

rbs-afflitto: Added tqdm to requirements.txt. tqdm does print to stdout, but it won't be written to `ofrak.log` as that is handled by `logging`.

PR #677, opened 2025-12-10
Add a PyGhidraCustomLoadAnalyzer

paulnoalhyt:

Add a PyGhidraCustomLoadAnalyzer to allow loading of custom binary formats in PyGhidra, from MemoryRegions. 

**Link to Related Issue(s)**

None.

With the Ghidra backend we had the `GhidraCustomLoadAnalyzer`, so this is a port to the PyGhidra backend:

- introduce `PyGhidraAutoLoadProject` which corresponds to the former `PyGhidraProject`. Introduce `PyGhidraCustomLoadProject` for binaries that are not in the `_GHIDRA_AUTO_LOADABLE_FORMATS` list.
- `PyGhidraAnalysisIdentifier` now checks the `_GHIDRA_AUTO_LOADABLE_FORMATS` list to add the `PyGhidraAutoLoadProject` or `PyGhidraCustomLoadProject` tag depending on the resource.
- introduce `PyGhidraCustomLoadAnalyzer` which gets the list of MemoryRegions in a program (created by a custom unpacker for example) and passes that to the PyGhidra backend.
- modified the `unpack` method in `pyghidra_analysis.py`: when provided with a list of `memory_regions`, it deletes all data in the program, then creates regions according to the list that was passed.

Another minor thing I fixed is in `CachedCodeRegionUnpacker`, with some binaries, I found that the PyGhidra backend tried to create a few ComplexBlocks outside of the CodeRegion. Instead of raising an alert I turned that into a warning to allow users to continue working with the PyGhidra backend.
[edit] I reverted my changes in the `CachedCodeRegionUnpacker`. It turned out I had a bug in the PyGhidra components: when running the `PyGhidraCodeRegionUnpacker` before any PyGhidra analysis, it was always running the `PyGhidraAutoAnalyzer` even if the `Program` resource did not have the `PyGhidraAutoLoadProject` tag. That resulted in loading the full file in Ghidra (not using the `PyGhidraCustomLoadAnalyzer`) so the region in PyGhidra was larger than the code region in OFRAK, producing `ComplexBlocks` outside of the `CodeRegion`. Now that I fixed the `PyGhidraCodeRegionUnpacker`, it's all good!

Also added a troubleshooting section to the PyGhidra documentation.

No.

### DIFF
--- a/disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
+++ b/disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
@@ -201,12 +201,12 @@ class PyGhidraCustomLoadAnalyzer(Analyzer[None, PyGhidraCustomLoadProject]):
         self.analysis_store = analysis_store
 
     async def analyze(self, resource: Resource, config: PyGhidraCustomLoadAnalyzerConfig):
-        try:
-            program_attrs = resource.get_attributes(ProgramAttributes)
-            language = _arch_info_to_processor_id(program_attrs)
-        except NotFoundError:
-            language = None
         if config is None:
+            try:
+                program_attrs = resource.get_attributes(ProgramAttributes)
+                language = _arch_info_to_processor_id(program_attrs)
+            except NotFoundError:
+                language = None
             decomp = False
         else:
             decomp = config.decomp
--- a/disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
+++ b/disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
@@ -98,7 +98,7 @@ class CachedGhidraCodeRegionModifier(CachedGhidraCodeRegionModifier):
 
 
 @dataclass
-class PyGhidraAutoAnalyzerConfig(ComponentConfig):
+class PyGhidraAnalyzerConfig(ComponentConfig):
     decomp: bool
     language: str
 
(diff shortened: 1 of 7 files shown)


disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/components/pyghidra_components.py
```diff
@@ -145,7 +167,72 @@ async def analyze(self, resource: Resource, config: PyGhidraAutoAnalyzerConfig =
...
+    ):
+        super().__init__(resource_factory, data_service, resource_service)
+        self.analysis_store = analysis_store
+
+    async def analyze(self, resource: Resource, config: PyGhidraCustomLoadAnalyzerConfig):
+        try:
+            program_attrs = resource.get_attributes(ProgramAttributes)
+            language = _arch_info_to_processor_id(program_attrs)
+        except NotFoundError:
+            language = None
```
rbs-jacob: We probably only need to run this if `config is None`, right? If so, maybe we move it within the first branch of the conditional immediately below

disassemblers/ofrak_pyghidra/src/ofrak_pyghidra/standalone/pyghidra_analysis.py
```diff
@@ -20,16 +21,64 @@ def _parse_offset(java_object):
...
             # Java packages must be imported after pyghidra.start or pyghidra.open_program
             from ghidra.app.decompiler import DecompInterface, DecompileOptions
             from ghidra.util.task import TaskMonitor
             from ghidra.program.model.block import BasicBlockModel
             from ghidra.program.model.symbol import RefType
             from java.math import BigInteger
+            from java.io import ByteArrayInputStream
 
+            # If memory_regions are provided, delete all data and create new regions:
+            if memory_regions and len(memory_regions) > 0:
```
rbs-jacob:
```suggestion
            if memory_regions:
```

rbs-jacob (changes requested):
These changes look good so far! (I know it's still marked "draft.")

Would love to see a test or two to validate that it didn't work before, but does with these changes.

rbs-jacob (approved):
Looks good to me!

Issue #681, opened 2026-01-20
Non-deterministic bug in `SegmentInjectorModifier`


paulnoalhyt:
**What is the problem? (Here is where you provide a complete Traceback.)**

In `ofrak_core/src/ofrak/core/patch_maker/modifiers.py`, in `SegmentInjectorModifier`, the `modify` method calls `get_mem_region_with_vaddr_from_sorted`. This is implemented in `ofrak_core/src/ofrak/core/memory_region.py`.

The issue I'm running into is when I have 2 CodeRegions that have the same virtual address and the same size (this happened to me because I had the same 2 CodeRegions created by the `ElfUnpacker` and by the `CachedProgramUnpacker`). In this case, the `sorted_regions` at the beginning of `modify` can be in either order:
- Case 1: first region has data, second doesn't
	- `Resource(resource_id=8a6a306cbbbe4c4890a11a2a62564df4, tag=[MemoryRegion,CodeRegion,ElfSectionStructure,NamedProgramSection,Addressable,ProgramSection,ElfSection], data=8a6a306cbbbe4c4890a11a2a62564df4), (MemoryRegion(0x0-0x1000))`
	- `Resource(resource_id=50f74147f96b424ba955ac1bc2ce3010, tag=[CodeRegion,Addressable,MemoryRegion]), (MemoryRegion(0x0-0x1000))`
- Case 2: first region doesn't have data, second does
	- `Resource(resource_id=d28f6e64f0aa4bb6b0099f36c4b11c96, tag=[CodeRegion,Addressable,MemoryRegion]), (MemoryRegion(0x0-0x1000))`
	- `Resource(resource_id=ad48310e2f644901aa7e9683c677c9b1, tag=[Addressable,NamedProgramSection,MemoryRegion,CodeRegion,ElfSectionStructure,ProgramSection,ElfSection], data=ad48310e2f644901aa7e9683c677c9b1), (MemoryRegion(0x0-0x1000))`

`get_mem_region_with_vaddr_from_sorted` always returns the first region that contains a given vaddr.
For case 1, calling `get_mem_region_with_vaddr_from_sorted` returns a region with data, allowing the patch. For case 2, calling `get_mem_region_with_vaddr_from_sorted` returns a region without data, which triggers the ValueError exception `Cannot inject patch because the memory region at vaddr...`, because data can't be patched in a resource without data.

**If you've discovered it, what is the root cause of the problem?**

See above

**How often does the issue happen?**

[...]

PR #682, opened 2026-01-21
Fix bug in Segment Injector

paulnoalhyt:

See issue for context. This patch solves the issue, favoring a region with data to inject the patch into.

**Link to Related Issue(s)**

https://github.com/redballoonsecurity/ofrak/issues/681

- Instead of using `MemoryRegion.get_mem_region_with_vaddr_from_sorted` which doesn't allow sorting on the data/dataless criteria, implement the search directly in the `SegmentInjectorModifier` `modify` method.
- Implemented regression tests

Reasons for going with this patch:
- keep `MemoryRegion.get_mem_region_with_vaddr_from_sorted` simple. There's no reason why this generic method should favor a region with data.
- localized change: doesn't change the `MemoryRegion.get_mem_region_with_vaddr_from_sorted`, doesn't change API

No

### DIFF
--- a/ofrak_core/src/ofrak/core/patch_maker/modifiers.py
+++ b/ofrak_core/src/ofrak/core/patch_maker/modifiers.py
@@ -21,7 +21,6 @@ from ofrak.model.component_model import ComponentConfig
 from ofrak.resource import Resource
 from ofrak.service.resource_service_i import ResourceFilter, ResourceSort, ResourceSortDirection
 from ofrak_type.memory_permissions import MemoryPermissions
-from ofrak_type.error import NotFoundError
 
 LOGGER = logging.getLogger(__file__)
 
@@ -260,21 +259,30 @@ class SegmentInjectorModifier(Modifier[SegmentInjectorModifierConfig]):
                 # See PatchFromSourceModifier
                 continue
 
-            try:
-                region = MemoryRegion.get_mem_region_with_vaddr_from_sorted(
-                    segment.vm_address, sorted_regions
-                )
-            except NotFoundError:
+            # Look for region to patch, preferably with data:
+            candidate = None
+            for mem_view in sorted_regions:
+                # the first region we find should be the largest
+                if mem_view.contains(segment.vm_address):
+                    if not candidate:
+                        candidate = mem_view
+                    elif candidate.resource.get_data_id() is None:
+                        candidate = mem_view
+                    elif candidate.resource.get_data_id():
+                        LOGGER.warning(
+                            f"Found more than one region to inject patch in, using the first one."
+                        )
+            if not candidate:
                 # uninitialized section like .bss mapped to arbitrary memory range without corresponding
                 # MemoryRegion resource, no patch needed.
                 if segment.is_bss:
                     continue
                 raise
 
-            region_mapped_to_data = region.resource.get_data_id() is not None
+            region_mapped_to_data = candidate.resource.get_data_id() is not None
             if region_mapped_to_data:
                 patches = [(segment.vm_address, segment_data)]
-                injection_tasks.append((region.resource, BinaryInjectorModifierConfig(patches)))
+                injection_tasks.append((candidate.resource, BinaryInjectorModifierConfig(patches)))
             else:
                 if segment.is_bss:
                     # uninitialized section like .bss mapped to arbitrary memory range without corresponding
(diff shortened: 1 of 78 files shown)


paulnoalhyt:
I've just applied PR feedback

ofrak_core/tests/components/test_patch_maker_component.py
```diff
@@ -330,3 +333,159 @@ async def test_segment_injector_deletes_patched_descendants(ofrak_context: OFRAK
...
+    even when a dataless CodeRegion has a higher resource ID.
+
+    Setup:
+    - Create root resource with data
+    - Create CodeRegion WITH data first (gets lower ID, e.g. ID=1)
+    - Create CodeRegion WITHOUT data second (gets higher ID, e.g. ID=2)
+    """
+    root_data = b"\x00" * SIZE
+    root_resource = await ofrak_context.create_root_resource("test_root", root_data)
+    root_resource.add_tag(MemoryRegion)
```
rbs-jacob: I'm not 100% sure, but I think you might be able to do `add_view` without calling `add_tag` first.

ofrak_core/tests/components/test_patch_maker_component.py
```diff
@@ -330,3 +333,159 @@ async def test_segment_injector_deletes_patched_descendants(ofrak_context: OFRAK
...
+        code_region_with_data,
+        data_range=Range(0, SIZE),  # Maps to root's data
+    )
+
+    # Create second CodeRegion WITHOUT data (higher resource ID)
+    code_region_without_data = CodeRegion(VADDR, SIZE)
+    await root_resource.create_child_from_view(
+        code_region_without_data,
+        # No data_range and no data = dataless resource
+    )
```
rbs-jacob, line 372: The lower/higher resource IDs are only guaranteed if we're using the sequential ID service, right? In other words, this test might become flaky if we use the UUID ID service implementation, for example?

Might want to document this somewhere or add an assertion that the fixture uses the sequential ID service.

ofrak_core/tests/components/test_patch_maker_component.py
```diff
@@ -330,3 +333,159 @@ async def test_segment_injector_deletes_patched_descendants(ofrak_context: OFRAK
...
+        access_perms=MemoryPermissions.RX,
+    )
+    patch_data = b"\x90" * SIZE  # NOP sled
+
+    config = SegmentInjectorModifierConfig(segments_and_data=((segment, patch_data),))
+
+    await root_resource.run(SegmentInjectorModifier, config)
+
+    patched_data = await root_resource.get_data()
+    assert patched_data == patch_data
```
rbs-jacob, line 438: Should this assertion be copied into the previous test?

ofrak_core/src/ofrak/core/patch_maker/modifiers.py
```diff
@@ -260,21 +259,31 @@ async def modify(self, resource: Resource, config: SegmentInjectorModifierConfig
...
+                if mem_view.contains(segment.vm_address):
+                    if not candidate:
+                        candidate = mem_view
+                    elif candidate.resource.get_data_id() is None:
+                        candidate = mem_view
+                    elif candidate.resource.get_data_id():
+                        LOGGER.warning(
+                            f"Found more than one region to inject patch in, using the first one."
+                        )
+                        break
```
rbs-jacob, line 274:
```suggestion
                    break
```

I might be misunderstanding the logic here, but shouldn't this break after the first candidate is set, instead of only in the last `elif`? Otherwise it picks the last valid `mem_view`, which is the smallest, instead of the first region, which is the largest (like the comments say).

rbs-jacob (changes requested):

ofrak_core/tests/components/test_patch_maker_component.py
paulnoalhyt: I actually removed creating that MemoryRegion, it was useless.

ofrak_core/tests/components/test_patch_maker_component.py
paulnoalhyt: I added a check at the top of the 3 new tests, which will skip if not using the `SequentialIDService`

ofrak_core/tests/components/test_patch_maker_component.py
paulnoalhyt: This was more of a regression test: the `SegmentInjectorModifier` would have failed with the old implementation. But we might as well check the injected content, you're right.

ofrak_core/src/ofrak/core/patch_maker/modifiers.py
paulnoalhyt: The point of this change is not to break on the first candidate, which could be a dataless region, so we continue iterating until we find a candidate region with data. I changed this a bit to make it clearer. Also removed the warning.

rbs-jacob (approved):
Thanks for these changes, looks good to me!

Issue #529, opened 2024-12-03
ExtPacker


whyitfor:
**What is the use case for the feature?**
Add support for packing Ext filesystems. Currently only unpacking is supported: https://ofrak.com/docs/reference/ofrak/core/extfs.html.

**Does the feature contain any proprietary information about another company's intellectual property?**
No.

**How would you implement this feature?**
Agnostic. It should correctly pack the filesystem such that the filesystem works when used on the target device. This might require updating the `ExtUnpacker` to make sure it is preserving all relevant information.

**Are there any (reasonable) alternative approaches?**
N/A.

**Are you interested in implementing it yourself?**

rbs-jacob:
For posterity, here is a relevant comment from the PR that initially added the EXT unpacker.

https://github.com/redballoonsecurity/ofrak/pull/337#issuecomment-1658912199

PR #709, opened 2026-02-09
EXT packer

paulnoalhyt:

Add EXT packer, analyzer and tests.

**Link to Related Issue(s)**

https://github.com/redballoonsecurity/ofrak/issues/529

- Add `ExtAnalyzer` that uses the `dumpe2fs` command to determine all the EXT attributes.
- Add `ExtPacker` that uses the `mke2fs` command and the analysis from the `ExtAnalyzer` to create an EXT2/3/4 filesystem without having to mount anything or use loop devices. 
- Add tests using `FilesystemPackUnpackVerifyPattern` and a round trip test to make sure attributes are preserved
- Note that I had to modify the `_validate_folder_equality` method in `pytest_ofrak/src/pytest_ofrak/patterns/pack_unpack_filesystem.py` because linux will sometimes create a `lost+found` directory when extracting the filesystem. But it could also be present in the original filesystem. So I just removed that from the equality comparison.

Compared to https://github.com/redballoonsecurity/ofrak/pull/558 PR, this merge request is more compact and less complex (no need to write a debugfs script, etc).

No

### DIFF
--- a/ofrak_core/src/ofrak/core/extfs.py
+++ b/ofrak_core/src/ofrak/core/extfs.py
@@ -2,8 +2,10 @@ import asyncio
 import tempfile312 as tempfile
 from dataclasses import dataclass
 from subprocess import CalledProcessError
+from typing import Optional
 
-from ofrak import Unpacker, Resource
+from ofrak import Analyzer, Unpacker, Resource
+from ofrak.component.packer import Packer, PackerError
 from ofrak.core import (
     GenericBinary,
     FilesystemRoot,
@@ -13,6 +15,7 @@ from ofrak.core import (
     MagicDescriptionPattern,
 )
 from ofrak.model.component_model import ComponentExternalTool, ComponentConfig
+from ofrak_type.range import Range
 
 _DEBUGFS = ComponentExternalTool(
     "debugfs", "https://e2fsprogs.sourceforge.net/", "-V", brew_package="e2fsprogs"
@@ -21,7 +24,18 @@ _DEBUGFS = ComponentExternalTool(
 
 @dataclass
 class ExtFilesystem(GenericBinary, FilesystemRoot):
-    pass
+    block_size: int
+    block_count: int
+    blocks_per_group: Optional[int] = None
+    inode_size: Optional[int] = None
+    number_of_inodes: Optional[int] = None
+    reserved_block_count: Optional[int] = None
+    creator_os: Optional[str] = None
+    filesystem_features: Optional[str] = None
+    filesystem_revision: Optional[int] = None
+    volume_label: Optional[str] = None
+    last_mounted_directory: Optional[str] = None
+    uuid: Optional[str] = None
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


pytest_ofrak/src/pytest_ofrak/patterns/pack_unpack_filesystem.py
```diff
@@ -163,6 +163,11 @@ def _validate_file_data_equality(self, old_path: str, new_path: str):
     def _validate_folder_equality(self, old_path: str, new_path: str):
         old_files = os.listdir(old_path)
         new_files = os.listdir(new_path)
+        # a "lost+found" directory might be created when flushed to disk, but might also happen to be present in the filesystem, so ignore it:
+        if "lost+found" in old_files:
+            old_files.remove("lost+found")
+        if "lost+found" in new_files:
+            new_files.remove("lost+found")
         assert len(old_files) == len(new_files) and set(old_files) == set(
```
whyitfor: This seems suspect? I don't think we want our packer to introduce files like this. Are we sure this is needed?

ofrak_core/src/ofrak/core/extfs.py
```diff
@@ -45,14 +59,83 @@ class Ext4Filesystem(ExtFilesystem):
...
+        size = len(await resource.get_data())
+        if size == 0:
+            return ExtFilesystem()
+        async with resource.temp_to_disk(suffix=".extfs") as temp_path:
+            proc = await asyncio.create_subprocess_exec(
+                "dumpe2fs",
+                "-h",
+                temp_path,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.DEVNULL,
```
rbs-jacob, line 82: We should probably capture `stderr` to attach to an exception if this fails.

ofrak_core/src/ofrak/core/extfs.py
```diff
@@ -79,6 +162,86 @@ async def unpack(self, resource: Resource, config: ComponentConfig = None) -> No
...
+                features = "none," + ext_view.filesystem_features.replace(" ", ",")
+                cmd.extend(["-O", features])
+            if ext_view.filesystem_revision:
+                cmd.extend(["-r", str(ext_view.filesystem_revision)])
+            if ext_view.volume_label:
+                cmd.extend(["-L", ext_view.volume_label])
+            if ext_view.last_mounted_directory:
+                cmd.extend(["-M", ext_view.last_mounted_directory])
+            if ext_view.uuid:
+                cmd.extend(["-U", ext_view.uuid])
```
rbs-jacob, line 226: If any of these values are zero, they will not be set. These checks should probably be explicit about confirming `is not None`.

ofrak_core/tests/components/test_extfs_component.py
```diff
@@ -212,3 +217,156 @@ async def get_descendants_to_verify(self, unpacked_root_resource: Resource) -> D
...
+                    ext_blob_path,
+                ]
+                subprocess.run(command, check=True, capture_output=True)
+                for item in os.listdir(temp_dir):
+                    src = os.path.join(temp_dir, item)
+                    dst = os.path.join(str(extract_dir), item)
+                    if os.path.isdir(src):
+                        shutil.copytree(src, dst, symlinks=True)
+                    else:
+                        shutil.copy2(src, dst)
```
rbs-jacob, line 251: Why do this instead of extracting directly to the extract dir?

rbs-jacob (changes requested):

pytest_ofrak/src/pytest_ofrak/patterns/pack_unpack_filesystem.py
paulnoalhyt: Although not an EXT requirement, it is a convention to have a `lost+found` directory at the root of the filesystem. The `mke2fs` creates that directory here, without any option to not create it: https://github.com/tytso/e2fsprogs/blob/43643a57fb2d3368fbacd181a8cd713102d52a1a/misc/mke2fs.c#L3619
As this is a special case for the EXT component, I moved it from `pytest_ofrak/src/pytest_ofrak/patterns/pack_unpack_filesystem.py` to `ofrak_core/tests/components/test_extfs_component.py` in https://github.com/redballoonsecurity/ofrak/pull/709/commits/bb7e728725981ce95a17b865f2fc686143a8394a

ofrak_core/src/ofrak/core/extfs.py
paulnoalhyt: Implemented in https://github.com/redballoonsecurity/ofrak/pull/709/commits/aed2fc107ad896cf1c17326a0db1f8f26a5d7bd6

ofrak_core/src/ofrak/core/extfs.py
paulnoalhyt: Implemented in https://github.com/redballoonsecurity/ofrak/pull/709/commits/aed2fc107ad896cf1c17326a0db1f8f26a5d7bd6

ofrak_core/tests/components/test_extfs_component.py
paulnoalhyt: Implemented in https://github.com/redballoonsecurity/ofrak/pull/709/commits/aed2fc107ad896cf1c17326a0db1f8f26a5d7bd6

PR #713, opened 2026-02-17
Install tqdm when using setup.py

rbs-jacob:

PR #672 added a dependency on `tqdm` in `ofrak_pyghidra`, and updated the `requirements.txt` accordingly. It did not also update the `setup.py` file with this dependency. This caused issues when loading that package in internal code that installs OFRAK via its `setup.py` files.


rbs-jacob:
I don't remember the details 100%, but I think Wyatt added the modified requirements to reduce user friction when installing OFRAK directly (as opposed to running it in Docker). My understanding was that it would be better for a user to have an easy time getting OFRAK, with a risk of something breaking due to dependency change, rather than have a hard time getting it installed because of dependencies they might not end up using. 

For tests, Docker images, etc., we want exact pinned dependency versions, which is why the tests don't cover the setup.py files, I think. I agree it would be good to have some sanity check, but it's hard to do without also risking constant breakage due to dependency changes. If you still think we should add a test, let me know, and I will add one to this PR.

PR #715, opened 2026-03-03
Docker image unpacker

paulnoalhyt:

Docker image tarball unpacker.

**Link to Related Issue(s)**

Adds a Docker image unpacker: unpack a docker image tar ball to a filesystem that represents the filesystem from this image  once run as a container. This unpacker walks the different image layers and applies them, including whiteouts and opaque whiteouts.

No

### DIFF
--- a/ofrak_core/src/ofrak/core/docker.py
+++ b/ofrak_core/src/ofrak/core/docker.py
@@ -1,0 +1,224 @@
+"""
+Docker image tarball unpacker for OFRAK.
+
+Supports Docker image tarballs created by ``docker save``. Parses the image manifest to determine
+layer ordering, extracts each layer in order, and applies Docker whiteout semantics to reconstruct
+the final container filesystem as it would appear inside a running container.
+
+Docker whiteout files (per the OCI image layer spec):
+
+- ``.wh.<name>``: Marks ``<name>`` in the same directory for deletion from a lower layer.
+- ``.wh..wh..opq``: Opaque whiteout — all contents from lower layers in this directory are hidden;
+  only files added by the current layer remain.
+
+References:
+
+- https://github.com/moby/moby/blob/master/image/spec/v1.2.md
+- https://github.com/opencontainers/image-spec/blob/main/layer.md
+- https://github.com/moby/moby/blob/master/pkg/archive/whiteouts.go
+"""
+
+import asyncio
+import json
+import logging
+import os
+import shutil
+import tempfile312 as tempfile
+from dataclasses import dataclass
+from subprocess import CalledProcessError
+from typing import List
+
+from ofrak.component.identifier import Identifier
+from ofrak.component.packer import Packer
+from ofrak.component.unpacker import Unpacker
+from ofrak.core.filesystem import File, Folder, SpecialFileType
+from ofrak.core.tar import TAR, TarArchive
+from ofrak.model.component_model import ComponentConfig
+from ofrak.resource import Resource
+
+LOGGER = logging.getLogger(__name__)
+
+_WHITEOUT_PREFIX = ".wh."
+_OPAQUE_WHITEOUT = ".wh..wh..opq"
+
+
+@dataclass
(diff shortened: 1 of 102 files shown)


rbs-jacob:
Is there a technical reason this depends on those other branches, or is it just a result of your local git tree? I'm assuming the latter, but want to know what I should look out for when reviewing.

ofrak_core/src/ofrak/core/docker.py
```diff
@@ -0,0 +1,200 @@
...
+            stderr=asyncio.subprocess.PIPE,
+        )
+        stdout, _ = await proc.communicate()
+        if proc.returncode:
+            return []
+
+        opaque_dirs: List[str] = []
+        for entry in stdout.decode().splitlines():
+            entry = entry.strip("/")
+            basename = os.path.basename(entry)
```
rbs-jacob: Might also want to call `os.path.normpath` in case there are leading `./`.

ofrak_core/src/ofrak/core/docker.py
```diff
@@ -0,0 +1,200 @@
...
+                proc = await asyncio.create_subprocess_exec(*cmd)
+                returncode = await proc.wait()
+                if returncode:
+                    raise CalledProcessError(returncode=returncode, cmd=cmd)
+
+                manifest_path = os.path.join(extract_dir, "manifest.json")
+                with open(manifest_path) as f:
+                    manifest = json.load(f)
+
+                layers = manifest[0]["Layers"]
```
rbs-jacob, line 113: Why only pick the first entry? I'm pretty sure (~75%) Docker can bundle multiple images in the same output tarball. Might be worth at least emitting a warning here if there are more than one?

ofrak_core/src/ofrak/core/docker.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+
+    targets = (DockerImage,)
+
+    async def pack(self, resource: Resource, config=None) -> None:
+        raise NotImplementedError(
+            "Packing a DockerImage back into a Docker image tarball is not supported. "
+            "The current unpacker merges all layers into a single filesystem, so the original "
+            "layer structure cannot be reconstructed."
+        )
```
rbs-jacob, line 181: Based on the comment in this error, I'm assuming the answer is "no," but: can we just use the `TarPacker` here to not specifically implement an image packer? 

Since this is probably not possible, feel free to resolve this without comment if that's the case.

ofrak_core/tests/components/test_docker_component.py
```diff
@@ -0,0 +1,104 @@
...
+class TestDockerImagePacker:
+    async def test_pack_not_implemented(self, ofrak_context: OFRAKContext):
+        with open(DOCKER_IMAGE_ASSET, "rb") as f:
+            data = f.read()
+        resource = await ofrak_context.create_root_resource("docker.tar", data)
+        await resource.identify()
+
+        await resource.unpack()
+        with pytest.raises(NotImplementedError):
+            await resource.pack()
```
rbs-jacob: I get that we might have to do this for test coverage, but it feels weird to assert that the packer is unimplemented in a test.

rbs-jacob (changes requested):
Some small stuff, then this should be good to go!

ofrak_core/src/ofrak/core/docker.py
paulnoalhyt: I assumed there would be only one image per tarball. For now, I added a warning that only the first one will be unpacked.

ofrak_core/tests/components/test_docker_component.py
paulnoalhyt: Good point, I deleted that test and removed that unimplemented packer from test coverage.

rbs-jacob (approved):
Pending passing tests, this looks good enough to me for now. May need to make some adjustments as we encounter more Docker images in the wild, but that shouldn't block merging this at the moment.

PR #721, opened 2026-03-17
Add compression level to zip packer and prevent thread-unsafe chdir

rbs-afflitto:

- Add compression level config to `ZipPacker` via `ZipPackerConfig`
- Fix thread-unsafe `os.chdir` in `ZipPacker` by passing `cwd` to subprocess 

**Link to Related Issue(s)**
N/A

Previously, ZipPacker would always run with the default compression level (6). This PR adds an optional config to ZipPacker to allow specifying a compression level 0-9.

@rbs-jacob

### DIFF
--- a/ofrak_core/version.py
+++ b/ofrak_core/version.py
@@ -1,1 +1,1 @@
-VERSION = "3.4.0rc8"
+VERSION = "3.4.0rc9"


ofrak_core/src/ofrak/core/zip.py
```diff
@@ -82,27 +94,34 @@ class ZipPacker(Packer[None]):
     targets = (ZipArchive,)
     external_dependencies = (ZIP_TOOL,)
 
-    async def pack(self, resource: Resource, config=None):
+    async def pack(self, resource: Resource, config: ZipPackerConfig = None) -> None:
```
rbs-jacob, line 98:
```suggestion
    async def pack(self, resource: Resource, config=None) -> None:
```

I have a vague memory that adding explicit type hinting here may cause MyPy failures. If not, then feel free to ignore this comment.

ofrak_core/src/ofrak/core/zip.py
```diff
@@ -82,27 +94,34 @@ class ZipPacker(Packer[None]):
...
-            *cmd,
-        )
-        returncode = await proc.wait()
-        if proc.returncode:
-            raise CalledProcessError(returncode=returncode, cmd=cmd)
-        os.chdir(cwd)
-        with open(temp_archive, "rb") as fh:
-            resource.queue_patch(Range(0, await zip_view.resource.get_data_length()), fh.read())
+
+        with tempfile.NamedTemporaryFile(suffix=".zip", delete_on_close=False) as temp_archive:
```
rbs-jacob, line 107: Why not delete file on close? The code that reads the data from disk happens within this context manager, and thus should happen before the file would be deleted anyway.

ofrak_core/src/ofrak/core/zip.py
rbs-afflitto: This follows type hinting used in other components, see Lz4Packer here: https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/src/ofrak/core/lz4.py

ofrak_core/src/ofrak/core/zip.py
rbs-afflitto: This follows other components, such as [TarPacker](https://github.com/redballoonsecurity/ofrak/blob/master/ofrak_core/src/ofrak/core/tar.py). I'm not sure this is necessary though. We should consider updating other components that use `delete_on_close` if it's not required.

ofrak_core/src/ofrak/core/zip.py
rbs-jacob: As long as it works!

ofrak_core/src/ofrak/core/zip.py
rbs-jacob: Thanks for pointing that out. We can leave it as-is for consistency for now. Will make a mental note to change this later. Perhaps worth opening an issue? What do you think?

rbs-afflitto: Sure, I opened an issue here https://github.com/redballoonsecurity/ofrak/issues/722

rbs-afflitto: @rbs-jacob FYI, I found that removing `delete_on_close` causes problems on Windows, I documented that on the issue: https://github.com/redballoonsecurity/ofrak/issues/722

Can't say for certain that we do actually need it in every case, but it's definitely necessary at least in some packers. 

With `delete_on_close=False`, the temp file should still be deleted, just when the `with` block exits rather than when the file handler is closed.

PR #724, opened 2026-03-18
Use magic.from_file to workaround libmagic 5.47 bug

rbs-afflitto:

**Link to Related Issue(s)**

### DIFF
--- a/ofrak_core/src/ofrak/core/magic.py
+++ b/ofrak_core/src/ofrak/core/magic.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import tempfile
 from dataclasses import dataclass
 from typing import Callable, Dict, Iterable, Union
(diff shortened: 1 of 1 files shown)


ofrak_core/src/ofrak/core/magic.py
```diff
@@ -72,8 +74,14 @@ async def analyze(self, resource: Resource, config=None) -> Magic:
...
             raise ComponentMissingDependencyError(self, LIBMAGIC_DEP)
         else:
-            magic_mime = magic.from_buffer(data, mime=True)
-            magic_description = magic.from_buffer(data)
+            with tempfile.NamedTemporaryFile(delete=False) as tmp:
+                tmp.write(data)
+                tmp_path = tmp.name
+            try:
+                magic_mime = magic.from_file(tmp_path, mime=True)
+                magic_description = magic.from_file(tmp_path)
```
rbs-jacob, line 82: Let's add a comment here documenting why we're doing it this way. I could see this being the sort of thing someone would want to change later on without understanding why it was done this way in the first place.

Also this should probably have a CHANGELOG entry.

whyitfor: Won't this be a lot slower than what we are doing currently?

PR #733, opened 2026-04-03
Replace 7-Zip download location with a more reliable URL

Hainish:
Quick fix to more reliable URLs, those linked to from 7-zip.org


rbs-jacob (approved):
Thanks for your fix!

PR #739, opened 2026-04-17
Add YAFFS2 support

rbs-jacob:

Add YAFFS2 support to OFRAK.

@Didkd


ofrak_core/tests/components/test_yaffs_component.py
```diff
@@ -0,0 +1,141 @@
...
+    Yaffs2Unpacker,
+)
+from ofrak_type.src.ofrak_type.endianness import Endianness
+from pytest_ofrak.patterns.unpack_modify_pack import UnpackModifyPackPattern
+
+from pathlib import Path
+
+import pytest
+
+from ofrak.core.yaffs import Yaffs2Filesystem
```
rbs-andre: Duplicated import?
```suggestion

```

ofrak_core/tests/components/test_yaffs_component.py
```diff
@@ -0,0 +1,141 @@
...
+        assert attrs.page_size == self.page_size
+        assert attrs.spare_size == self.spare_size
+        assert attrs.endian == self.endian
+        await root_resource.unpack_recursively()
+
+    async def modify(self, unpacked_root_resource: Resource) -> None:
+        view = await unpacked_root_resource.view_as(Yaffs2Filesystem)
+        entry = await view.get_entry(MODIFY_TARGET)
+        # Patch "YAFFS2\n" -> "ofrak!\n" (same length, 7 bytes) starting at offset 6.
+        await entry.resource.run(StringPatchingModifier, StringPatchingConfig(6, "ofrak!\n"))
```
rbs-andre: I would clarify here. Reading the test before reading how the unpacker or `get_entry` works made it seem like implicitly the file contents started at offset 6, rather than entry being the entire file contents and we happen to modify starting at offset 6. 
It's not clear whether `entry` is some structure that contains metadata or just file content.

ofrak_core/src/ofrak/core/yaffs.py
```diff
@@ -0,0 +1,251 @@
...
+def detect_page_size(data: bytes) -> int:
+    """
+    Detect page size by looking for spare magic at known page size offsets.
+    """
+    for page_size in PAGE_SIZES:
+        for magic in SPARE_MAGICS:
+            end = page_size + len(magic)
+            if end <= len(data) and data[page_size:end] == magic:
+                return page_size
+    return 0
```
rbs-andre, line 142: Are we worried about spurious hits? I would consider checking at least another page to verify

rbs-jacob: I've updated, but the new logic may be hard to follow. Let me know if it looks good, or if you think I should revise.

rbs-andre: New logic makes sense to me. I wonder if it's what is making the tests fail on CI however.

rbs-jacob: I doubt they ever would have passed. I think I had some issues in how the tests were written. Fixing now.

PR #743, opened 2026-05-04
Parallel make bom

paulnoalhyt:

Add a method for parallel generation of a BOM.

**Link to Related Issue(s)**

(None)

Add a `make_bom_parallel` that uses a `ThreadPoolExecutor` to parallelize the generation of a BOM.

No


ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
```diff
@@ -273,6 +274,64 @@ def make_bom(
             self._toolchain.segment_alignment,
         )
 
+    def make_bom_parallel(
```
rbs-jacob: If this is supposed to be functionally identical to `make_bom`, I'm curious why make a separate function for it, instead of just adding concurrency to `make_bom`?

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
```diff
@@ -273,6 +274,64 @@ def make_bom(
...
+        object_map = {}
+        for o_file in object_list:
+            assembled_object = self.prepare_object(o_file)
+            object_map.update({o_file: assembled_object})
+
+        out_dir = os.path.join(self.build_dir, name + "_bom_files")
+        os.mkdir(out_dir)
+
+        c_files = [f for f in source_list if f.endswith(".c")]
+        asm_files = [f for f in source_list if f.endswith(".as") or f.endswith(".S")]
```
rbs-jacob: I'm a little concerned that this duplicates a lot of code in the original `make_bom`. Is there a way to rewrite this so that it makes many parallel calls to the original `make_bom` function instead of duplicating parts of it? 

If not, do you have any ideas for strategies that we can use to prevent this code from diverging from the original? A code comment in `make_bom` pointing here could potentially be enough.

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
```diff
@@ -273,6 +274,64 @@ def make_bom(
...
+                for f in c_files
+            ]
+            futures.extend(
+                executor.submit(
+                    self._create_object_file, f, header_dirs, out_dir, SourceFileType.ASM
+                )
+                for f in asm_files
+            )
+            for future in futures:
+                object_map.update(future.result())
```
rbs-jacob: Will this block the async event loop?

ofrak_patch_maker/tests/toolchain_c.py
```diff
@@ -269,3 +269,64 @@ def run_relocatable_test(toolchain_under_test: ToolchainUnderTest, build_dir):
...
+
+def run_make_bom_parallel_test(toolchain_under_test: ToolchainUnderTest):
+    """
+    Verify that make_bom_parallel produces an equivalent BOM to make_bom for the same source
+    list, exercising parallel compilation across the thread pool.
+
+    This test verifies that:
+    - make_bom_parallel returns the same set of object_map keys as make_bom
+    - the same strong and unresolved symbols are produced for each object
+    - the resulting object files are byte-identical between the two BOMs
```
rbs-jacob: Do we need [references to requirements](https://ofrak.com/docs/contributor-guide/requirements.html) here? You'd know better than I would, so I am assuming not in this case.

rbs-jacob (changes requested):

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
paulnoalhyt: You're right, no need to keep 2 separate functions, merged the 2!

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
paulnoalhyt: This will block the event loop, but the original implementation too. In the end, it uses `subprocess.run` which is blocking. So this new implementation blocks the event loop, but for a shorter time, as it uses threads.

ofrak_patch_maker/tests/toolchain_c.py
paulnoalhyt: this is an implementation detail, so no.

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
paulnoalhyt: solved with merge of the 2 functions.

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
rbs-jacob: Someday, it might be nicer to eliminate the thread pool and just use `asyncio.subprocess` elsewhere in the PatchMaker. Though there are good reasons not to do that: for example, if we want to avoid sprinkling `async` throughout the codebase. For now, I'm not going to let that discussion block this PR.

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
```diff
@@ -246,20 +247,18 @@ def make_bom(
...
             itertools.repeat(SourceFileType.ASM),
         )
-        result = itertools.starmap(self._create_object_file, asm_args)
-        for r in result:
-            object_map.update(r)
+        with ThreadPoolExecutor() as executor:
+            for r in executor.map(lambda args: self._create_object_file(*args), c_args):
+                object_map.update(r)
+            for r in executor.map(lambda args: self._create_object_file(*args), asm_args):
+                object_map.update(r)
```
rbs-jacob:
```suggestion
            for r in executor.map(lambda args: self._create_object_file(*args), itertools.chain(c_args, asm_args)):
                object_map.update(r)
```

I think using [`itertools.chain`](https://docs.python.org/3/library/itertools.html#itertools.chain) is better here. Without it, the C step and ASM step run serially, whereas with it, they run together in parallel.

Likely will need to lint after applying this change, as I think the line has gotten a bit long.

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
```diff
@@ -246,20 +247,18 @@ def make_bom(
...
         asm_args = zip(
             asm_files,
             itertools.repeat(header_dirs),
             itertools.repeat(out_dir),
             itertools.repeat(SourceFileType.ASM),
         )
-        result = itertools.starmap(self._create_object_file, asm_args)
-        for r in result:
-            object_map.update(r)
+        with ThreadPoolExecutor() as executor:
```
rbs-jacob: I might be overthinking things here, but [according to the docs, the `max_workers` default value is `min(32, os.cpu_count() + 4)`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor). We might want to manually set it to drop the `min` so we use all cores on a blade (or maybe just a higher min than 32). Moreover, since this is probably a CPU-bound task (rather than I/O-bound), we don't need `cpu_count() + 4` and could probably settle for just `cpu_count()`. 

Or we can leave it as-is, since it probably won't make much difference either way. Up to you. Not worth blocking on this.

rbs-jacob (changes requested):
Very small changes (one that we don't have to make at all), and then this is ready to go!

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
paulnoalhyt: Looks good, applied it myself with linting

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
paulnoalhyt: I went with `ThreadPoolExecutor(max_workers=os.cpu_count())`

rbs-jacob (approved):

ofrak_patch_maker/src/ofrak_patch_maker/patch_maker.py
whyitfor: @rbs-jacob, I agree, I do not see the value of sprinkling `async` into `ofrak-patch-maker`.

PR #749, opened 2026-05-22
Fix binwalk (and entropy) file descriptor leak

paulnoalhyt:

Fix binwalk and entropy file descriptor leak

**Link to Related Issue(s)**

None.

I was running into errors like this: `OSError: [Errno 24] Too many open files: '/usr/local/lib/python3.9/site-packages/pygments/__init__.py'`.

It turns out that binwalk creates a `ProcessPoolExecutor`, which opens a lot of file descriptors for pipes (depending on the CPU core count). When running tests that used the binwalk analyzer, it was creating a lot of FDs for each test case, which rapidly scaled and triggered the `OSError: [Errno 24] Too many open files`

A simple test I ran was this one:
```
import asyncio
import os

from ofrak import OFRAK
from ofrak.core.binwalk import BinwalkAttributes


def fd_count() -> int:
    return len(os.listdir(f"/proc/{os.getpid()}/fd"))

async def main(ofrak_context):
    resource = await ofrak_context.create_root_resource_from_file("/usr/bin/ls")
    before = fd_count()
    print(f"FDs before binwalk: {before}")
    await resource.analyze(BinwalkAttributes)
    after = fd_count()
    print(f"FDs after binwalk:  {after}  (delta = {after - before})")

if __name__ == "__main__":
    ofrak = OFRAK()
    ofrak.run(main)
```

Before my fix, this resulted in `FDs after binwalk:  116  (delta = 96)`. With the proposed fix, I saw: `FDs after binwalk:  14  (delta = -6)`.

I added a regression test for this.

@rbs-jacob ?

### DIFF
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -37,7 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 - Fix bug in Segment Injector: favor regions with data when injecting data ([#682](https://github.com/redballoonsecurity/ofrak/pull/682))
 - Pass `usedforsecurity=False` to non-cryptographic `hashlib` calls to prevent failures when Python links against FIPS OpenSSL ([#744](https://github.com/redballoonsecurity/ofrak/pull/744))
 - Fix `chdir` in UEFI components causing failing tests. ([#747](https://github.com/redballoonsecurity/ofrak/pull/747))
-- Fix binwalk file descriptor leak. ([#749](https://github.com/redballoonsecurity/ofrak/pull/749))
+- Fix binwalk and entropy file descriptor leak. ([#749](https://github.com/redballoonsecurity/ofrak/pull/749))
 
 ## [3.3.0](https://github.com/redballoonsecurity/ofrak/compare/ofrak-v3.2.0...ofrak-v3.3.0) - 2025-10-03
 
--- a/ofrak_core/CHANGELOG.md
+++ b/ofrak_core/CHANGELOG.md
@@ -37,6 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 - Fix bug in Segment Injector: favor regions with data when injecting data ([#682](https://github.com/redballoonsecurity/ofrak/pull/682))
 - Pass `usedforsecurity=False` to non-cryptographic `hashlib` calls to prevent failures when Python links against FIPS OpenSSL ([#744](https://github.com/redballoonsecurity/ofrak/pull/744))
 - Fix `chdir` in UEFI components causing failing tests. ([#747](https://github.com/redballoonsecurity/ofrak/pull/747))
+- Fix binwalk file descriptor leak. ([#749](https://github.com/redballoonsecurity/ofrak/pull/749))
 
 ## [3.3.0](https://github.com/redballoonsecurity/ofrak/compare/ofrak-v3.2.0...ofrak-v3.3.0) - 2025-10-03
 
--- a/ofrak_core/src/ofrak/core/binwalk.py
+++ b/ofrak_core/src/ofrak/core/binwalk.py
@@ -1,5 +1,5 @@
 import asyncio
-from concurrent.futures.process import ProcessPoolExecutor
+from concurrent.futures import ProcessPoolExecutor
 from dataclasses import dataclass
 from typing import Dict
 
@@ -35,8 +35,6 @@ class _BinwalkExternalTool(ComponentExternalTool):
 
 BINWALK_TOOL = _BinwalkExternalTool()
 
-pool = ProcessPoolExecutor(max_workers=1)
-
 
 @dataclass(**ResourceAttributes.DATACLASS_PARAMS)
 class BinwalkAttributes(ResourceAttributes):
@@ -66,9 +64,12 @@ class BinwalkAnalyzer(Analyzer[None, BinwalkAttributes]):
         async with resource.temp_to_disk() as temp_path:
             # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
             # overkill here.
-            offsets = await asyncio.get_running_loop().run_in_executor(
-                pool, _run_binwalk_on_file, temp_path
-            )
+            # Spin up a dedicated process pool for this analysis and shut it down afterwards so its
+            # worker process and the FDs it opens are released.
+            with ProcessPoolExecutor(max_workers=1) as pool:
+                offsets = await asyncio.get_running_loop().run_in_executor(
+                    pool, _run_binwalk_on_file, temp_path
+                )
         return BinwalkAttributes(offsets)
 
 
--- a/ofrak_core/src/ofrak/core/binwalk.py
+++ b/ofrak_core/src/ofrak/core/binwalk.py
@@ -3,7 +3,7 @@ from concurrent.futures.process import ProcessPoolExecutor
 from dataclasses import dataclass
 from typing import Dict
 
-from ofrak.resource import ResourceFactory, Resource
+from ofrak.resource import Resource
 
 from ofrak.model.resource_model import ResourceAttributes
 
@@ -19,8 +19,6 @@ except ImportError:
 
 from ofrak.core.binary import GenericBinary
 from ofrak.model.component_model import ComponentExternalTool
-from ofrak.service.data_service_i import DataServiceInterface
-from ofrak.service.resource_service_i import ResourceServiceInterface
 
 
 class _BinwalkExternalTool(ComponentExternalTool):
@@ -60,25 +58,16 @@ class BinwalkAnalyzer(Analyzer[None, BinwalkAttributes]):
     outputs = (BinwalkAttributes,)
     external_dependencies = (BINWALK_TOOL,)
 
-    def __init__(
-        self,
-        resource_factory: ResourceFactory,
-        data_service: DataServiceInterface,
-        resource_service: ResourceServiceInterface,
-    ):
-        super().__init__(resource_factory, data_service, resource_service)
-
     async def analyze(self, resource: Resource, config=None) -> BinwalkAttributes:
         if not BINWALK_INSTALLED:
             raise ComponentMissingDependencyError(self, BINWALK_TOOL)
-        pool = ProcessPoolExecutor()
-        async with resource.temp_to_disk() as temp_path:
-            # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
-            # overkill here.
-            offsets = await asyncio.get_running_loop().run_in_executor(
-                pool, _run_binwalk_on_file, temp_path
-            )
-        pool.shutdown(wait=True, cancel_futures=True)
+        with ProcessPoolExecutor(max_workers=1) as pool:
+            async with resource.temp_to_disk() as temp_path:
+                # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
+                # overkill here.
+                offsets = await asyncio.get_running_loop().run_in_executor(
+                    pool, _run_binwalk_on_file, temp_path
+                )
         return BinwalkAttributes(offsets)
 
 
--- a/ofrak_core/src/ofrak/core/binwalk.py
+++ b/ofrak_core/src/ofrak/core/binwalk.py
@@ -62,8 +62,6 @@ class BinwalkAnalyzer(Analyzer[None, BinwalkAttributes]):
         if not BINWALK_INSTALLED:
             raise ComponentMissingDependencyError(self, BINWALK_TOOL)
         async with resource.temp_to_disk() as temp_path:
-            # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
-            # overkill here.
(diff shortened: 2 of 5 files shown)


ofrak_core/src/ofrak/core/binwalk.py
```diff
@@ -67,17 +67,18 @@ def __init__(
...
             raise ComponentMissingDependencyError(self, BINWALK_TOOL)
+        pool = ProcessPoolExecutor()
         async with resource.temp_to_disk() as temp_path:
             # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
             # overkill here.
             offsets = await asyncio.get_running_loop().run_in_executor(
-                self.pool, _run_binwalk_on_file, temp_path
+                pool, _run_binwalk_on_file, temp_path
             )
+        pool.shutdown(wait=True, cancel_futures=True)
```
rbs-jacob: This should maybe be in a finally block to ensure it happens even if there is an exception, or use with ProcessPoolExecutor() as pool context manager.

Alternatively, would a single process pool executor at the module level instead of the class instance level make more sense? That way we don't create a new one on every function call, but don't have to worry about a ton of leaked file descriptors.

ofrak_core/src/ofrak/core/binwalk.py
```diff
@@ -67,17 +67,18 @@ def __init__(
         resource_service: ResourceServiceInterface,
     ):
         super().__init__(resource_factory, data_service, resource_service)
-        self.pool = ProcessPoolExecutor()
 
     async def analyze(self, resource: Resource, config=None) -> BinwalkAttributes:
         if not BINWALK_INSTALLED:
             raise ComponentMissingDependencyError(self, BINWALK_TOOL)
+        pool = ProcessPoolExecutor()
```
rbs-jacob:
```suggestion
        pool = ProcessPoolExecutor(max_workers=1)
```

We only ever want this to spawn one direct child process, rather than a pool with one child for each CPU.

Since we only ever want to run this one thing concurrently, does it still make sense to use a process pool and not a thread pool (not sure on the state of this stuff relative to the GIL these days)?

ofrak_core/src/ofrak/core/binwalk.py
```diff
@@ -67,17 +67,18 @@ def __init__(
         resource_service: ResourceServiceInterface,
     ):
         super().__init__(resource_factory, data_service, resource_service)
```
rbs-jacob: With this change, it no longer makes sense to override `__init__`.

ofrak_core/tests/components/test_binwalk_component.py
```diff
@@ -79,3 +79,32 @@ async def test_binwalk_component(ofrak_context, test_case):
...
+@pytest.mark.skipif(
+    not os.path.isdir(f"/proc/{os.getpid()}/fd"),
+    reason="Requires /proc/<pid>/fd (Linux only)",
+)
+@pytest.mark.skipif_missing_deps([BinwalkAnalyzer])
+async def test_binwalk_does_not_leak_fds(ofrak_context):
+    """
+    Regression test for the ProcessPoolExecutor FD leak in BinwalkAnalyzer.
+    """
+    fd_dir = f"/proc/{os.getpid()}/fd"
```
rbs-jacob, line 96: This test will fail on Windows (which I think is okay for now). 

Did you test on macOS in addition to Linux? I don't fully know the ways in which the `/proc` filesystem on macOS differs from Linux.

ofrak_core/tests/components/test_binwalk_component.py
```diff
@@ -79,3 +79,32 @@ async def test_binwalk_component(ofrak_context, test_case):
...
+    asset_path = os.path.join(BINWALK_ASSETS_PATH, "dirtraversal.tar")
+
+    root_resource = await ofrak_context.create_root_resource_from_file(asset_path)
+    before = len(os.listdir(fd_dir))
+    await root_resource.analyze(BinwalkAttributes)
```
rbs-jacob: Why analyze once outside the loop?

ofrak_core/CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix bug in Segment Injector: favor regions with data when injecting data ([#682](https://github.com/redballoonsecurity/ofrak/pull/682))
 - Pass `usedforsecurity=False` to non-cryptographic `hashlib` calls to prevent failures when Python links against FIPS OpenSSL ([#744](https://github.com/redballoonsecurity/ofrak/pull/744))
 - Fix `chdir` in UEFI components causing failing tests. ([#747](https://github.com/redballoonsecurity/ofrak/pull/747))
+- Fix binwalk file descriptor leak. ([#749](https://github.com/redballoonsecurity/ofrak/pull/749))
```
rbs-jacob: I'm not totally convinced that Binwalk is the only analyzer that operates this way. We might want to do a quick search (or ask a friendly LLM) to see if this pattern is used elsewhere in the code.

rbs-jacob (changes requested):

ofrak_core/src/ofrak/core/binwalk.py
whyitfor: This seems like a usecase for the `AbstractOfrakService` -- you could add it in and implement run, shutdown to create and destroy this pool...

ofrak_core/tests/components/test_binwalk_component.py
paulnoalhyt: The test is skipped if `/proc/<pid>/fd` is not found.

ofrak_core/src/ofrak/core/binwalk.py
paulnoalhyt: I moved the `ProcessPoolExecutor` to the module level, with `max_workers=1`.

ofrak_core/src/ofrak/core/binwalk.py
paulnoalhyt: @whyitfor I think that having to start and stop a binwalk service wouldn't be very intuitive for users. We don't expect binwalk to keep running in the background when we're not using it. Also I don't think any other OFRAK analyzer is defined as an `AbstractOfrakService`

ofrak_core/CHANGELOG.md
paulnoalhyt: I found only one other component: `DataSummaryAnalyzer` (`entropy.py`) uses a `ProcessPoolExecutor` that it initializes in `__init__`. So I fixed it too and added a regression test.

ofrak_core/src/ofrak/core/entropy/entropy.py
```diff
@@ -78,23 +80,24 @@ async def get_data_summary(self, resource: Resource) -> DataSummary:
...
             magnitude = await asyncio.get_running_loop().run_in_executor(
-                self.pool, sample_magnitude, data
+                pool, sample_magnitude, data
             )
             data_summary = DataSummary(entropy, magnitude)
             return data_summary
         except BrokenProcessPool:
             # If the previous one was aborted, try again with a new pool
-            self.pool = ProcessPoolExecutor()
+            pool = ProcessPoolExecutor()
```
rbs-jacob:
```suggestion
            pool = ProcessPoolExecutor(max_workers=1)
```

Should this not also initialize with `max_workers=1`?

ofrak_core/src/ofrak/core/binwalk.py
```diff
@@ -60,23 +60,14 @@ class BinwalkAnalyzer(Analyzer[None, BinwalkAttributes]):
...
     async def analyze(self, resource: Resource, config=None) -> BinwalkAttributes:
         if not BINWALK_INSTALLED:
             raise ComponentMissingDependencyError(self, BINWALK_TOOL)
         async with resource.temp_to_disk() as temp_path:
             # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
             # overkill here.
             offsets = await asyncio.get_running_loop().run_in_executor(
-                self.pool, _run_binwalk_on_file, temp_path
+                pool, _run_binwalk_on_file, temp_path
             )
```
rbs-jacob: Should this also be wrapped in a `try` like in the entropy components? Or should that `try` be removed?

rbs-jacob (changes requested):

paulnoalhyt:
I did more research on whether we needed to use a ProcessPoolExecutor for entropy and binwalk, and how to best use them:

For entropy, when using the C version, the GIL is freed, so it's better to use threads. I added a test that verifies that running entropy analysis in parallel is faster than sequential. When using the python version, the GIL is held, so we won't see any improvement when running parallel versus sequential. But I think that's fine, anyone looking for performance should use the C version anyway.

For binwalk:
- as it pure python, executing in a thread would block the GIL.
- using a class-level `ProcessPoolExecutor()` like we did in the past uses a lot of FDs, because it is never shutdown.
- using a module-level `ProcessPoolExecutor()` makes it hard to recover failures (would have to handle `BrokenProcessPool`) and maintains a lot of FDs (never shutdown).
    - if we set `ProcessPoolExecutor(max_workers=1)`, then we lose the ability to execute parallel binwalk analysis.

So what I reverted to was using a new `with ProcessPoolExecutor(max_workers=1) as pool` every time `analyze` is called (using `with` it is automatically shutdown). This has some overhead, but I thought that this was the best compromise between speed, not keeping a lot of FDs opened, and preserving the ability to run analyzers in parallel. As for entropy, I added a test that verifies that parallel analysis is faster than sequential.

ofrak_core/src/ofrak/core/binwalk.py
paulnoalhyt: Moved away from `ProcessPoolExecutor` for entropy, using threads instead. See my comment below

ofrak_core/tests/components/test_binwalk_component.py
```diff
@@ -79,3 +82,60 @@ async def test_binwalk_component(ofrak_context, test_case):
...
+        f"BinwalkAnalyzer leaked {delta} FDs across {iterations} iterations "
+        f"({before} -> {after})."
+    )
+
+
+@pytest.mark.skipif_missing_deps([BinwalkAnalyzer])
+async def test_binwalk_parallel_faster_than_sequential(ofrak_context):
+    """
+    Time two binwalk analyses run sequentially vs. run concurrently with
+    ``asyncio.gather``, and assert that the concurrent version is faster.
```
rbs-jacob:
```suggestion
    `asyncio.gather`, and assert that the concurrent version is faster.
```

Claude often gets confused that we use Sphinx-style doc comment annotations for function parameters and return values while using regular markdown for the docstring contents. This is something worth watching out for

ofrak_core/tests/components/test_entropy_component.py
```diff
@@ -82,3 +84,60 @@ def _almost_equal(bytes1: bytes, bytes2: bytes) -> bool:
...
+    assert delta < 10, (
+        f"DataSummaryAnalyzer leaked {delta} FDs across {iterations} iterations "
+        f"({before} -> {after})."
+    )
+
+
+async def test_entropy_parallel_faster_than_sequential(ofrak_context: OFRAKContext):
+    """
+    Time two entropy analyses run sequentially vs. run concurrently with
+    ``asyncio.gather``, and assert that the concurrent version is faster.
```
rbs-jacob:
```suggestion
    `asyncio.gather`, and assert that the concurrent version is faster.
```

See previous comment.

rbs-jacob (changes requested):

rbs-jacob:
Pending the tiny changes above, this looks ready to merge!

ofrak_core/src/ofrak/core/entropy/entropy.py
```diff
@@ -77,30 +73,18 @@ async def get_data_summary(self, resource: Resource) -> DataSummary:
...
-            return await self._compute_data_summary(resource, depth=depth + 1)
-
-
-def sample_entropy(
-    data: bytes, resource_id: bytes, window_size=256, max_samples=2**20
-) -> bytes:  # pragma: no cover
+    async def _compute_data_summary(self, resource: Resource) -> DataSummary:
+        data = await resource.get_data()
+        # When using the C version, the GIL is released, improving parallel performance
+        # When using the python version, the GIL is hold, so no improvements when running multiple analysis in parallel
```
rbs-jacob:
```suggestion
        # When using the python version, the GIL is held, so no improvements when running multiple analysis calls in parallel
```

ofrak_core/src/ofrak/core/binwalk.py
```diff
@@ -60,24 +58,18 @@ class BinwalkAnalyzer(Analyzer[None, BinwalkAttributes]):
...
-    ):
-        super().__init__(resource_factory, data_service, resource_service)
-        self.pool = ProcessPoolExecutor()
-
     async def analyze(self, resource: Resource, config=None) -> BinwalkAttributes:
         if not BINWALK_INSTALLED:
             raise ComponentMissingDependencyError(self, BINWALK_TOOL)
         async with resource.temp_to_disk() as temp_path:
             # Should errors be handled the way they are in the `DataSummaryAnalyzer`? Likely to be
             # overkill here.
```
rbs-jacob:
```suggestion
```

This comment is stale

rbs-jacob (changes requested):

